Issue #3, opened 2015-02-17
Support repository alias


gimite:
e.g., https://www.google.org/personfinder/2014-jammu-kashmir-floods
-> https://www.google.org/personfinder/jkf

Useful with the g.co/pf short URL. Above can be accessed with http://g.co/pf/jkf

PR #18, opened 2015-02-19
Add support of repository name aliases.

gimite:
#3

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -806,6 +806,34 @@ class BaseHandler(webapp.RequestHandler):
                 return date + timedelta(0, 3600*self.config.time_zone_offset)
             return date
 
+    def maybe_redirect_for_repo_alias(self, request):
+        """If the specified repository name is an alias, redirects to the URL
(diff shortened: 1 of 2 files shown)


gimite:
Ping? @skywhale

app/utils.py
```diff
@@ -870,6 +898,8 @@ def __init__(self, request, response, env):
         if not model.Repo.get_by_key_name(self.repo):
             if legacy_redirect.do_redirect(self):
                 return legacy_redirect.redirect(self)
+            if self.maybe_redirect_for_repo_alias(request):
```
skywhale: maybe we want to do this before the authorization check, which would be wasted if we decide to redirect?

skywhale:
Thanks for the change. This is going to be super useful !!

LGTM modulo one comment.

skywhale:
And I'm really sorry for the delayed response.

app/utils.py
gimite: Done.

gimite:
Thanks for the review. Pulled.

PR #24, opened 2015-03-27
Do not perform translation if translate_api_key is None.

gimite:
Also, output error message to console.error if translation API call fails.

### DIFF
--- a/app/resources/forms.js
+++ b/app/resources/forms.js
@@ -142,7 +142,7 @@ function remove_profile_entry(profile_entry_index) {
 // target language, and translated text.
 var translate_callback_id = 0;
 function translate(source, target, text, continuation) {
-  if (source === target || !translate_api_key || translate_api_key == 'None') {
+  if (source === target) {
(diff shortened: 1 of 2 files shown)


app/resources/forms.js
```diff
@@ -158,6 +158,8 @@ function translate(source, target, text, continuation) {
       } else if (target.length > 2) {
         // Try falling back to "fr" if "fr-CA" didn't work.
         translate(source, target.slice(0, 2), text, continuation);
+      } else {
+        console_error("Call to Translate API has failed: ", response);
```
skywhale: Are you no longer checking translate_api_key?

Is it based on the assumption that translate_api_key should always be set?

gimite: With the fix in base.html.template, translate() should no longer be called if the translate_api_key is not set, with check here:
https://github.com/google/personfinder/blob/ichikawa-translate-no-key/app/resources/forms.js#L215

The check didn't work as expected before because string 'None' was set.

PR #35, opened 2015-04-28
Redesign the UI

kevinalle:


gimite:
All of the comment below are optional, and you can add TODO and fix it later if it takes time to fix them.

"test mode" message is now less prominent than the original:
before: https://screenshot.googleplex.com/D42H9DomCTz
after: https://screenshot.googleplex.com/MStCPQ2rLUo

It may be better to keep the old style for it, or at least keep the text in red?

Also, it would be better to put some margin between the "test mode" message and "I'm looking for someone" here:
https://screenshot.googleplex.com/wfCOyV3wdAx

Also, the UI of person record view/edit page looks a bit messy:
https://screenshot.googleplex.com/JZCgDrLAFvh
Is this something you or Jessie will fix later?

PR #40, opened 2015-05-01
Check subscription check box by default.

gimite:
Also modify subscription form UI text to make it less confusing for users who don't intend to subscribe.

### DIFF
--- a/app/resources/subscribe_captcha.html.template
+++ b/app/resources/subscribe_captcha.html.template
@@ -16,12 +16,23 @@
 {% block content %}
 
 <div>
-  <h2>{% blocktrans with person.primary_full_name as full_name %}Subscribe to updates about {{full_name}}{% endblocktrans %}</h2>
-  {% if subscribe_email %}
+
+  {% if context == 'create_person' %}
+    <h2>{% blocktrans with person_record_link_html|safe as full_name %}The record for {{full_name}} has been registered{% endblocktrans %}</h2>
+  {% else %}{% if context == 'add_note' %}
+    <h2>{% blocktrans with person_record_link_html|safe as full_name %}The note for {{full_name}} has been added{% endblocktrans %}</h2>
+  {% else %}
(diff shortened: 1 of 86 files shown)


app/resources/subscribe_captcha.html.template
```diff
@@ -16,12 +16,23 @@
 {% block content %}
 
 <div>
-  <h2>{% blocktrans with person.primary_full_name as full_name %}Subscribe to updates about {{full_name}}{% endblocktrans %}</h2>
-  {% if subscribe_email %}
+
+  {% if context == 'create_person' %}
+    <h2>{% blocktrans with person_record_link_html|safe as full_name %}The record for {{full_name}} has been registered{% endblocktrans %}</h2>
+  {% else %}{% if context == 'add_note' %}
```
skywhale: is elseif still not supported?

app/subscribe.py
```diff
@@ -189,3 +196,8 @@ def post(self):
         html = ' <a href="%s">%s</a>' % (url, link_text)
         message_html = _('You have successfully subscribed.') + html
         return self.info(200, message_html=message_html)
+
+    def __get_person_record_link_html(self, person):
+        return '<a href="%s" target=_blank>%s</a>' % (
+            django.utils.html.escape(self.get_url('/view', id=person.person_record_id)),
+            person.primary_full_name)
```
skywhale: do you need to escape primary_full_name as well?

skywhale:
LGTM

app/resources/subscribe_captcha.html.template
gimite: I tried it and it didn't work. I believe we need to update the library to use it.

app/subscribe.py
gimite: Good point. Done.

gimite:
Thanks for the review. Merged to the default branch.

PR #41, opened 2015-05-04
Support xsl and xslx import

skywhale:


polo46:
This new feature looks to be based on lessons learned by the Nepal earthquake. Vert usefull, thanks!

gimite:
LGTM

Issue #44, opened 2015-05-14
Do not allow open redirect by "target" parameter of /reveal


gimite:
We should validate that it is in the same domain.

gimite:
#45 was submitted, but we also need a check that "target" parameter doesn't contain a host name at line 104? Otherwise malicious user can still put a full URL to the "target" param.

PR #45, opened 2015-05-14
Only allow /reveal to redirect to URLs within the app.

zestyping:


zestyping:
Addresses issue #44.

gimite:
Merged the change, but I will leave the original bug open due to the issue above.

PR #46, opened 2015-05-14
Add author_email and author_phone counts to the dashboard.

zestyping:
Might be interesting to know how often these fields are populated.


gimite:
LGTM. Merged.

PR #49, opened 2015-05-15
Create page takes up 1 column on mobile

Jessie-Newman:
CSS changes proposed by Rupert Breheny. On mobile (and other thin screens), the create view shows up with one column instead of two. Also various minor CSS tweaks.


gimite:
LGTM after you tested basic flow in major browsers + UC browser.

PR #58, opened 2015-06-12
Remove "overflow: hidden" from the CSS to avoid hiding content in admin pages, etc.

gimite:
Verified that major pages look fine after this change on mobile. Verified with Rupert that this is probably OK.


skywhale:
Good change. For an application like Person Finder, we should not be using "overflow: hidden".

Issue #23, opened 2015-03-23
Show all languages in the language menu


gimite:
- Show languages selected per repository first
- Show the rest of all languages below it

PR #61, opened 2015-06-30
Show all languages in the language menu. #23

gimite:
It first shows languages in the repository config, followed by all languages.


skywhale:
Which issue is this trying to solve? I'd like to know the context a bit more.

gimite:
There was a feature request, probably originally by Kazawa-san, that we should list all available languages, which is helpful for people who speak languages not manually selected.

app/main.py
```diff
@@ -207,12 +207,30 @@ def get_repo_options(request, lang):
...
+        const.LANGUAGE_ENDONYMS.keys(),
+        key=lambda s: const.LANGUAGE_ENDONYMS[s])
+    return {
+        'primary':
+            [get_language_option(request, lang, lang == current_lang)
+             for lang in primary_langs],
+        'all':
+            [get_language_option(
+                request, lang,
+                lang == current_lang and lang not in primary_langs)
```
skywhale: Why do you not want the language to be selected if it's not primary? I'm probably missing something here.

skywhale:
Thanks gimite@ for the context, and the CL LGTM. Please feel free to go ahead and merge, but I left one question. Would be nice if you could answer that.

app/main.py
gimite: Added a comment:
+            # We put both 'primary' and 'all' languages into a single <select>
+            # box (See app/resources/language-menu.html.template).
+            # If current_lang is in the primary languages, we mark the
+            # language as is_selected in 'primary', not in 'all', to make sure
+            # a single option is selected in the <select> box.

Also renamed is_current to is_selected.

gimite:

PR #63, opened 2015-07-01
Fix fraky server test.

gimite:


gimite:
e.g., test_have_information_regular was fraky.

PR #64, opened 2015-07-01
Remove support for legacy URL.

gimite:
This enables us to use http://version.appname.appspot.com/ to test pre-release versions.


skywhale:
Thanks for the long-due clean up :)

PR #68, opened 2015-07-06
Throws an exception when a reserved key is used in template vars.

gimite:
Throws an exception when "env", "config", "params" (populated by default) are used in template vars.

This is to avoid hiding these default ones by accident. Also remove code to pass "params" parameter explicitly, which is redundant.


skywhale:
Nice change :)

PR #99, opened 2015-07-21
Simple SMS interface to add a found person.

nworden:
I tested this by setting up the config to enable SMS self-additions and to add "+12345678901" as a receiver phone number, adding an API key for it, then running this small Python script with a command like:
`python sms_tester.py "i am nick"`
```
import scrape
import sys

RD = '<?xml version="1.0" encoding="utf-8"?><request>    <message_text>%s</message_text>    <receiver_phone_number>+12345678901</receiver_phone_number></request>'
URL='http://localhost:8000/testrepo1/api/handle_sms?key=API_KEY&lang=en'
s = scrape.Session(verbose=1)
doc = s.go(URL, data=RD % sys.argv[1], type='application/xml')
print doc.content
```

Could you let me know which fields are required for Person records? I found that indexing.py seems to require family_name (and, I think, given_name): if I have two people that match a search, it tries to rank them, and fails on line 129 of indexing.py (because re.match doesn't want a None argument), but there's a comment on line 118 that say those fields are optional. I am currently only filling in full_name, but I'll either need to fill in the other fields somehow or change indexing.py.

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -629,7 +629,9 @@ class HandleSMS(utils.BaseHandler):
             person = Person.create_original(
                 repo,
                 entry_date=utils.get_utcnow(),
-                full_name=name_string)
+                full_name=name_string,
+                family_name='',
+                given_name='')
             person.update_index(['old', 'new'])
             note = Note.create_original(
                 repo,
--- a/app/api.py
+++ b/app/api.py
@@ -632,9 +632,8 @@ class HandleSMS(utils.BaseHandler):
             # sender phone number and be safe with names with whitespace).
             person_record_id = '%s.%s/%s.%s-%s' % (
                 repo, HOME_DOMAIN, 'person', sender_phone_number, name_string)
-            person = Person.create_original_with_record_id(
+            person = Person.create_original(
                 repo,
-                person_record_id,
                 entry_date=utils.get_utcnow(),
                 full_name=name_string)
             person.update_index(['old', 'new'])
--- a/app/api.py
+++ b/app/api.py
@@ -37,7 +37,6 @@ import simplejson
 import subscribe
 import utils
 import xlrd
-from const import HOME_DOMAIN
 from model import Person, Note, ApiActionLog
 from text_query import TextQuery
 from utils import Struct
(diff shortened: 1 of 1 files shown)


gimite:
Thanks for implementing this!

Yeah, family_name/given_name should be optional per spec: http://zesty.ca/pfif/1.4/
So it's a bug in indexing.py. I filed a bug #100 which I will take a look.
It seems working by specifying empty family_name/given_name:
```
            person = Person.create_original(
                repo,
                entry_date=utils.get_utcnow(),
                full_name=name_string,
                family_name='',
                given_name='')
```
So can you use the workaround for now?

gimite:
Also, can you require domain_write_permission in addition to search_permission for HandleSMS because it can now also write records?

```
        if not (self.auth and self.auth.search_permission and self.auth.domain_write_permission == '*'):
            self.error(403, message=...)
            return
```

app/api.py
```diff
@@ -621,8 +624,33 @@ def post(self):
...
+                text=message_text)
+            db.put(note)
+            model.UserActionLog.put_new('add', note, copy_properties=False)
+            person.update_from_note(note)
+            db.put(person)
+            model.UserActionLog.put_new('add', person, copy_properties=False)
+            responses.append('Added record for found person: %s' % name_string)
         else:
-            responses.append('Usage: Search John')
+            usage_str = 'Usage: Search John'
```
gimite: You should also add quote here: 'Usage: "Search John"' for consistency.

app/api.py
```diff
@@ -36,6 +37,7 @@
 import subscribe
 import utils
 import xlrd
+from const import HOME_DOMAIN
```
gimite: Is this used?

nworden:
Thanks! I used empty strings for family and given name, added quotes on the other usage example, and dropped the HOME_DOMAIN import (it was there because, previously, I was trying to come up with a meaningful person record ID, which used HOME_DOMAIN, but now it just lets an ID get generated to be consistent with what the rest of PF seems to do).

My only question is about write permission. That field seems to be filled in with an actual domain name (or is left blank if the key shouldn't be granted write permission). Is it expected to sometimes be an asterisk (as in the code you put in your comment)? What I've done for now is just added "and self.auth.domain_write_permission" to the condition.

gimite:
Thanks for the fix!

Yes, you are right that domain_write_permission has been filled with actual domain names. But it is tricky in this use case because the domain we use for normal input (via Web UI) is different for each repository e.g., japan.personfinder.google.org, test.personfinder.google.org.

So my proposal here is to allow domain_write_permission to be "*", which allows the API key to use any domains.

Alternative idea is to use an unique (and fixed) domain for SMS-originated records e.g., sms.personfinder.google.org. Either sounds fine to me, so it's up to you.

nworden:
Ok, I see. I've made it require "*".

gimite:
Thanks! LGTM. Pulled.

Issue #13, opened 2015-02-17
Add photo upload API


gimite:
One of frequent requests by third party developers, including a developer of PF mobile app.

kspviswa:
Hi,

Could you please elaborate on this issue. I do see a phtoto upload dialog in official google pf site. Since photo upload feature is already available, I'm trying to understand what you meant by photo upload API. I'm sorry if this is obvious question. I'm quite new to pf

gimite:
Sorry for slow response.

It should be fine to reuse "write" API access to permit photo upload.

Your suggested flow should work too, but I prefer a flow like this:

- A user performs POST request to /REPO/api/upload_photo?key=API_KEY with a image data. (Directly putting image data to the HTTP request body would be good.)
- It returns simple XML just with a URL to the uploaded photo (with randomly generated ID) e.g., 
`<response><url>http://....jpg</url></response>`
- The user then hits normal "write" API with the URL to the photo.

I prefer this flow because the person/note record in Person Finder is basically immutable. We don't allow people to modify them after they are created. In your suggested flow, we ask people to modify a part of person record (i.e., the photo) after it is created.

@skywhale , any comment?

skywhale:
I agree with you. When we built the photo uploading demo for the intern's
Android app, we basically did exactly what you described.

kspviswa:
@gimite Thanks for your valuable comment. I have a follow-up question though. So the photo-upload API, would just save the uploaded photo with random-ID and return the generated random-ID as a response. Are we going to maintain a seperate dir to save the photo locally inside pf directory structure or do you prefer some cloud API to be used. [google-drive, onedrive etc].

gimite:
You can do the same thing as existing logic to upload a photo while
creating a new record on the web UI. It stores the image in Photo model in
datastore.

You can call photo.create_photo in the similar way as in
create.Handler#post:
https://github.com/google/personfinder/blob/master/app/create.py#L96
The method does "generate random ID" and "save the image in the datastore".

2015年5月8日(金) 14:53 kspviswa <notifications@github.com>:

kspviswa:
@gimite 

Is this issue still open? After pretty long gap I'm back to PF :) . So just want to know this, before resuming on this issue.

kspviswa:
Hi,

I had created a pull request for this feature request here : https://github.com/google/personfinder/pull/109

Please review and merge the pull request if deemed fit.

BR,
VIswa

kspviswa:
@gimite 

[...]

PR #109, opened 2015-08-05
Contains implementation for feature request PhotoUpload API #13

kspviswa:
I had implemented a new Photo_upload handler to upload photos to pf as a separate api.

HTTP URI : <PF URL>/<REPO>/api/photo_upload?key=<Write API Key>
Eg : http://localhost:8000/haiti/api/photo_upload?key=gD1q-zIfaAo3YoGT

Response will be like this.

```
kspviswa@kspviswa-VirtualBox:/media/sf_Ubuntu_Workspace/kspviswa/pf-viswa$ ./fire.sh 
<?xml version="1.0"?>
<response><url>http://localhost:8000/haiti/photo?id=4642138092470272</url></response>
```

Screen Shots
-----------------
Firing CURL query and getting the Photo-ID
====
image

Querying the photo with generated Photo-id 
=====
image


gimite:
Thanks for the pull request! I will take a look.

app/api.py
```diff
@@ -180,9 +184,9 @@ def post(self):
             self.error(400, message='Problem in the uploaded file: %s' % e)
         except runtime.DeadlineExceededError, e:
             self.error(400, message=
-                'Sorry, the uploaded file is too large. Try splitting it into '
-                'smaller files (keeping the header rows in each file) and '
-                'uploading each part separately.')
+            'Sorry, the uploaded file is too large. Try splitting it into '
```
gimite: It looks you broke indents here and other places. Can you revert them?

app/api.py
```diff
@@ -297,6 +301,37 @@ def get(self):
             self, ApiActionLog.READ, len(records), len(notes))
 
 
+class PhotoUpload(utils.BaseHandler):
+    https_required = True
+
+    def post(self):
+        if not (self.auth):
```
gimite: Redundant bracket. Should be if not self.auth:

gimite: I believe we should allow this API only for people with "write" access. Can you require self.auth and self.auth.domain_write_permission, in the same way as Write class?

app/api.py
```diff
@@ -297,6 +301,37 @@ def get(self):
...
+        if photo is not None:
+            photo = photo.read()
+            photo = images.Image(photo)
+            photo, photo_url = create_photo(photo, self)
+            if photo:
+                photo.put()
+        else :
+            self.info(
+                403,
+                message='Photo URL is wrong',
```
gimite: The message seems wrong. Something like "Request body must not be empty"?

app/api.py
```diff
@@ -297,6 +301,37 @@ def get(self):
...
+    def post(self):
+        if not (self.auth):
+            self.info(
+                403,
+                message='Missing or invalid authorization key',
+                style='plain')
+            return
+        photo = self.request.body_file
+        if photo is not None:
+            photo = photo.read()
```
gimite: Can you apply some size restriction (maybe 10MB?) to avoid uploading very large file?

gimite: Using the same variable "photo" for various types (file, string, Image) is confusing. Can you use different variable names for each of them?

app/api.py
```diff
@@ -297,6 +301,37 @@ def get(self):
...
+            photo = images.Image(photo)
+            photo, photo_url = create_photo(photo, self)
```
gimite: This function can throw SizeTooLargeError and PhotoError. Can you catch the exceptions and output appropriate error message?

app/api.py
```diff
@@ -297,6 +301,37 @@ def get(self):
...
+            if photo:
```
gimite: I believe photo can never be None, so this if-statement is not needed.

app/api.py
```diff
@@ -297,6 +301,37 @@ def get(self):
...
+            self.info(
+                403,
+                message='Photo URL is wrong',
+                style='plain')
+            return
+
+        self.response.headers['Content-Type'] = 'application/xml'
+        self.write('<?xml version="1.0"?>\n')
+        self.write('<response><url>')
+        self.write(photo_url)
```
gimite: Can you apply XML escaping just in case? It may contain "&" which should be escaped.

gimite:
Overall looks good, thank you! I added some line comments. Can you resolve them?

kspviswa:
@gimite 

Thanks for the review. I shall work on this items and revert back with another commit.

kspviswa:
This pull request is depreciated by https://github.com/google/personfinder/pull/129

PR #116, opened 2015-08-10
change json fields in admin page #105

prazuma:
#105 
change json fields(before, we have to write json directly) to per languages fields as below
- before: https://screenshot.googleplex.com/ai3wP0CGEm5
- after: https://screenshot.googleplex.com/BKw6QSaOEfX
- demo: http://personfindertest-1034.appspot.com/haiti/admin


app/admin.py
```diff
@@ -41,6 +41,7 @@ def get(self):
         view_config_json = dict(
                 (name, encoder.encode(value))
                 for name, value in view_config.iteritems())
+        configsan = encoder.encode(view_config)
```
skywhale: sorry, what does "san" stand for? "sanitized"?

app/resources/admin.html.template
```diff
@@ -67,61 +67,83 @@
   function validate_settings() {
     //find all languages that the user specified with "Use?" checkboxes and
     //their associated titles.  Only allows titles from used languages.
+    var custom_message_table_ids = ['repo_titles', 'start_page_custom_htmls', 'results_page_custom_htmls', 'view_page_custom_htmls', 'seek_query_form_custom_htmls', 'footer_custom_htmls'];
```
skywhale: could you try to keep the line under 80 characters by inserting proper indentations? this is not a strict rule in html files, but it's still helpful for those developing in limited terminal window :)

app/admin.py
gimite: This was temporary name and we planned to discuss a better name. How about "all_view_config_json"?

app/resources/admin.html.template
skywhale: And let's define this variable just before L92. In general, it makes the code more tractable to define local variables immediately before their use.

app/resources/admin.html.template
```diff
@@ -67,61 +67,83 @@
...
-      var language_code = $('language_selection_' + dropdown_index).value;
-      var language_title = $('language_title_' + dropdown_index).value;
+      var language_code = $('repo_titles_language_selection_' + dropdown_index).value;
       if (language_code === 'unused') {
         continue;
       }
       if (arr_contains(used_languages, language_code)) {
         alert("You entered the same language (" + language_code + ") twice. " +
-              "That's not allowed.");
+             "That's not allowed.");
```
skywhale: nit: Please align the indentation.

(BTW, "nit" means minor comments)

app/resources/admin.html.template
```diff
@@ -138,559 +160,616 @@
...
-        Map size in pixels [width, height]:
-      </label>
-      <div class="response">
-        <input name="map_size_pixels" id="map_size_pixels" size=24
-            value="{{view_config_json.map_size_pixels}}">
-      </div>
-    </div>
-
-  </fieldset>
+   function insert_profile_websites_example() {
```
skywhale: Do you know what made Github to think that there is a huge lines of code changed below? Did indentation change?

skywhale:
A general feedback on changeset description.

Let's avoid using the word "better" and instead describe what exactly made the UI better. Is it an additional syntax check, or just visual aesthetics?

The changeset description is an important context for the reviewer, and making it concise and right is very important.

app/resources/admin.html.template
gimite: It looks like 8 whitespaces are replaced by a tab character, which is default behavior of Emacs. Hashimoto-san, do you know how to suppress it in Emacs? Also, can you replace the tab characters back to 8 whitespaces?

gimite:
Also, in the pull request description, please mention that this change is to fix issue #105 .

prazuma:
apply comment

app/resources/admin.html.template
```diff
@@ -68,12 +68,12 @@
     //find all languages that the user specified with "Use?" checkboxes and
     //their associated titles.  Only allows titles from used languages.
     var used_languages = [];
-    var languages_to_titles = {}
+    var languages_to_values = {};
```
skywhale: ditto. could you move this to around L99?

("ditto" means "alike")

app/resources/admin.html.template
```diff
@@ -68,12 +68,12 @@
...
+
     //there is a title row in the table, so the dropdowns are 1-indexed
-    for (var dropdown_index = 1; $('language_selection_' + dropdown_index) !=
+    for (var dropdown_index = 1; $('repo_titles_language_selection_' + dropdown_index) !=
```
skywhale: Please try to keep it under 80 characters.

Could you ask Ichikawa-san how to configure Emacs so that characters exceeding 80 chars are highlighted? I use vim.

app/resources/admin.html.template
```diff
@@ -68,12 +68,12 @@
...
          null; dropdown_index++) {
-      var language_code = $('language_selection_' + dropdown_index).value;
-      var language_title = $('language_title_' + dropdown_index).value;
+      var language_code = $('repo_titles_language_selection_' + dropdown_index).value;
```
skywhale: ditto

app/resources/admin.html.template
```diff
@@ -83,45 +83,74 @@
...
+    }
+  }
+
   /* Adds a row to the languages table so that the user can add a new language
    * and give it a title.
    * Returns the number of the row that was added
    */
-  function add_language() {
-    var table = $('language_table');
+  function add_language(table_id) {
```
skywhale: table_id_prefix

app/resources/admin.html.template
```diff
@@ -83,45 +83,74 @@
...
       used_languages.push(language_code);
-      languages_to_titles[language_code] = language_title;
     }
 
     // Copy the dropdowns and the titles into the text boxes that will
     // actually be POSTed (language_menu_options and repo_titles).
     $('language_menu_options').value = JSON.stringify(used_languages);
-    $('repo_titles').value = JSON.stringify(languages_to_titles);
-
+    var custom_message_table_ids = [
```
skywhale: Sorry I didn't notice that this variable is multiple functions. We should make it a global const and share it across functions.

var CUSTOM_MESSAGE_TABLE_IDS = [
   ...
];

app/resources/admin.html.template
```diff
@@ -83,45 +83,74 @@
...
+      'repo_titles', 
+      'start_page_custom_htmls', 
+      'results_page_custom_htmls', 
+      'view_page_custom_htmls', 
+      'seek_query_form_custom_htmls', 
+      'footer_custom_htmls'
+    ];
+    for(var id_num = 0; id_num < custom_message_table_ids.length; id_num++){
+      var table_name = custom_message_table_ids[id_num];
+      for(var i = 1; i < dropdown_index; i++){
```
skywhale: s/dropdown_index/used_languages.length/

(s/A/B/ means "replace A with B")

skywhale: Shouldn't we start with i=0? Like

for(var i = 0; i < used_languages.length; i++){ ...

I may be missing something.

app/resources/admin.html.template
```diff
@@ -83,45 +83,74 @@
...
+        var used_language = used_languages[i - 1];
+        languages_to_values[used_language] = language_value;
+      }
+      $(table_name).value = JSON.stringify(languages_to_values);
+    }
     return true;
   }
 
+  function change_text(row_number) {
+    var language_codes = document.getElementById('repo_titles_language_selection_' + row_number);
```
skywhale: s/document.getElementById/$/

app/resources/admin.html.template
```diff
@@ -138,559 +167,616 @@
...
+
+    var title_col = row.insertCell(cur_col++);
+    var title = document.createElement('textarea');
+    title.cols = 72;
+    title.id = table_id + '_language_value_' + iteration;
+    title_col.appendChild(title);
+    return iteration; 
+  }
+
+  function add_another_language() {
```
skywhale: Doesn't have to be in this changeset, but I think it's a good timing to move to object-oriented Javascript programming as we are trying to do relatively complex stuff. It will help readability and maintainability.

var custom_message_tables = [];

function init_custum_message_tables() {
  custom_message_tables.push(new CustomMessageWidget($('repo_title'));
  custom_message_tables.push(new CustomLanguageMessageWidget($('start_page_custom_htmls'));
  ...
}

function add_language_to_custom_message_tables(language_code, language_name) {
  for (var i = 0; i < custom_message_tables.length; i++) {
    custom_message_tables[i].add_language(language_code, language_name);
  }
}

function remove_language_from_custom_message_tables(language_code) {
  for (var i = 0; i < custom_message_tables.length; i++) {
    custom_message_tables[i].remove_language(language_code);
  }
}

function validate_settings() {
  for (var i = 0; i < custom_message_tables.length; i++) {
    var table = custom_message_tables[i];
    $(table.id() + "_value").value = JSON.stringify(table.compile_value());
  }
}

skywhale:
Is there a demo server that I can play with?

app/resources/admin.html.template
prazuma: removed the param

app/resources/admin.html.template
prazuma: added TODO comment(move to objectoriented).
I'll do that on another branch

app/resources/admin.html.template
```diff
@@ -83,23 +94,42 @@
...
+        var used_language = used_languages[i];
+        languages_to_values[used_language] = language_value;
+      }
+      $(table_name).value = JSON.stringify(languages_to_values);
+    }
     return true;
   }
 
+  function change_text(row_number) {
+    var language_codes = $('repo_titles_language_selection_' + row_number);
```
skywhale: where is repo_titles_language_selection_i defined?

app/resources/admin.html.template
```diff
@@ -116,12 +146,15 @@
     //add the languages dropdown
     var languages_col = row.insertCell(cur_col++);
     var languages = document.createElement('select');
-    languages.id = 'language_selection_' + iteration;
+    languages.id = table_id + '_language_selection_' + iteration;
+    languages.onchange = function() {change_text(iteration)};
```
skywhale: you need a semi-colon after change_text()

app/resources/admin.html.template
```diff
@@ -83,23 +94,42 @@
...
+    }
+  }
+
   /* Adds a row to the languages table so that the user can add a new language
    * and give it a title.
    * Returns the number of the row that was added
    */
   function add_language() {
-    var table = $('language_table');
+    var table_id = 'repo_titles';
```
skywhale: Let's not define table_id. We should just use constants like 'repo_titles_table' and 'repo_titles_language_selection_'.

Why? That way, one can search within this file for keyword like 'repo_titles_language_selection_' and understand how code is structured. In the current way it's written, it's very difficult to do so.

app/resources/admin.html.template
```diff
@@ -83,23 +94,42 @@
...
+        var language_value = $(table_name + '_language_value_' + (i+1)).value;
+        var used_language = used_languages[i];
+        languages_to_values[used_language] = language_value;
+      }
+      $(table_name).value = JSON.stringify(languages_to_values);
+    }
     return true;
   }
 
+  function change_text(row_number) {
```
skywhale: Can you give a better (more descriptive name) to this function? I'm trying to understand what this function does.

app/resources/admin.html.template
skywhale: Did you manage to fix it?

prazuma:
- a demo server: http://hohoho.googlepersonfinder.appspot.com/japan/admin

app/resources/admin.html.template
prazuma: done

prazuma:
- a demo server: http://personfindertest-1034.appspot.com/japan/admin

app/resources/admin.html.template
```diff
@@ -138,35 +170,81 @@
...
-    title.id = 'language_title_' + iteration;
+    title.id = 'repo_titles_language_value_' + iteration;
     title_col.appendChild(title);
-
     return iteration;
   }
 
+  function add_language_span(table_id, language) {
+    var table = $(table_id + '_table');
+    iteration = table.rows.length;
```
skywhale: var iteration = ...

app/resources/admin.html.template
```diff
@@ -138,35 +170,81 @@
...
+        if (table_id == 'repo_titles') {
+          var row_number = add_language();
+          $(table_id + '_language_selection_' + row_number).value = language;
+        } else {
+          var row_number = add_language_span(table_id, language);
+          update_language_selected(row_number);
+        }
+        if (table_name[language] != undefined) {
+          $(table_id + '_language_value_' + row_number).value
+          = table_name[language];
```
skywhale: new line should indent by 4 spaces like:

$(table_id + '_language_value_' + row_number).value
    = table_name[language];

app/resources/admin.html.template
```diff
@@ -138,35 +170,81 @@
...
+      for (var lang_num = 0; lang_num < LANGUAGES.length; lang_num++) {
+        var language = LANGUAGES[lang_num];
+        if (table_id == 'repo_titles') {
+          var row_number = add_language();
+          $(table_id + '_language_selection_' + row_number).value = language;
+        } else {
+          var row_number = add_language_span(table_id, language);
+          update_language_selected(row_number);
+        }
+        if (table_name[language] != undefined) {
```
skywhale: "!== undefined" is the recommended way of checking inequality. I believe "null != undefined" return false.

skywhale:
Only minor comments left. LGTM otherwise.

I guess we can do a bit more cleanup as we refactor the code into OO.

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
```diff
@@ -68,38 +69,74 @@
...
-        alert("You entered the same language (" + language_code + ") twice. " +
-              "That's not allowed.");
-        return false;
-      }
       used_languages.push(language_code);
-      languages_to_titles[language_code] = language_title;
     }
 
+    var filtered_index = [];
+    used_languages.filter(function(element, index, self){
```
gimite: This part should be instead:
```
+    used_languages = used_languages.filter(function(element, index, self) {
+      if (self.indexOf(element) === index) {
+        filtered_index.push(index);
+        return true;
+      } else {
+        return false;
+      }
+    });
```

filter() function returns an array of elements for which the inner function returns true. So you should use it.

app/resources/admin.html.template
```diff
@@ -39,17 +39,18 @@
...
-      if (arr[i] === elem) {
-        return true;
-      }
-    }
-    return false;
-  }
+//TODO: move to object-oriented!!
+
+  var LANGUAGE_EXONYMS = {{language_exonyms_json|safe}};
+  var CUSTOM_MESSAGE_TABLE_IDS = [
```
gimite: repo_titles is not a custom message, so naming this "CUSTOM_MESSAGE_TABLE_IDS" is a bit weird. Can you rename it to e.g., "MULTILINGUAL_TABLE_NAMES"?

app/resources/admin.html.template
```diff
@@ -68,38 +69,74 @@
...
+      }
+    });
+
     // Copy the dropdowns and the titles into the text boxes that will
     // actually be POSTed (language_menu_options and repo_titles).
     $('language_menu_options').value = JSON.stringify(used_languages);
-    $('repo_titles').value = JSON.stringify(languages_to_titles);
+    var languages_to_values = {};
+
+    for (var id_num = 0; id_num < CUSTOM_MESSAGE_TABLE_IDS.length; id_num++){
```
gimite: Put whitespace between ) and { here and other places.

    for (var id_num = 0; id_num < CUSTOM_MESSAGE_TABLE_IDS.length; id_num++) {

app/resources/admin.html.template
```diff
@@ -116,12 +153,15 @@
     //add the languages dropdown
     var languages_col = row.insertCell(cur_col++);
     var languages = document.createElement('select');
-    languages.id = 'language_selection_' + iteration;
+    languages.id = 'repo_titles_language_selection_' + iteration;
+    languages.onchange = function() {update_language_selected(iteration);check_language_selected(iteration);};
```
gimite: This line is over 80 characters. Break this line into multiple lines.

app/resources/admin.html.template
```diff
@@ -68,38 +69,74 @@
...
+        var used_language = used_languages[index];
+        languages_to_values[used_language] = language_value;
+      }
+      $(table_name).value = JSON.stringify(languages_to_values);
+    }
 
     return true;
   }
 
+  function check_language_selected(row_number) {
```
gimite: validate_language_selected may be a clearer name. "Validate" means checking if the input is valid. "Check" is more ambiguous.

gimite: As chatted offline, this doesn't work if you add the same new language twice. Please fix.

app/resources/admin.html.template
```diff
@@ -68,38 +69,74 @@
...
+  }
+
+  function update_language_selected(row_number) {
+    var language_codes = $('repo_titles_language_selection_' + row_number);
+    var selected_index = language_codes.selectedIndex;
+    var language = language_codes[selected_index].text;
+    var language_selected_elements = document.getElementsByClassName(
+                                     'language_selected_' + row_number);
+    for (var i = 0; i < language_selected_elements.length; i++) {
+      language_selected_elements[i].innerHTML = language;
```
gimite: Can you apply HTML escaping e.g., "<" to "&lt;" to language? It's unlikely that language name contains such characters, but it would be better to prepare for it.

app/resources/admin.html.template
```diff
@@ -68,38 +69,74 @@
...
+    });
+
     // Copy the dropdowns and the titles into the text boxes that will
     // actually be POSTed (language_menu_options and repo_titles).
     $('language_menu_options').value = JSON.stringify(used_languages);
-    $('repo_titles').value = JSON.stringify(languages_to_titles);
+    var languages_to_values = {};
+
+    for (var id_num = 0; id_num < CUSTOM_MESSAGE_TABLE_IDS.length; id_num++){
+      var table_name = CUSTOM_MESSAGE_TABLE_IDS[id_num];
```
gimite: You sometimes call this "table_name", and sometimes "table_id". It would be better to be consistent.

app/resources/admin.html.template
gimite: It's too confusing to have add_another_language() and add_language(). Maybe rename add_language() to a clearer name e.g., add_language_row_with_drop_down()?

app/resources/admin.html.template
```diff
@@ -39,17 +39,18 @@
...
-        return true;
-      }
-    }
-    return false;
-  }
+//TODO: move to object-oriented!!
+
+  var LANGUAGE_EXONYMS = {{language_exonyms_json|safe}};
+  var CUSTOM_MESSAGE_TABLE_IDS = [
+    'repo_titles',
```
gimite: Can you add a comment saying that the first element of this array must be 'repo_titles'? Some of your code assumes that.

app/resources/admin.html.template
```diff
@@ -138,35 +178,80 @@
...
-    $('language_selection_' + row_number).value = 'unused';
+    for (var id_num = 0; id_num < CUSTOM_MESSAGE_TABLE_IDS.length; id_num++) {
+      var table_name = CUSTOM_MESSAGE_TABLE_IDS[id_num];
+      var table = $(table_name + '_table');
+      var row = table.rows[row_number];
+      row.hidden = true;
+      if (table_name == 'repo_titles') {
+        $(table_name + '_language_selection_' + row_number).value = 'unused';
+      } else {
+        document.getElementsByClassName('language_selected_' + row_number).value
```
gimite: This is actually not working because the return value of document.getElementsByClassName() is an array, not an element. Please fix.

app/resources/admin.html.template
```diff
@@ -138,35 +178,80 @@
...
+            = 'unused';
+      }
+    }
   }
 
   // Adds rows to the languages table for each language already in the database
   function add_initial_languages() {
     // Saved languages and titles for this repository
+    var all_view_config_json = {{all_view_config_json|safe|default:'[]'}};
```
gimite: Can you rename this variable to just all_view_config?  i.e.,
```
var all_view_config = {{all_view_config_json|safe|default:'[]'}};
```
Because the value of the variable is just a JavaScript object, no longer a JSON string.

app/resources/admin.html.template
```diff
@@ -138,35 +178,80 @@
...
     var title = document.createElement('input');
     title.type = 'text';
-    title.id = 'language_title_' + iteration;
+    title.id = 'repo_titles_language_value_' + iteration;
     title_col.appendChild(title);
-
     return iteration;
   }
 
+  function add_language_span(table_id, language) {
```
gimite: The name of this function is a bit confusing because it doesn't just add a span. Can you rename this to e.g., add_language_row_with_span?

app/resources/admin.html.template
```diff
@@ -138,35 +178,80 @@
...
   }
 
   // Adds rows to the languages table for each language already in the database
   function add_initial_languages() {
     // Saved languages and titles for this repository
+    var all_view_config_json = {{all_view_config_json|safe|default:'[]'}};
     var LANGUAGES = {{view_config_json.language_menu_options|safe|default:'[]'}};
-    var TITLES = {{view_config_json.repo_titles|safe|default:'[]'}};
-
     //foreach language, add that language and its title
```
gimite: Can you update this comment? Now this block does more than just adding a language and a title.

app/resources/admin.html.template
```diff
@@ -68,38 +69,74 @@
...
+    var selected_index = language_codes.selectedIndex;
+    var language_code = language_codes[selected_index].value;
+    var used_languages =
+        {{view_config_json.language_menu_options|safe|default:'[]'}};
+    if (used_languages.indexOf(language_code) >= 0) {
+      alert('You entered the same language!!');
+    }
+  }
+
+  function update_language_selected(row_number) {
```
gimite: Same as above. Rename this to e.g., update_selected_languages.

app/resources/admin.html.template
```diff
@@ -138,35 +178,80 @@
...
+      var table_id = CUSTOM_MESSAGE_TABLE_IDS[id_num];
+      var table_name = all_view_config_json[table_id];
+      for (var lang_num = 0; lang_num < LANGUAGES.length; lang_num++) {
+        var language = LANGUAGES[lang_num];
+        if (table_id == 'repo_titles') {
+          var row_number = add_language();
+          $(table_id + '_language_selection_' + row_number).value = language;
+        } else {
+          var row_number = add_language_span(table_id, language);
+          update_language_selected(row_number);
```
gimite: update_language_selected() updates all tables, so actually you don't need to call this for each table. Can you move this code out of the inner for-loop?

app/resources/admin.html.template
```diff
@@ -138,35 +178,80 @@
...
     //foreach language, add that language and its title
-    for (var lang_num = 0; lang_num < LANGUAGES.length; lang_num++) {
-      var row_number = add_language();
-      var language = LANGUAGES[lang_num];
-      $('language_selection_' + row_number).value = language;
-      if (TITLES[language] != undefined) {
-        $('language_title_' + row_number).value = TITLES[language];
+    for (var id_num = 0; id_num < CUSTOM_MESSAGE_TABLE_IDS.length; id_num++) {
+      var table_id = CUSTOM_MESSAGE_TABLE_IDS[id_num];
+      var table_name = all_view_config_json[table_id];
```
gimite: Can you rename this to e.g., "values"? They are values, not a name.

app/resources/admin.html.template
```diff
@@ -68,38 +69,74 @@
...
+        languages_to_values[used_language] = language_value;
+      }
+      $(table_name).value = JSON.stringify(languages_to_values);
+    }
 
     return true;
   }
 
+  function check_language_selected(row_number) {
+    var language_codes = $('repo_titles_language_selection_' + row_number);
```
skywhale: This does not prevent the user from adding multiple of the same language in this session. What about:

var used_languages = new Array();
for (var index = 1; $('repo_titles_language_selection_' + index) != null; index++) {
  var language_code = $('repo_titles_language_selection_' + index).value;
  if (used_languages[language_code] != null) {
    alert(...);
  }
  used_languages[language_code] = 1;
}

app/resources/admin.html.template
```diff
@@ -68,38 +69,74 @@
...
+                             dropdown_index).value;
       if (language_code === 'unused') {
         continue;
       }
-      if (arr_contains(used_languages, language_code)) {
-        alert("You entered the same language (" + language_code + ") twice. " +
-              "That's not allowed.");
-        return false;
-      }
       used_languages.push(language_code);
```
skywhale: Let's use an associative array, then you can remove the following code with filter().

var language_code_to_index_map = new Array();

http://blog.xkoder.com/2008/07/10/javascript-associative-arrays-demystified/

app/resources/admin.html.template
```diff
@@ -68,38 +69,74 @@
...
+    for (var i = 0; i < language_selected_elements.length; i++) {
+      language_selected_elements[i].innerHTML = language;
+    }
+  }
+
   /* Adds a row to the languages table so that the user can add a new language
    * and give it a title.
    * Returns the number of the row that was added
    */
   function add_language() {
```
skywhale: Let's rename to add_language_to_repo_titles_table

app/resources/admin.html.template
skywhale: Please break lines to fit lines under 80 chars:

languages.onchange = function() {
    update_language_selected(iteration);
    check_language_selected(iteration);
};

app/resources/admin.html.template
skywhale: Let's rename to add_language_to_custom_message_table

app/resources/admin.html.template
skywhale: Let's rename to custom_messages or something alike. table_name sounds like name of <table> element.

app/resources/admin.html.template
```diff
@@ -138,35 +178,80 @@
...
-      if (TITLES[language] != undefined) {
-        $('language_title_' + row_number).value = TITLES[language];
+    for (var id_num = 0; id_num < CUSTOM_MESSAGE_TABLE_IDS.length; id_num++) {
+      var table_id = CUSTOM_MESSAGE_TABLE_IDS[id_num];
+      var table_name = all_view_config_json[table_id];
+      for (var lang_num = 0; lang_num < LANGUAGES.length; lang_num++) {
+        var language = LANGUAGES[lang_num];
+        if (table_id == 'repo_titles') {
+          var row_number = add_language();
+          $(table_id + '_language_selection_' + row_number).value = language;
```
skywhale: Again, let's hardcode 'repo_titles_language_selection_'

app/resources/admin.html.template
```diff
@@ -195,7 +280,7 @@
 
 <p>
   <form method="get" style="display: inline-block" class="admin">
-    <select name="repo" id="select_repo" onchange="select_repo_changed()">
+    <select name="repo" id="select_repo" onchange="select_repo_changed();">
```
skywhale: you don't need the semicolon. let's not introduce unnecessary diffs.

app/resources/admin.html.template
skywhale: We probably should remove 'repo_titles' from CUSTOM_MESSAGE_TABLE_IDS? Then we can keep the const name as is. Since you have a lot of special casing for 'repo_titles', it doesn't probably make sense to be a member of this list.

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
gimite: If you use associative array, it should be instead:

    var language_code_to_index_map = {};

I'm not sure why the page uses new Array().

prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
```diff
@@ -68,38 +68,99 @@
     //find all languages that the user specified with "Use?" checkboxes and
     //their associated titles.  Only allows titles from used languages.
     var used_languages = [];
-    var languages_to_titles = {}
```
skywhale: nit: need a semicolon

app/resources/admin.html.template
```diff
@@ -68,38 +68,99 @@
...
-        alert("You entered the same language (" + language_code + ") twice. " +
-              "That's not allowed.");
-        return false;
-      }
       used_languages.push(language_code);
-      languages_to_titles[language_code] = language_title;
+    }
+
+    var language_code_to_index_map = {};
+    for (var i = 0; i < used_languages.length; i++) {
```
skywhale: Can you combine this for loop with the one above? Please just ping me if what I'm saying is not clear.

app/resources/admin.html.template
```diff
@@ -68,38 +68,99 @@
...
     // Copy the dropdowns and the titles into the text boxes that will
     // actually be POSTed (language_menu_options and repo_titles).
     $('language_menu_options').value = JSON.stringify(used_languages);
-    $('repo_titles').value = JSON.stringify(languages_to_titles);
+
+    for (var id_num = 0; id_num < MULTILINGUAL_TABLE_IDS.length; id_num++) {
+      var table_id = MULTILINGUAL_TABLE_IDS[id_num];
+      var index = 1;
+      for(var key in language_code_to_index_map){
+        var language_value = $(table_id + '_language_value_' + index++).value;
```
skywhale: var index = language_code_to_index_map[key];

app/resources/admin.html.template
```diff
@@ -68,38 +68,99 @@
...
+        var used_language = key;
+        language_code_to_index_map[used_language] = language_value;
+      }
+      console.log(language_code_to_index_map);
```
skywhale: Remove?

app/resources/admin.html.template
```diff
@@ -68,38 +68,99 @@
...
+    escaped_str = str.replace(/[<>&"\']/g, function (match) {
+      return escape_table[match];
+    });
+    return escaped_str;
+  }
+
+  function update_language_selected(row_number) {
+    var language_codes = $('repo_titles_language_selection_' + row_number);
+    var selected_index = language_codes.selectedIndex;
+    var language = language_codes[selected_index].text;
```
skywhale: selected_language to be consistent with above.

app/resources/admin.html.template
```diff
@@ -68,38 +68,99 @@
...
+    var language_selected_elements = document.getElementsByClassName(
+                                     'language_selected_' + row_number);
+    for (var i = 0; i < language_selected_elements.length; i++) {
+      escaped_language = escape_html(language);
```
skywhale: Let's do this outside the for loop.

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: removed

app/resources/admin.html.template
prazuma: index is used for 'table_id_language_value_'+number(<-index)

app/resources/admin.html.template
prazuma: ? it is deleted

app/resources/admin.html.template
```diff
@@ -68,38 +68,97 @@
...
+      if (language_code != 'unused' && used_languages[language_code] != null) {
+        alert('You entered the same language!!');
+      }
+      used_languages[language_code] = 1;
+    }
+  }
+
+  function escape_html(str) {
+    var escape_table = {
+      '<': '&lt',
```
gimite: I believe you need ";" after each escaped character e.g., '&lt;', '&gt;'

app/resources/admin.html.template
```diff
@@ -68,38 +68,97 @@
...
+      '>': '&gt',
+      '&': '&amp',
+      '"': '&quot',
+      '\'': '&#x27',
+    };
+    escaped_str = str.replace(/[<>&"\']/g, function (match) {
```
gimite: Must be "var escaped_str", or you can just return it directly here.
Without "var", escaped_str will be a global variable.

app/resources/admin.html.template
```diff
@@ -138,37 +203,94 @@
...
+              add_language_to_custom_message_table(table_id, language);
+        if (table_values !== undefined && table_values[language] !== undefined) {
+          $(table_id + '_language_value_' + row_number).value
+              = table_values[language];
+        }
       }
     }
+
+    var language_num = $('repo_titles_table').rows.length - 1;
+    for (var row_num = 0; row_num < language_num; row_num++) {
```
gimite: It would be easier to read if this row_num iterates from 1 to $('repo_titles_table').rows.length - 1 instead. Then you can remove these -1, +1.

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
```diff
@@ -68,38 +68,97 @@
...
     //there is a title row in the table, so the dropdowns are 1-indexed
-    for (var dropdown_index = 1; $('language_selection_' + dropdown_index) !=
+    for (var dropdown_index = 1; $('repo_titles_language_selection_' + dropdown_index) !=
          null; dropdown_index++) {
-      var language_code = $('language_selection_' + dropdown_index).value;
-      var language_title = $('language_title_' + dropdown_index).value;
+      var language_code = $('repo_titles_language_selection_' + 
+                             dropdown_index).value;
+      if (language_code_to_index_set[language_code] == 1) {
+        used_languages.splice(i-1, 1)
```
gimite: Maybe this code is not working? Where is this "i" defined?

app/resources/admin.html.template
```diff
@@ -68,38 +68,97 @@
     //find all languages that the user specified with "Use?" checkboxes and
     //their associated titles.  Only allows titles from used languages.
     var used_languages = [];
-    var languages_to_titles = {}
+    var language_code_to_index_set = {};
```
gimite: Is this related to "index"? Otherwise how about renaming it to just "used_language_set" or "language_code_set"?

app/resources/admin.html.template
```diff
@@ -68,38 +68,97 @@
...
+    for (var used_language in language_code_to_index_set) {
+      var language_value = $('repo_titles_language_value_' + index++).value
+      language_code_to_index_set[used_language] = language_value;
+    }
+    $('repo_titles').value = JSON.stringify(language_code_to_index_set);
+
+    for (var id_num = 0; id_num < MULTILINGUAL_TABLE_IDS.length; id_num++) {
+      var table_id = MULTILINGUAL_TABLE_IDS[id_num];
+      index = 1;
+      for(var used_language in language_code_to_index_set){
```
gimite: Put whitespace between } and (

app/resources/admin.html.template
prazuma: fixed

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
```diff
@@ -68,38 +68,99 @@
...
+        $('repo_titles_language_selection_' + row_number);
+    var selected_index = selected_language_codes.selectedIndex;
+    var selected_language = selected_language_codes[selected_index].value;
+    return selected_language;
+  }
+
+  function validate_selected_languages(row_number) {
+    var used_languages = {};
+    var language_num = $('repo_titles_table').rows.length;
+    var selected_language = get_selected_language(row_number);
```
skywhale: selected_language is not used.

app/resources/admin.html.template
```diff
@@ -138,37 +205,94 @@
...
+            repo_title_values[language];
       }
     }
+
+    for (var id_num = 0; id_num < MULTILINGUAL_TABLE_IDS.length; id_num++) {
+      var table_id = MULTILINGUAL_TABLE_IDS[id_num];
+      var table_values = all_view_config[table_id];
+      for (var lang_num = 0; lang_num < LANGUAGES.length; lang_num++) {
+        var language = LANGUAGES[lang_num];
+          var row_number =
```
skywhale: Please fix indentation and 80 char rule.

app/resources/admin.html.template
```diff
@@ -68,38 +68,99 @@
...
+    var escaped_str = str.replace(/[<>&"\']/g, function (match) {
+      return escape_table[match];
+    });
+    return escaped_str;
+  }
+
+  function update_language_selected(row_number) {
+    var language_codes = $('repo_titles_language_selection_' + row_number);
+    var selected_index = language_codes.selectedIndex;
+    var selected_language = language_codes[selected_index].text;
```
skywhale: Let's get get_selected_language()

skywhale:
Please address the remaining comments, and LGTM :)

Please also update the demo and check if the function is working end to end.

Thank you for your patience.

app/resources/admin.html.template
prazuma: deleted

app/resources/admin.html.template
prazuma: done

app/resources/admin.html.template
prazuma: get_selected_language() returns language_code.
in update_language_selected(), selected_language = text(repo_titles_language_selected)

skywhale:
お疲れ様！
Nice work and thanks for your patience all the way!!

prazuma:
٩(๑❛ᴗ❛๑)۶ﾜｰｲヽ(ﾟ∀ﾟ)ﾒ(ﾟ∀ﾟ)ﾒ(ﾟ∀ﾟ)ﾉﾜｰｲﾜｰｲヽ(ﾟ∀ﾟ)ﾒ(ﾟ∀ﾟ)ﾒ(ﾟ∀ﾟ)ﾉﾜｰｲ٩(๑❛ᴗ❛๑)۶

Issue #78, opened 2015-07-06
Automate Environment setup


gimite:
The Getting Started Guide explains how to install all tools required to develop PersonFinder as well as their version.
However, sometimes it can be painful to make sure this is working fine and that we have the right setup.

It would be great to be able to have something more automatic: either some sort of image of a setup to clone (and project owner would update this image regularly when the tools versions are upgraded) or some script to run which will install what is missing with the correct version...

clobrano:
What about a Docker container? I just started looking around in this project, so I am not 100% sure that this solution will suit it

clobrano:
Well, I had time to made some tests and I managed to complete the steps from the "Getting Started" page within a docker image with this configuration:

* all the tools and dependencies in the docker container
* the personfinder's source code in a host's folder shared with the container
* the gae test server running inside the container, but accessible from the host's web browser

gimite:
Thanks for the suggestion and pull request! I will take a look.

There is another separated issue that you need to call setup_datastore() manually when you create a new app instance, as instructed in https://github.com/google/personfinder/wiki/GettingStarted . It should be done automatically on the first visit.

clobrano:
I thought it was only a way to get some example data to start working on something.
I will have a look on how to do it automatically on container's creation.

gimite:
Yeah but anyway you need some repositories to test anything, and I believe setup_datastore() also performs some global initialization. It's required also when you create a new real AppEngine instance (not local one) for testing. So I prefer a solution in AppEngine code, not limited to a local instance in Docker container.

clobrano:
I'm not sure I have understood. Is it that you would prefer that running the web server the first time will call somehow setup_datastore programmatically?

I found out that there is the possibility to run both the Person Finder's server AND another script that calls setup_datastore the first time the Docker container is run. There still a problem tough, since the Person Finder's server is not run in localhost (otherwise it won't be visible in host's browser) and then the access to db would require a password (root is fine) I am not able to pass it automatically, yet.

gimite:
> I'm not sure I have understood. Is it that you would prefer that running the web server the first time will call somehow setup_datastore programmatically?

Yes.

> I found out that there is the possibility to run both the Person Finder's server AND another script that calls setup_datastore the first time the Docker container is run. 

I prefer to implement the auto setup inside our AppEngine code, not as part of Docker initialization. The reason is that we also sometimes run our test server in AppEngine, not localhost:
[...]

PR #133, opened 2015-08-20
Automate Environment setup #78

clobrano:
Hi,

this branch adds a *docker* folder to the project, containing all the files (and some utilities) to automatically create a Docker container to develop and test Personfinder on Docker

please let me know what you think about it

BR,
Carlo


gimite:
Thank you for the pull request! Looks interesting. I will try it and get back to you.

gimite:
I'm trying to set up Docker in my Ubuntu machine following:
http://docs.docker.com/linux/started/
But I couldn't run hello-world:

```
$ docker run hello-world
Post http:///var/run/docker.sock/v1.20/containers/1694d7945c57e828c0d72db07df16664d5509b9571681bd6637d0e2a473fec98/start: EOF.
* Are you trying to connect to a TLS-enabled daemon without TLS?
* Is your docker daemon up and running?
```

I will continue investigating. I'm also asking my coworkers to try it, to see if it's an issue of our corp machine environment or I just made some mistake.

clobrano:
Have you tried to add your user to the docker group?

   $ sudo usermod -aG docker ubuntu

(log-out and log-in again)

gimite:
Yeah I tried it but still it fails with the same message. It even doesn't run as root:

```
$ sudo docker run hello-world
Post http:///var/run/docker.sock/v1.20/containers/9907ada3354ce9c9c47cf2e8ab880ca5f798a1506f8d22cfed4d622169c2c95b/start: EOF.
* Are you trying to connect to a TLS-enabled daemon without TLS?
* Is your docker daemon up and running?
```

clobrano:
You might need to install `apparmor` or (sorry for the stupid suggestion) restart the docker daemon

gimite:
It looks apparmor is already installed. I tried restarting by "sudo service docker restart", but still the same error.

clobrano:
That is really weird. Which Ubuntu version are you using?

gimite:
It's 14.04.

clobrano:
I only tested on 15.04. I will set up a virtual machine to try and
reproduce your issue

clobrano:
Hi Hiroshi,

I just tried a vbox ubuntu 14.04.3. I didn't follow the guide, though, I
just installed docker from the repositories:

    $ sudo apt-get install docker.io
    $ sudo usermod -aG docker $USER

and that worked just fine. Could you try the same and tell me if that's ok
for you too?

gimite:
Thanks it finally worked! I will make some feedback via line comments.

docker/README.md
```diff
@@ -0,0 +1,46 @@
...
+-----------------------------
+
+1. Setup the environment for Docker (see Docker's webpage)
+2. From `personfinder/docker` folder, run
+
+    $ docker build -t <image-name> .
+
+    e.g.
+
+    $ docker build -t personfinder-image
```
gimite: You need a period at the end of the command.

docker/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+    This will create and run the container image.
+
+4. Run the Personfinder local server and test it from host's web browser (https://<container-ip-address>:8000
+
+    $ gae-run-app.sh
+
+5. To initialize the datastore, open one more Shell inside the running container with the following command
+
+    $ docker exec -it <container-name> /bin/bash
```
gimite: This doesn't seem to work with Docker 1.0.1, which is installed via apt-get. I guess docker exec is not supported yet in the version. Do you know if any workaround? I can run

    $ gae-run-app.sh &

but it's still a bit inconvenient.

docker/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+    $ ./run-container.sh <image-name> [<container-name>]
+
+    e.g.
+
+    $ ./run-container.sh personfinder-image personfinder-container  # Container's name is optional
+
+    This will create and run the container image.
+
+4. Run the Personfinder local server and test it from host's web browser (https://<container-ip-address>:8000
```
gimite: This doesn't work well with our corporate environment which has HTTP proxy. Is there a way to forward the port to a port in the host machine instead of accessing container-ip-address?

docker/README.md
clobrano: Sure:

    docker start <container-name>
    docker attach <container-name>

should do the same job as `exec` does

docker/README.md
clobrano: Yes, I have not such a configuration, so I can not try it, but I pushed a
new version of run-container.sh script that forwards the PF's ports to the
host. You should be able to connect to PF at the address `localhost:8000`

clobrano:
Hi @gimite ,

not sure about the next step here. Is there something you are waiting from me?

docker/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+    $ gae-run-app.sh
```
gimite: It would be nice if you clarify that this command must be run inside Docker container's shell launched by the previous ./run-container.sh command, not in the host's shell. Same for setup_datastore.sh below. Also, it may be better to use "#" instead of "$" for these because the prompt actually looks like this:

    root@62e7813ec752:/opt/personfinder#

gimite:
Sorry for late. I was busy and I didn't have time to try it again.

I tried it again today and it worked nicely. Great! I added one more comment for the README.

clobrano:
No worries, I'm glad you found it useful. I will commit the changes you proposed.

clobrano:
Changes committed

docker/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+    You will be able to access the server from the Host's web browser at https://<container-ip-address>:8000
+
+
+5. Initialize the datastore:
+
+    open one more SHELL inside the running container with the following command (Docker version 1.3.0 or higher,
+    for older Docker versions, docker attach command should do the job as well).
+
+        # docker exec -it <container-name> /bin/bash
```
gimite: Shouldn't this command be run on the host's shell, not on the container's shell?

clobrano: Yes, but the README says "open one more shell ... using the following command...". Do you think it is not clear enough?

gimite: Oh I see. You should use "$" instead of "#" for this line because it's a command running on the host. Also, it would be clearer if you say:

> open one more SHELL inside the running container by running the following command on the host's shell

clobrano:
I agree, changes committed

gimite:
Thanks! Looks good to me.

Before merging your code, can you read through:
https://github.com/google/personfinder/blob/master/CONTRIBUTING.md
and agree on this form?:
https://cla.developers.google.com/about/google-individual

Sorry for taking time, but it is required to accept patches to projects owned by Google. It is necessary mainly because you own the copyright to your changes, even after your contribution becomes part of our codebase, so we need your permission to use and distribute your code.

clobrano:
No worries, I understand.
Signed, by the way

gimite:
Thank you! Merged.

gimite:
FYI The docker image is now hosted on Docker Hub: https://hub.docker.com/r/gimite/personfinder/

clobrano:
Great! Thank you :+1:

Issue #123, opened 2015-08-13
Add explicit check box for "write" permission of API key in admin page


gimite:
Currently, filling the domain name for write access subsumes the "write" permission, but it's confusing. We should add an explicit check box for "write" permission.

kspviswa:
@gimite 

I would like to work on this issue.

gimite:
It would be great! To clarify, we should have both checkbox and the text field for the domain name. If the user checks the checkbox and leave the domain field empty, we should show an error.

kspviswa:
@gimite 

Thanks for the clarification. I shall work on it and raise a pull request for you to review.

PR #134, opened 2015-08-22
Contains the bugfix for issue #123

kspviswa:
Added an explicit **Write Permission** checkbox to work-along with **Domain write permission**.

Following are the highlights of this new feature.

*  **By default, *domain write permission*  field will be disabled and this will be toggled with *Write Permission* checkbox**

   * **Default position**
image

   * **Upon toggling the *Write permission* checkbox**
   image

   image

*  **If *Write permission* checkbox is *checked* & no input is given for *Write domain text field*, the field will be highlighted with *red* color and *error* will be *alerted*.**

    *  image

   *  image

*  **If *Write permission* is checked & *domain write value* is provided, the normal work flow goes-on.**

   *  image

Please review the changelist & approve this merge request if deemed fit.

### DIFF
--- a/app/resources/admin_api_key_form.html.template
+++ b/app/resources/admin_api_key_form.html.template
@@ -14,14 +14,11 @@
   function $(id) {
     return document.getElementById(id);
   }
-  function updateWriteDomain() {
+  function update_write_domain() {
     var write_field = $('write_permission');
-    var domain_write_field = $('domain_write_permission');
-    if(write_field.checked) {
-      $('domain_write_permission').disabled = false;
-    }
-    else {
-      $('domain_write_permission').disabled = true;
+    $('domain_write_permission').disabled = !write_field.checked;
+    if(!write_field.checked) {
+      $('domain_write_permission').value = "";
     }
   }
   function validate_fields() {
@@ -41,7 +38,7 @@
     // Check for explicit write permission
     var write_field = $('write_permission');
     var domain_write_field = $('domain_write_permission');
-    if(write_field.checked && domain_write_field.value.length == 0) {
+    if (write_field.checked && domain_write_field.value.length == 0) {
        alert('Enter proper value for domain name');
        domain_write_field.focus();
        $('domain_write_label').setAttribute('class', 'mandatory');
@@ -104,9 +101,9 @@
 	Permission flags
       </label>
       <div class="response">
-        <input type="checkbox" name="write_permission" id="write_permission" onChange="updateWriteDomain();">
+        <input type="checkbox" name="write_permission" id="write_permission" onChange="update_write_domain();">
         Write permission
-	<span class="warning">This permission should go with domain write permission.</span>
+	<span class="warning">This permission should go with the domain name above.</span>
       </div>
       <div class="response">
         <input type="checkbox" name="read_permission" id="read_permission"
--- a/app/resources/admin_api_key_form.html.template
+++ b/app/resources/admin_api_key_form.html.template
@@ -14,6 +14,16 @@
   function $(id) {
     return document.getElementById(id);
   }
+  function updateWriteDomain() {
+    var write_field = $('write_permission');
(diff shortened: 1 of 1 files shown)


app/resources/admin_api_key_form.html.template
```diff
@@ -14,6 +14,16 @@
   function $(id) {
     return document.getElementById(id);
   }
+  function updateWriteDomain() {
```
gimite: Can you clear domain_write_field when the checkbox is unchecked? Otherwise, I believe, if the user fills in the domain name, uncheck the checkbox and submit, then the API key will end up having write permission.

Also, can you rename this to update_write_domain to be consistent with other method names?

app/resources/admin_api_key_form.html.template
```diff
@@ -14,6 +14,16 @@
...
+    var write_field = $('write_permission');
+    var domain_write_field = $('domain_write_permission');
+    if(write_field.checked) {
```
gimite: This part can be simplified to:

    $('domain_write_permission').disabled = !write_field.checked;

app/resources/admin_api_key_form.html.template
```diff
@@ -28,6 +38,17 @@
         return false;
       }
     }
+    // Check for explicit write permission
+    var write_field = $('write_permission');
+    var domain_write_field = $('domain_write_permission');
+    if(write_field.checked && domain_write_field.value.length == 0) {
```
gimite: Can you add a whitespace between "if" and "("?

app/resources/admin_api_key_form.html.template
```diff
@@ -82,6 +104,11 @@
 	Permission flags
       </label>
       <div class="response">
+        <input type="checkbox" name="write_permission" id="write_permission" onChange="updateWriteDomain();">
+        Write permission
+	<span class="warning">This permission should go with domain write permission.</span>
```
gimite: It would be clearer to say:

    This permission should go with the domain name above.

Actually it may be even better to move the domain name text box just below the "Write permission" checkbox (i.e., above "Read permission" checkbox). But it's up to you. We can do that later.

gimite:
Thanks for the pull request! Overall looks good to me. Left some line comments. Can you address these?

kspviswa:
@gimite 

Thanks for the review comments. I had addressed those comments and updated the pull request. Please review the same

app/resources/admin_api_key_form.html.template
```diff
@@ -14,6 +14,13 @@
   function $(id) {
     return document.getElementById(id);
   }
+  function update_write_domain() {
+    var write_field = $('write_permission');
+    $('domain_write_permission').disabled = !write_field.checked;
+    if(!write_field.checked) {
```
gimite: Can you put a whitespace between "if" and "("?

app/resources/admin_api_key_form.html.template
```diff
@@ -82,6 +101,11 @@
 	Permission flags
       </label>
       <div class="response">
+        <input type="checkbox" name="write_permission" id="write_permission" onChange="update_write_domain();">
+        Write permission
+	<span class="warning">This permission should go with the domain name above.</span>
```
gimite: One more whitespace indent here, to align with the previous line.

gimite:
Made a few more minor style comments.

kspviswa:
@gimite 

Addressed the white space change in Line number . In lines 104-106, not sure what are you referring about whitespace. Lines 104-106 start at 9th character from beginning of the line. Let me know what should be done here.

gimite:
Oh I found that line 104-105 uses whitespaces but line 106 uses a tab character for indent. Can you use whitespaces? That was why it looked they don't align in Github.

kspviswa:
Edited app/resources/admin_api_key_form.html.template to use whitespace instead of TAB @ line number 106.

gimite:
Looks good to me. Merged. Thanks!

Issue #67, opened 2015-07-06
Free-text search (search by person name + location)


gimite:
It would be nice if users can search by person name + location (city, town, etc.). To implement it, we probably need to switch to full-text search API.

gimite:
Copying @skywhale 's comment from another issue:

This should be relative simple to implement with the new Search API provided by GAE.
https://developers.google.com/appengine/docs/python/search/

Please note that we probably need to make sure a search query matches at least a part of person name; otherwise, the feature could allow querying for a neighborhood name and scrape a list of people there, which we agreed is a capability we don't want to provide with ordinary users to protect users' privacy.

gimite:
@prazuma will work on this.

PR #136, opened 2015-08-25
Allowed full text search by person name and location #67

prazuma:


app/full_text_search.py
```diff
@@ -22,6 +22,15 @@
 
 # The index name for full text search
 PERSON_FULL_TEXT_INDEX_NAME = 'person_information'
+PERSON_LOCATION_FULL_TEXT_INDEX_NAME = 'person_location_information'
```
gimite: Can you add a short comment which explains the difference between PERSON_FULL_TEXT_INDEX_NAME and PERSON_LOCATION_FULL_TEXT_INDEX_NAME?

app/full_text_search.py
```diff
@@ -22,6 +22,15 @@
...
+
+def create_or_query(query_txt):
```
gimite: Can you add a document comment for this function?

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py
```diff
@@ -20,8 +20,26 @@
...
+
+    Returns:
+        query_word OR query_word OR ...
+    """
+    query_words = query_txt.split(' ')
+    query_string = ''
+    for word in query_words:
+        if (word != ''):
+            query_string += word + ' OR '
+    return query_string[:-4]
```
kevinalle: this is a bit hacky. you can use "join"
`return ' OR '.join([word for word in query_words if word != '']`

app/full_text_search.py
```diff
@@ -20,8 +20,22 @@
 
 import model
 
-# The index name for full text search
+# The index name for full text search. This index contains name.
```
gimite: "name" is ambiguous. Can you say "person name" instead? Same below.

app/full_text_search.py
```diff
@@ -20,8 +20,22 @@
...
+    """
+    Creates query for person_information index.
+    Args:
+        query_txt: Search query
+
+    Returns:
+        query_word OR query_word OR ...
+    """
+    query_words = query_txt.split(' ')
+    return ' OR '.join([word for word in query_words if word != ''])
```
gimite: Just "if word" instead of "if word != ''" (per style guide)

gimite: What will happen if one of the word in the query_txt is "OR", or the word contains double-quote(") or colon(:)?

app/full_text_search.py

app/full_text_search.py
```diff
@@ -42,13 +56,21 @@ def search(repo, query_txt, max_results):
...
+        name=PERSON_LOCATION_FULL_TEXT_INDEX_NAME)
     options = appengine_search.QueryOptions(
         limit=max_results,
         returned_fields=['record_id'])
-    index_results = index.search(appengine_search.Query(
-        query_string=query_txt, options=options))
+    or_query = create_or_query(query_txt) + ' AND (repo: ' + repo + ')'
+    person_index_results = person_index.search(appengine_search.Query(
+        query_string=or_query, options=options))
+    query_txt += ' AND (repo: ' + repo + ')'
```
gimite: Modifying a parameter can be a cause of bugs. (e.g., If you happen to put this line before line 65, it breaks.) Can you define a new variable instead of modifying query_txt?

app/full_text_search.py
```diff
@@ -42,13 +56,21 @@ def search(repo, query_txt, max_results):
...
+    person_index = appengine_search.Index(name=PERSON_FULL_TEXT_INDEX_NAME)
+    person_location_index = appengine_search.Index(
+        name=PERSON_LOCATION_FULL_TEXT_INDEX_NAME)
     options = appengine_search.QueryOptions(
         limit=max_results,
         returned_fields=['record_id'])
-    index_results = index.search(appengine_search.Query(
-        query_string=query_txt, options=options))
+    or_query = create_or_query(query_txt) + ' AND (repo: ' + repo + ')'
+    person_index_results = person_index.search(appengine_search.Query(
```
gimite: As discussed, use larger max_results for this search.

app/full_text_search.py
```diff
@@ -20,8 +20,22 @@
...
+    Args:
+        query_txt: Search query
+
+    Returns:
+        query_word OR query_word OR ...
+    """
+    query_words = query_txt.split(' ')
+    return ' OR '.join([word for word in query_words if word != ''])
 
 def search(repo, query_txt, max_results):
```
gimite: Please document the condition when the person record matches the query.

app/full_text_search.py
If query_txt contains these words, can't search. -> fixed?(ignore these words)

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py
```diff
@@ -42,15 +56,34 @@ def search(repo, query_txt, max_results):
     results = []
     if not query_txt:
         return results
-    index = appengine_search.Index(name=PERSON_FULL_TEXT_INDEX_NAME)
-    query_txt += ' AND (repo: ' + repo + ')'
+
+    query_txt = re.sub('|:|OR|AND|"', '', query_txt)
```
gimite: How about putting each word into quote e.g., [hoge OR foo] -> ["hoge" "OR" "foo"]? Will it work? For double-quotes, you may just remove them (as you already do) or somehow escape them.

Also, please add a comment why you do this.

app/full_text_search.py
```diff
@@ -20,8 +20,22 @@
...
+    """
+    Creates regular expression.
+    Args:
+        query_txt: Search query
+
+    Returns:
+        query_word | query_word | ...
+    """
+    query_words = query_txt.split(' ')
+    return '|'.join([word for word in query_words if word])
```
gimite: Apply escaping to the word. e.g., | -> \|
There should be a function for it.

app/full_text_search.py
```diff
@@ -20,8 +20,22 @@
 
 import model
 
-# The index name for full text search
-PERSON_FULL_TEXT_INDEX_NAME = 'person_information'
+# The index name for full text search.
+# This index contains person name and location.
+PERSON_LOCATION_FULL_TEXT_INDEX_NAME = 'person_location_information'
+
+def make_regexp(query_txt):
```
gimite: Use clearer method name and description e.g., "make_or_regexp"

app/full_text_search.py
```diff
@@ -42,15 +56,34 @@ def search(repo, query_txt, max_results):
...
+        appengine_search.Query(
+            query_string=and_query, options=options))
+
+    index_results = []
+    regexp = make_regexp(query_txt)
+    for document in person_location_index_results:
+        for field in document.fields:
+            if field.name == 'names':
+                names = field.value
+            else:
```
gimite: Explicitly check if it's 'record_id'. Otherwise it can cause a bug when you add new field to returned_fields.

app/full_text_search.py
```diff
@@ -42,15 +56,34 @@ def search(repo, query_txt, max_results):
...
+                id = field.value
+
+        pattern = re.compile(regexp, re.I)
```
gimite: How about calling re.compile inside make_regexp i.e., let make_regexp return compiled regexp instead?

app/full_text_search.py
```diff
@@ -42,15 +56,34 @@ def search(repo, query_txt, max_results):
...
+        match = pattern.search(names)
+        if match:
+            index_results.append(id)
+
+    for id in index_results:
```
gimite: You can simplify this part by:

    return [model.Person.get_by_key_name(repo + ':' + id) for id in index_results]

app/full_text_search.py
```diff
@@ -75,14 +108,28 @@ def add_record_to_index(person):
...
+        name=PERSON_LOCATION_FULL_TEXT_INDEX_NAME)
+    name_params = [person.given_name,
+                   person.family_name,
+                   person.full_name,
+                   person.alternate_names]
+    names =  ':'.join([name for name in name_params if name])
+    person_location_index.put(create_document(
         record_id=person.record_id,
         repo=person.repo,
+        names = names,
```
gimite: Do not put whitespace around "=".

app/full_text_search.py
gimite: It looks this has not been done yet.

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

works

app/full_text_search.py

app/full_text_search.py
```diff
@@ -20,19 +20,50 @@
...
+    """
+    Enclosesd query_txt in double quotes.
+    Args:
+        query_txt: Search query
+
+    Returns:
+        "query_words" + " " + "query_words" + " " + ...
+    """
+    query_words = query_txt.split(' ')
+    enclosed_query = '"'
```
gimite: It would be easier to read if you write line 50-52 in one line:

    return '"' + '" "'.join([word for word in query_words if word]) + '"'

app/full_text_search.py
```diff
@@ -20,19 +20,50 @@
...
+    Returns:
+        query_word | query_word | ...
+    """
+    query_words = query_txt.split(' ')
+    regexp = '|'.join([re.escape(word) for word in query_words if word])
+    return re.compile(regexp, re.I)
+
+def enclose_in_double_quotes(query_txt):
+    """
+    Enclosesd query_txt in double quotes.
```
gimite: s/Enclosesd/Encloses/

Also you should say "Encloses each word in query_txt in double quotes." Otherwise it looks the whole query_txt is enclosed in a single quote.

app/full_text_search.py
```diff
@@ -20,19 +20,50 @@
...
+    Args:
+        query_txt: Search query
+
+    Returns:
+        "query_words" + " " + "query_words" + " " + ...
```
gimite: This looks confusing. It looks it just joins the words with whitespaces. It can be e.g.,

    '"query_word1" "query_word2" ...'

app/full_text_search.py
```diff
@@ -42,25 +73,40 @@ def search(repo, query_txt, max_results):
...
-        query_string=query_txt, options=options))
-    for document in index_results:
-        id = document.fields[0].value
-        result = model.Person.get(repo, id, filter_expired=True)
-        if result:
-            results.append(result)
-    return results
+        returned_fields=['record_id', 'names'])
+
+    and_query = enclose_in_double_quotes(query_txt) + ' AND (repo: ' + repo + ')'
```
gimite: Add a comment which explains why enclose_in_double_quotes() is needed.

app/full_text_search.py
```diff
@@ -42,25 +73,40 @@ def search(repo, query_txt, max_results):
     results = []
     if not query_txt:
         return results
-    index = appengine_search.Index(name=PERSON_FULL_TEXT_INDEX_NAME)
-    query_txt += ' AND (repo: ' + repo + ')'
+
+    # escape double quote for enclose each query_txt
```
gimite: It is not "escape". Escaping is something like:

    " -> \"

You should say "remove double quotes" instead. The whole sentence can be e.g.,
"Remove double quotes so that we can safely apply enclose_in_double_quotes()."

app/full_text_search.py
```diff
@@ -42,25 +73,40 @@ def search(repo, query_txt, max_results):
...
+        for field in document.fields:
+            if field.name == 'names':
+                names = field.value
+            if field.name == 'record_id':
+                id = field.value
 
+        if regexp.search(names):
+            index_results.append(id)
+
+    return [model.Person.get(repo, id, filter_expired=True)
```
gimite: This code can call model.Person.get() twice for a same record, which is bad. Can you avoid it? You may give up using [... for ... in ...] syntax.

app/full_text_search.py
```diff
@@ -20,19 +20,50 @@
...
 
-# The index name for full text search
-PERSON_FULL_TEXT_INDEX_NAME = 'person_information'
+# The index name for full text search.
+# This index contains person name and location.
+PERSON_LOCATION_FULL_TEXT_INDEX_NAME = 'person_location_information'
+
+def make_or_regexp(query_txt):
+    """
+    Craetes compiled regular expression for or search.
```
gimite: s/Craetes/Creates/

Also, "for or search" is a bit hard to read. Should be e.g., "for OR search"

app/full_text_search.py
```diff
@@ -42,25 +73,40 @@ def search(repo, query_txt, max_results):
     results = []
     if not query_txt:
         return results
```
gimite: Maybe this is the only place where "results" is used? If so, just return [] instead.

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py
```diff
@@ -20,47 +20,94 @@
...
+    return re.compile(regexp, re.I)
+
+def enclose_in_double_quotes(query_txt):
+    """
+    Encloses each word in query_txt in double quotes.
+    Args:
+        query_txt: Search query
+
+    Returns:
+        '"query_words" "query_words" ...'
```
gimite: It should be "query_word" or "query_word1" or something.

app/full_text_search.py
```diff
@@ -20,47 +20,94 @@
...
     options = appengine_search.QueryOptions(
         limit=max_results,
-        returned_fields=['record_id'])
-    index_results = index.search(appengine_search.Query(
-        query_string=query_txt, options=options))
-    for document in index_results:
-        id = document.fields[0].value
+        returned_fields=['record_id', 'names'])
+
+    # enclose_in_double_quotes used for escaping.
```
gimite: Be more precise. What do you want to avoid?

app/full_text_search.py
```diff
@@ -20,47 +20,94 @@
...
+        - []: If query_txt doesn't contain a part of person name
 
     Raises:
         search.Error: An error occurred when the index name is unknown
                       or the query has syntax error.
     """
     #TODO: Sanitaize query_txt
-    results = []
     if not query_txt:
         return results
```
gimite: You must return [] instead.

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

gimite:
LGTM after the change you just made.

Issue #142, opened 2015-08-27
500 Error when address is multiple lines


tingar:
When adding information about a person, if the address is more than 1 line, submitting the form results in a 500 error.

Error from the console log:
```
ERROR    2015-08-27 04:16:45,658 webapp2.py:1528] Property last_known_location is not multi-line
Traceback (most recent call last):

  File "personfinder/app/main.py", line 504, in serve
    getattr(handler, request.method.lower())()  # get() or post()
  File "personfinder/app/view.py", line 245, in post
    photo_url=photo_url)
  File "personfinder/app/model.py", line 223, in create_original
    return cls(key_name=repo + ':' + record_id, repo=repo, **kwargs)

BadValueError: Property last_known_location is not multi-line
INFO     2015-08-27 04:16:45,662 module.py:808] default: "POST /japan/view HTTP/1.1" 500 228
```

gimite:
Thanks for the report. Fixed.

PR #145, opened 2015-08-27
A fix for https://github.com/google/personfinder/issues/142

hidetokazawa:
This change allows multiline text in the last known location.
I am not sure how this could affect existing records. If this isn't good,
please let me know. I will try to escape line separaters instead.


gimite:
LGTM. Merged.

PR #146, opened 2015-08-27
Fix issue 128 - Mark duplicates table is too wide

iniju:


iniju:
It affects the way the "Back to start" link appears. Hopefully you can see the attached images.

With end-multi-columns:
with end-multi-columns

Without end-multi-columns:
without end-multi-columns

gimite:
How about omitting "end" class and just specify "column"? Then it won't introduce "float: left;", so I believe you don't need end-multi-columns.

iniju:
Yeap, that works, I've updated the pull request.

gimite:
LGTM. Thanks! Merged.

PR #151, opened 2015-08-27
Highlight newly shown duplicate notes.

### LINKED ISSUE #80: "Show who marked these duplicates" is very confusing
When I mark 2 records as duplicates, then it adds to each record the following text.
"Possible duplicates
PERSON NAME 
 (links_to_record_)
Show who marked these duplicates"

"Show who marked these duplicates" is a link and when I click on it, I expect to see information about the person.
What happens is actually, that we do display a note on the right side of the record. But it is hard to notice, for a while I though it just linked me to the duplicated record.

Maybe we should just highlight the new notes with a color?

tingar:
Highlights notes revealed by the "Show who marked these duplicates" link with a yellow background, to make them more visible.

duplicate

Fixes #80.

### DIFF
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -447,6 +447,10 @@ img.icon {
   color: #888;
 }
 
+.view.note.duplicate {
+  background: #fe8;
+}
+
 .note .author {
   font-weight: bold;
   color: #000;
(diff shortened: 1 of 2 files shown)


gimite:
LGTM. Merged.

PR #141, opened 2015-08-27
On mobile, get the user's current location from GPS signal using HTML5 API

skywhale:
We probably can not publish the precise location, and some sort of redaction may be necessary, but for the initial demo, we can start with precise latlng and reverse geocoded address.


gimite:
Let's discuss if it's OK to integrate this with Nobu-san in the next meeting.

gimite:
LGTM. Merged.

Issue #138, opened 2015-08-26
Do not show expired records in full text search mode


gimite:
Also, if the user delete the record and then undo the deletion, it must show up again.

@prazuma will work on this.

prazuma:
done

PR #153, opened 2015-08-28
Fixed full-text search (do not show expired record)#138

prazuma:


app/full_text_search.py
```diff
@@ -51,7 +58,9 @@ def search(repo, query_txt, max_results):
         query_string=query_txt, options=options))
     for document in index_results:
         id = document.fields[0].value
-        results.append(model.Person.get_by_key_name(repo + ':' + id))
+        result = get_record_from_datastore(repo, id)
```
gimite: Use Person.get() instead:
https://github.com/google/personfinder/blob/master/app/model.py#L209

Person.get() also has feature to check if it's expired.

app/full_text_search.py
```diff
@@ -51,7 +58,9 @@ def search(repo, query_txt, max_results):
...
+        if result:
+            results.append(get_record_from_datastore(repo, id))
```
gimite: Reuse "result" instead of calling get_record_from_datastore() twice.

app/full_text_search.py

gimite:
LGTM. Merged.

prazuma:
(∩´∀｀)∩ﾜｰｲ

Issue #101, opened 2015-07-28
Use Babel to localize date/time format


gimite:
Babel seems better than the one currently used.

e.g., Django formatter (currently used) doesn't support Bulgarian where the date format should be:
DD.MM.YYYY г. 

Not sure if Babel supports Bulgarian, though.

PR #154, opened 2015-08-28
Format datetime by babel

yamaguchi-am:
For Issue #101 .
https://github.com/google/personfinder/issues/101

This change will resolve the original issue with Bulgarian:
> e.g., Django formatter (currently used) doesn't support Bulgarian where the date format should be:
> DD.MM.YYYY г.

However it will change formatting for some other languages, depending on the default style by Babel.
For example, in Japanese:
Django: 2014年10月17日20:43:50 
Babel: 2014/10/17 20:43:50


yamaguchi-am:
Done.

gimite:
LGTM. Merged. Thanks!

PR #155, opened 2015-08-30
UI Refresh

Jessie-Newman:
Moved add notes about a person to a new page. Lots of CSS changes on the view, create, and notes pages. Reordered a lot of fields.

You can view this change at jessien-pf-6.appspot.com


app/resources/add_note.html.template
```diff
@@ -18,35 +18,116 @@
...
-    {% if not markdup %}
-      {% trans "Tell us the status of this person" %}
-    {% else %}
-      {% trans "Mark records as duplicate" %}
-    {% endif %}
-  </h2>
+  <div class="header section">
+    <h1>
+      {% if not markdup %}
+        {% trans "Status of " %}{{person.full_name}}
```
gimite: This part should be instead:

    {% blocktrans %}Status of {{person.full_name}}{% endblocktrans %}

Because it may be in different order in different languages.

Also, fall back to "Status of the person" if person.full_name is missing. This can happen when you create a new person record:
https://screenshot.googleplex.com/UxYBwLh2g4e

skywhale:
Thanks for the changeset. I'm going to review it now. The demo looks great BTW!!

app/resources/status.html.template
```diff
@@ -0,0 +1,83 @@
+{# Copyright 2010 Google Inc.  Licensed under the Apache License, Version   #}
```
skywhale: I'm trying to think what's the good name for this file. Since PFIF spec specifically calls out the concept of Person and Note, I think we should be using note as part of the file name. What about renaming this file to add_note.html.template and renaming the existing add_note.html.template to add_note_base.html.template? I believe the latter is now used only by included in other files?

app/status.py
```diff
@@ -0,0 +1,289 @@
+#!/usr/bin/python2.7
```
skywhale: I think this is introducing a lot of duplicate code in the codebase. We should be either:
  - removing the duplicate code from create.py and make create.py import some shared functionality from this file, or vise verse, or
  - add some conditions to create.py so it can also handle status requests.

Also same as the above comment. Please rename this to add_note.py

skywhale: If you want to focus on the UI refresh only in this changeset, it's also okay to leave the code as is with a TODO to cleanup duplicate code in this file at a later iteration and file an issue in github so we don't forget it.

app/resources/status.html.template
skywhale: BTW, you can find PFIF spec at the following site. Person Finder is a reference implementation of the spec.

http://zesty.ca/pfif/1.4/

skywhale:
Hi Jessie,

Let's first discuss how we want to proceed with these changes. I noticed that the new style is only applied to a subset of pages. Merging this change into the master branch will put it at a state where we can not push it to the production Person Finder. We have roughly two options:

1) Keep working on this branch, until all the pages are reflected with the new styles, and the important user flows are all tested with desktop and mobile browsers.

2) Make the new style guarded with a flag. We can start merging pieces of code into the master branch with the flag originally turned off, and once we have all the changes in and tested, we can finally flip the flag.

What would you prefer?

Jessie-Newman:
I pushed fixes to your comments. And UI changes to the "I am looking
for"/"I have information" screen & the search screens.

app/add_note.py
```diff
@@ -0,0 +1,291 @@
...
+        for website in handler.config.profile_websites or []:
+            if ('url_regexp' in website and
+                re.match(website['url_regexp'], profile_url)):
+                profile_page = add_profile_icon_url(website, handler)
+                profile_page['url'] = profile_url
+                break
+        profile_pages.append(profile_page)
+    return profile_pages
+
+class Handler(BaseHandler):
```
skywhale: two blank lines between top-level functions and classes

skywhale:
I'm sorry for stepping back, but is there an existing issue or document that describes the motivation for moving the add_note form to a separate page?

skywhale:
Can you also apply the new button styles (primary action + secondary actions) to the subscription pages? subscribe.html, extend.html etc.

kevinalle@ will be doing the secondary review for the styling. He's been working on UI refresh and optimization for mobile phones, so he should have more holistic view for reviewing your CL.

app/add_note.py
```diff
@@ -0,0 +1,291 @@
...
+from django.utils.translation import ugettext as _
+from urlparse import urlparse
+
+# TODO(jessien): Clean up duplicate code here and in create.py.
+
+# how many days left before we warn about imminent expiration.
+# Make this at least 1.
+EXPIRY_WARNING_THRESHOLD = 7
+
+def get_profile_pages(profile_urls, handler):
```
skywhale: This function is not needed in this file. Please see my other comment below.

app/add_note.py
```diff
@@ -0,0 +1,291 @@
+#!/usr/bin/python2.7
+# Copyright 2010 Google Inc.
```
skywhale: 2015

app/add_note.py
```diff
@@ -0,0 +1,291 @@
...
+from utils import *
+from detect_spam import SpamDetector
+import extend
+import reveal
+import subscribe
+
+from django.utils.translation import ugettext as _
+from urlparse import urlparse
+
+# TODO(jessien): Clean up duplicate code here and in create.py.
```
skywhale: Can you create a Github issue and link it here, so we don't forget it?

app/add_note.py
```diff
@@ -0,0 +1,291 @@
...
+            person.provider_name = person.get_original_domain()
+
+        sanitize_urls(person)
+        for note in notes:
+            sanitize_urls(note)
+
+        if person.profile_urls:
+            person.profile_pages = get_profile_pages(person.profile_urls, self)
+
+        self.render('add_note.html',
```
skywhale: You can remove a lot of code from this file. For example, only record_id, notes_disabled and full_name properties are used for person object. We don't need to set any other fields e.g. profile_pages and provider_name. Other variables passed to add_note.html are also not used e.g. notes, expiration_days, linked_person_info and so forth. Could you go through them and clean them up? I feel that we can significantly slim down this file.

app/resources/add_note.html.template
```diff
@@ -8,281 +8,76 @@
...
+    <div class="view">
+
+      <div class="column centered">
+        {% if person.notes_disabled %}
+          <div class="mandatory">
+            {% trans "The author has disabled notes on this record." %}
+            <p>
+            <input type="button"
+              value="{% trans "Enable notes on this record" %}"
+              onclick="javascript:location.href='{{enable_notes_url}}'"
```
skywhale: Would this button work for old tier2 phones that do not support javascript? Since there are many of those phones still in emerging markets, where Person Finder is used, we should gracefully fallback when javascript is not supported at least for core user flows.

app/resources/app-base.html.template
```diff
@@ -13,7 +13,12 @@
 {% block title %}{{block.super}}{% endblock title %}
 
 {% block logo %}
-  <a href="{{env.global_url}}"><img src="{{env.global_url}}/logo.png" alt=""></a>
+  <a href="{{env.global_url}}">
+    <div class="logo">
```
skywhale: Do you mind moving the logo change to a separate changeset? This particular one will need a bit more discussion, involving PM.

app/view.py
```diff
@@ -110,7 +110,14 @@ def get(self):
         feed_url = self.get_url(
             '/feeds/note',
             person_record_id=self.params.id,
-            repo=self.repo)
+            repo=self.repo)   
+        add_note_url = self.get_url(
+            '/add_note',
+             id=self.params.id,
+             role=self.params.role,
```
skywhale: I don't think we need any url parameters other than "id"?

app/resources/view.html.template
```diff
@@ -290,55 +289,53 @@
...
-            <div>
-              <input type="button"
-                value="{% trans "Delete this record" %}"
-                onclick="javascript:location.href='{{delete_url}}'"
-                id="delete_btn">
-            </div>
-          {% endif %}
+            {% if env.enable_captcha %}
+              <div class="links section">
+                <a id="delete-btn" href={{delete_url}}>
```
skywhale: please have double quotations around {{delete_url}}

app/resources/view.html.template
```diff
@@ -290,55 +289,53 @@
...
-            <div>
-              <input type="button"
-                value="{% trans "Subscribe to updates about this person" %}"
-                onclick="javascript:location.href='{{subscribe_url}}'"
-                id="subscribe_btn">
-            </div>
-          {% endif %}
+            {% if env.enable_captcha %}
+              <div class="links section">
+                <a id="subscribe_btn" href={{subscribe_url}}>
```
skywhale: please have double quotations around {{subscribe_url}}

app/resources/view.html.template
```diff
@@ -290,55 +289,53 @@
...
-            <div>
-              <input type="button"
-                value="{% trans "Disable notes on this record" %}"
-                onclick="javascript:location.href='{{disable_notes_url}}'"
-                id="disable_notes_btn">
-            </div>
-          {% endif %}
+            {% if env.enable_captcha and person.author_email and not person.notes_disabled %}
+              <div class="links section">
+              <a id="disable_notes_btn" href={{disable_notes_url}}>
```
skywhale: ditto

app/resources/view.html.template
```diff
@@ -290,55 +289,53 @@
...
-            {% if expiration_days %}
-              <div class="warning">
-                {% if expiration_days > 1 %}
-                  {% blocktrans %}
-                    Warning: this record will expire in less than {{expiration_days}} days.
-                  {% endblocktrans %}
-                {% else %}
-                  {% blocktrans %}
-                    Warning: this record will expire in less than a day.
+          <div class="links section">
```
skywhale: I don't think you need class="links" here? It seems each the link below is surrounded by <div class="links">

app/resources/view.html.template
```diff
@@ -265,7 +264,7 @@
 
             {% if linked_person_info %}
               <div class="other section">
-                <h3>{% trans "Possible duplicates" %}</h3>
+                <h2>{% trans "Possible Duplicates" %}</h2>
```
skywhale: Other <h2> titles are sentence cased. We should keep this consistent, too, unless you feel strongly about changing it to title cased. Changing message will require re-translation using Translation Console, which will take weeks for minor languages. Unless we have a strong reason, we shouldn't change messages.

app/resources/view.html.template
```diff
@@ -181,7 +178,7 @@
             {% endif %}
 
             <div class="source section">
-              <h3>{% trans "Source of this record" %}</h3>
+              <h2>{% trans "Source of this Record" %}</h2>
```
skywhale: see my comment below

app/resources/view.html.template
```diff
@@ -54,32 +54,31 @@
...
             </h2>
 
             <div class="name section">
-              <h3>{% trans "Name" %}</h3>
-              <div class="field">
-                <span class="label">{% trans "Full name" %}:</span>
-                <span class="value">{{person.full_name_list|join:"<br>"}}</span>
-              </div>
+              <h2>{% trans "Identifying Information" %}</h2>
+              {% if person.full_name_list|length > 2 %}
```
skywhale: Do you mind helping me understand why you are checking ">2" ?

app/resources/create.html.template
```diff
@@ -31,169 +31,161 @@
...
-                  </label>
-                </span>
-                <span class="value">
-                  <input name="family_name" id="family_name"
-                      class="medium-text-input"
-                      value="{{params.family_name}}"
-                      autocomplete="off">
-                </span>
+            <h2>
+              Identifying Information
```
skywhale: Need to wrap in {% trans %}

app/resources/add_note.html.template
skywhale: NVM, this is pre-existing.

app/resources/view.html.template
skywhale: I noticed that you are trying to change all the titles to title cased. What was the motivation?

app/resources/view.html.template
Jessie-Newman: This whole full name section doesn't need to be here since the full name is now at the top. I must have made a mistake while doing this and then not caught it since it will never be greater than 2.
Removed.

app/resources/view.html.template
Jessie-Newman: I thought it looked nicer in headers. I didn't realize there would be translation implications. I'll revert them but I do think it's something to consider.

app/resources/app-base.html.template
Jessie-Newman: Sure. Removed.

Jessie-Newman:
Fixed your comments. I'm not sure how to respond directly to your comments. I ended up editing the text when I wanted to respond.

app/add_note.py
```diff
@@ -0,0 +1,191 @@
...
+
+from django.utils.translation import ugettext as _
+from urlparse import urlparse
+
+# TODO(jessien): Clean up duplicate code here and in create.py.
+# https://github.com/google/personfinder/issues/157
+
+# how many days left before we warn about imminent expiration.
+# Make this at least 1.
+EXPIRY_WARNING_THRESHOLD = 7
```
skywhale: not used

skywhale:
Thank you :)

Looks good overall.

Kevin, could you proceed with the UI review?

I know it's a headache, but it's important so that other people don't break your code. Can you please update server_tests? They are in tests/server_test_cases/{person_note,photo,read_only}_tests.py. I suppose some of the tests are currently broken by your change? Please fix them and add a test that verifies behavior in add_note page.

Jessie-Newman:
Updated tests. We already have testing for the add note section so I've modified the existing tests to navigate to the new page before testing it. I believe the existing tests will test all the behavior on the new page.

I was having some appengine difficulties so I had to set up a new server. Please use jessien-pf-8.appspot.com for the UI review.

skywhale:
Kevin, please continue the style review. Thanks :)

gimite:
"Family name" field doesn't say (required) but in red, which is confusing. It's actually required, right? Can you add (required)?
https://screenshot.googleplex.com/SNbHGbp3Hfd
http://jessien-pf-8.appspot.com/haiti/create?add_note=&age=&alternate_family_names=&alternate_given_names=&author_email=&author_made_contact=&author_name=&author_phone=&clone=&confirm=&contact_email=&contact_name=&content_id=&context=&cursor=&date_of_birth=&description=&domain_write_permission=&dupe_notes=&email_of_found_person=&error=&family_name=ichikawa&given_name=hiroshi&home_city=&home_country=&home_neighborhood=&home_postal_code=&home_state=&home_street=&id=&id1=&id2=&id3=&key=&last_known_location=&max_results=&note_photo_url=&omit_notes=&operation=&organization_name=&person_record_id=&phone_of_found_person=&photo_url=&profile_url1=&profile_url2=&profile_url3=&query=&query_type=&referrer=&role=provide&search_engine_id=&sex=&signature=&skip=&small=&source=&source_date=&source_name=&source_url=&status=&style=&subscribe=&subscribe_email=&suppress_redirect=&target=&text=&ui=&utcnow=

app/add_note.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def get(self):
+        # Check the request parameters.
+        if not self.params.id:
+            return self.error(404, 'No person id was specified.')
+        try:
+            person = Person.get(self.repo, self.params.id)
+        except ValueError:
+            return self.error(404,
+                _("This person's entry does not exist or has been deleted."))
```
kevinalle: use " and ' consistently throughout the file

app/add_note.py
```diff
@@ -0,0 +1,191 @@
...
+# Make this at least 1.
+EXPIRY_WARNING_THRESHOLD = 7
+
+
+class Handler(BaseHandler):
+
+    def get(self):
+        # Check the request parameters.
+        if not self.params.id:
+            return self.error(404, 'No person id was specified.')
```
kevinalle: don't these need to be translated?

app/add_note.py
```diff
@@ -0,0 +1,191 @@
...
+        if not self.params.author_name:
+            return self.error(
+                200, _('Your name is required in the "About you" section.  '
+                       'Please go back and try again.'))
+
+        if (self.params.status == 'is_note_author' and
+            not self.params.author_made_contact):
+            return self.error(
+                200, _('Please check that you have been in contact with '
+                       'the person after the earthquake, or change the '
```
kevinalle: should be something more general than "earthquake"

app/add_note.py
```diff
@@ -0,0 +1,191 @@
...
+        if self.params.subscribe:
+            return self.redirect('/subscribe',
+                                 id=person.record_id,
+                                 subscribe_email=self.params.author_email,
+                                 context='add_note')
+
+        # Redirect to this page so the browser's back button works properly.
+        self.redirect('/add_note', id=self.params.id, query=self.params.query)
+
+    def add_fields_to_notes(self, note):
```
kevinalle: is this function unused?

app/resources/add_note.html.template
```diff
@@ -8,281 +8,76 @@
...
+        } else {
+          id_or_elem.style.display = hide_text;
+        }
+      }
+
+      function hide_unhide_note_contents(note_contents_id) {
+        var note = document.getElementById(note_contents_id + '-contents');
+        var hidden = note.style.display == 'none';
+        set_display(note, !hidden);
+        set_display('reveal-note', hidden);
```
kevinalle: Is there just one reveal-note for all notes? Shouldn't it be something like (note_contents_id + 'reveal-note')?

kevinalle: indeed, there seems to be a bug. when there are 2 hidden notes, revealing one changes the button on the other one too: https://screenshot.googleplex.com/qrvHSum2CHv.png

Jessie-Newman: This is the current behavior in the live app, though I do agree it's confusing. I think it would be good to discuss if we actually want to change this behavior (if a record has 50 notes, we probably don't want someone to have to enter a captcha for each note) and work on it in a separate pull request.

app/add_note.py
Jessie-Newman: This section was copied from view.py. Fixed

app/add_note.py
Jessie-Newman: This section was copied from view.py. Fixed

app/add_note.py
Jessie-Newman: This section was copied from view.py. 
Switched to using "disaster" which is consistent with the person finder homepage descrpition.

app/add_note.py
Jessie-Newman: Whoops, bad copy/paste. Fixed.

Jessie-Newman:
Added (required) to family name in [Given Name Family Name] view. Fixed minor text/translation issues. As for the click to reveal contact information, right now clicking on any of them reveals the contact information for all of them. I'm doing it this way because this is what we already do. But we should definitely consider how we can make the contact information reveal less confusing.

app/resources/add_note.html.template
kevinalle: i see, sgtm

kevinalle:
LGTM :) please push in a separate branch until the whole UI refresh is ready

gimite:
Merged this branch into new-ui branch.

Issue #135, opened 2015-08-25
Unable to register and search by kanji if "Require query words to have at least 2 characters" is selected


prazuma:
When "Require query words to have at least 2 characters" is selected, can't register record with given/family name with 2-letter kanji characters. It should be accepted since it says "at least 2 characters".

gimite:
Clarified the bug description. Since it says "require at least 2 characters", it would be fine to reject one-letter given/family names. But still it should not reject 2-letter names.

PR #156, opened 2015-08-31
Fix issue #135

ezbr:
Fixed https://github.com/google/personfinder/issues/135 by counting series of CJK characters as one word.


app/results.py
```diff
@@ -51,6 +51,19 @@ def is_possible_phone_number(query_str):
     return re.search(POSSIBLE_PHONE_NUMBER_RE,
         unicodedata.normalize('NFKC', unicode(query_str)))
 
+def max_word_length(query_words):
+    max_length = max(map(len, query_words))
+    # Count series of CJK characters as one word.
```
gimite: Are you doing this because each CJK character is treated as one word in query_words even if it originally doesn't have whitespace e.g., "山田" -> ["山", "田"]? Have you verified that behavior? If so, can you explain it in the comment here?

app/results.py
```diff
@@ -51,6 +51,19 @@ def is_possible_phone_number(query_str):
...
+    cjk_length = 0
+    for i in range(len(query_words)):
```
gimite: You can simplify this to "for word in query_words".

app/results.py
```diff
@@ -51,6 +51,19 @@ def is_possible_phone_number(query_str):
...
+        word = query_words[i]
+        if u'\u3400' <= word <= u'\u9fff':
```
gimite: Add a comment that this range is copied from the logic in text_query.py. Otherwise it is a bit confusing because this range actually doesn't represent CJK well :)

app/results.py
ezbr: Yes, I verified that behavior. Comment added.

app/results.py
ezbr: Done.

app/results.py
ezbr: Comment added.

gimite:
Somehow your pull request started to show so many diffs:
https://github.com/google/personfinder/pull/156/files
including recent change to add Babel. Can you try:

    $ git pull origin master
    $ git merge master

on your branch, maybe?

ezbr:
Yeah, I don't know what caused that. When I tried those two commands, it
just says "Already up-to-date." If you look at this page, you can see the
real changes:
https://github.com/google/personfinder/compare/ezbr-fix-issue_135. I think
all the other changes are the changes that have been merged to master since
I created my branch maybe, but I don't know.

gimite:
Can you try "git push origin ezbr-fix-issue_135" after running the commands?

ezbr:
I tried, it just says "Everything up-to-date". I think it is showing all
those files because I merged the changes I had on that branch into a git
directory on another computer that had all of these other changes made to
master that didn't exist in the ezbr-fix-issue_135. I'm pretty sure that
the only difference between master and ezbr-fix-issue_135 now are the
changes shown here:
https://github.com/google/personfinder/compare/ezbr-fix-issue_135.

gimite:
Hmm I believe it should automatically compare against the master branch you merged last time, but something may be broken.

The change LGTM. Can you pull and merge latest master again and try:

    tools/all_tests

to see if the tests are not broken?

You need to pull again because I have just fixed a few tests. You can ignore PhotoUploadTest failure because it depends on some image libraries.

ezbr:
Okay, I merged master and now it correctly shows only the file that I
changed. I also ran tools/all_tests and all unit and server tests passed.

gimite:
Merged. Thanks!

Issue #140, opened 2015-08-26
Automatically call setup_datastore() on the first HTTP access to the newly created application instance


gimite:
Currently you need to manually call setup_datastore() for a newly created application instance, as instructed in https://github.com/google/personfinder/wiki/GettingStarted . It would be nice to automate it.

To implement it, we need to be careful about mutual exclusion. It's quite likely that multiple HTTP requests are made in parallel. We need some kind of mutual exclusion, or maybe just show a [initialize] button when not initialized, and let user manually click the button to initialize.

This is subitem of #78 .

clobrano:
Hi, I would like to work on this issue

gimite:
https://github.com/google/personfinder/pull/190

PR #158, opened 2015-09-03
 Automatically call setup_datastore() on the first HTTP access to the newly created application instance #140 

clobrano:
The main issue here is that the `setup_datastore` function is outside PF visibility when it is run in its virtual environment by GAE. However, `setup_datastore` is called only in some server_tests and changing the location of the module that contains this function does not affect them, so I think it is safe to move the whole module under `app` folder and let PF calls setup_datastore at the first HTML request when the Repos are empty.

Let me know what do you think about this solution.
Best regards


gimite:
As I said in #158:

To implement it, we need to be careful about mutual exclusion. It's quite likely that multiple HTTP requests are made in parallel. We need some kind of mutual exclusion, or maybe just show a [initialize] button when not initialized, and let user manually click the button to initialize.

In your current code, it looks it runs setup_datastore() twice when multiple HTTP requests are made in parallel, which I believe is problematic.

clobrano:
Right @gimite, I'll fix that

clobrano:
Hi @gimite,

please review the last change. I added a critical area around the call to setup_datastore.

Thank you

gimite:
AppEngine app can run in multiple processes in multiple machines. So threading.RLock doesn't work. Maybe it's hard to perform a global lock in AppEngine. If you find a solution, it would be great, but you can instead show an [initialize] button, and require human to perform mutual exclusion, which is awkward but would be easy to implement.

clobrano:
Ok, I will try with the "button" solution then

clobrano:
Hi @gimite,

could you please review this change?

gimite:
Oops sorry somehow I didn't notice that you pushed a new change.

Somehow https://github.com/google/personfinder/pull/158/files contains 1623 files which are unrelated to your change. Can you look into it? How about pulling and merging master branch again? Or maybe recreate a pull request?

clobrano:
Sorry, I didn't check the diff, before pushing that. I'll check which
solution is faster.

clobrano:
Hi @gimite,

I opted to create another pull request

gimite:
Thanks! Closing this one.

Issue #128, opened 2015-08-14
"Mark records as duplicate" table has too much space


gimite:
![e50zdfufg](https://cloud.githubusercontent.com/assets/15363/9402234/b56d8f68-4815-11e5-9d9e-fcda7a73d247.png)

To see this page, create two person records with a common family name, search for the family name, click "duplicat marking mode", check both of them, and click "Mark records as duplicate".

kspviswa:
@gimite 

I couldn't view the screenshot. On clicking the link, I was asked enter my google credentials. On entering my gmail ID, it doesn't proceed further. Maybe the sreenshot reside in Google's private servers.

Let me know if I can work on this.

BR,
Viswa

gimite:
Yeah sorry it was our internal page. I replaced it with an externally visible one.

gimite:
It would be great if you work on it.

kspviswa:
Hi, 

I had created a pull request containing the fix for this issue. It is here : https://github.com/google/personfinder/pull/160 .

Please review & accept the merge, if deemed fit.

gimite:
As I mentioned in #160, it has actually been fixed by #146. Sorry for miscommunication.

PR #160, opened 2015-09-04
Pushing the changes containing bugfix for issue #128

kspviswa:
**Root Cause:**

The `add_note.html.template` is used for *2-in-1* purpose. It is used for showing the notes page in the create_person view & also for showing the form in the duplicate entry page.

By default, the form in  `add_note.html.template` does not have any div alignment or div class. In case of create_person results view, the notes page is tagged with `<div class="column end">`, where as the actual person information is tagged with `<div class="column start">`. However in case of duplicate records view, the  `add_note.html.template`  does not form under any *parent <div>*. Hence you are getting the *weird*  look and feel .

Since the  `add_note.html.template` has been used for both create_person and duplicate record, the fix has to be done carefully, so that it does not affect the create_person results view.

**Fix:**

The fix is to add a `<div class>` especially for duplicate records view, so that the view is rendered correctly

**Unit Testing:**

Duplicate record view, after fixing the bug.
image

Create person results view, after fixing the bug.
image

### DIFF
--- a/app/resources/add_note.html.template
+++ b/app/resources/add_note.html.template
@@ -17,6 +17,10 @@
   <input type="hidden" name="id" value="{{params.id}}">
 {% endif %}
 
+{% if markdup %}
+  <div class="column middle">
+{% endif %}
+
 <div class="fields-table note" role="complementary">
   <h2>
     {% if not markdup %}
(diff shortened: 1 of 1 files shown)


gimite:
Thanks for the contribution, but actually it has already been fixed by #146. Sorry I didn't notice that #128 wasn't auto-updated with reference to #146.

PR #163, opened 2015-09-08
Added rank order #67

prazuma:
- person name match higher than location match
  - e.g., [name=Rin Shibuya, location=Shinjuku], [name=Rin Tosaka, location=Shibuya]
-> if search by "Rin Shibuya", 1st: name = Rin Shibuya, 2nd: name=Rin Tosaka 
- demo: http://personfindersan.appspot.com/765production


app/full_text_search.py
```diff
@@ -87,6 +110,7 @@ def search(repo, query_txt, max_results):
     person_location_index_results = person_location_index.search(
         appengine_search.Query(
             query_string=and_query, options=options))
+    logging.info(person_location_index_results)
```
gimite: Remove this logging.

app/full_text_search.py
```diff
@@ -37,6 +37,22 @@ def make_or_regexp(query_txt):
     regexp = '|'.join([re.escape(word) for word in query_words if word])
     return re.compile(regexp, re.I)
 
+def create_sort_expression(**kwargs):
+    """
+    Creates sortExpressions for ranking.
```
gimite: s/sortExpressions/SortExpression's/
(If it's "SortExpressions", it looks like the class name is "SortExpressions". So I use ['s] instead. Not sure if it's a common way.)

gimite:
Can you set up a demo server for this change?

app/full_text_search.py

app/full_text_search.py

gimite:
Write more detailed description of what you have changed in the pull request description.

app/full_text_search.py
```diff
@@ -37,6 +37,22 @@ def make_or_regexp(query_txt):
...
+    """
+    Creates SortExpression's for ranking.
+    Args:
+        **kwargs: key=field_name, value=field_rank_value
+    Returns:
+        array of SortExpression
+    """
+    expressions = []
+    for field in kwargs:
+        expressions.append(appengine_search.SortExpression(
```
gimite: It looks you need to use "count(field_name)" instead:

https://cloud.google.com/appengine/docs/python/search/sortexpressionclass
> The expression can simply be a field name, or some compound expression such as _score + count(likes) * 0.1 which will add the score from a scorer to a count of the values of a likes field weighted by a factor of 0.1.

app/full_text_search.py
```diff
@@ -38,6 +38,22 @@ def make_or_regexp(query_txt):
...
 
+def create_sort_expression():
+    """
+    Creates SortExpression's for ranking.
+    Args:
+        **kwargs: key=field_name, value=field_rank_value
+    Returns:
+        array of SortExpression
+    """
+    expressions = []
```
gimite: You can simplify here by:

    return [appengine_search.SortExpression(...)]

app/full_text_search.py
```diff
@@ -130,6 +152,12 @@ def create_jp_name_fields(**kwargs):
                         name=field+'_romanized_by_jp_name_dict',
                         value=romanized_japanese_name)
                 )
+                connected_romanized_jp_name = ''
+                for x in range(5):
```
gimite: This can be simplified by:

    ''.join(romanized_japanese_name for _ in range(5))

_ is a commonly used name of a variable which is unused.

Also, wouldn't this results in a string like 'yamadayamadayamadayamada'? Don't you need a whitespace?

Also, please define a constant for 5 and use it here instead of directly using 5.

app/full_text_search.py
```diff
@@ -130,6 +152,12 @@ def create_jp_name_fields(**kwargs):
                         name=field+'_romanized_by_jp_name_dict',
                         value=romanized_japanese_name)
                 )
+                connected_romanized_jp_name = ''
```
gimite: "repeated" may be better than "connected".

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py
```diff
@@ -171,10 +199,18 @@ def create_document(record_id, repo, **kwargs):
...
 
     # Add name and location romanized by unidecode
+    name_fields = ['given_name', 'full_name', 'family_name', 'alternate_names']
     for field in kwargs:
         romanized_value = script_variant.romanize_word(kwargs[field])
         fields.append(
             appengine_search.TextField(name=field, value=romanized_value))
+        if romanized_value and field in name_fields:
+            connected_romanized_value = ''
+            for x in range(5):
```
gimite: This code is very similar to line 156-157. Can you define a function for this instead of writing the same code twice? Repeating the same code twice is usually a bad sign :)

app/full_text_search.py
```diff
@@ -171,10 +199,18 @@ def create_document(record_id, repo, **kwargs):
     fields.append(appengine_search.TextField(name='record_id', value=record_id))
 
     # Add name and location romanized by unidecode
+    name_fields = ['given_name', 'full_name', 'family_name', 'alternate_names']
     for field in kwargs:
         romanized_value = script_variant.romanize_word(kwargs[field])
         fields.append(
             appengine_search.TextField(name=field, value=romanized_value))
```
gimite: Is this field (without _for_rank) used for anything?

app/script_variant.py
```diff
@@ -94,4 +94,4 @@ def romanize_text(query_txt):
         script varianted query_txt (except kanji)
     """
     query_words = query_txt.split(' ')
-    return ' '.join([romanize_word(word) for word in query_words])
+    return ' '.join([romanize_word(word).lower() for word in query_words])
```
gimite: Why did you change this?

app/full_text_search.py
prazuma: - done
- 'yamadayamada...' works
- done

tests/test_full_text_search.py
```diff
@@ -182,6 +204,11 @@ def test_search_by_name_only(self):
         results = full_text_search.search('haiti', 'Makoto', 5)
         assert not results
 
+        # Check rank order
```
gimite: Please explain why haiti/0810 should be ranked higher.

gimite:
Can you update the pull request comment to explain how is the new ranking determined?

app/full_text_search.py

app/full_text_search.py
prazuma: location: used, name: not necessary.
- >changed: add only location(without_for_rank)

app/script_variant.py
prazuma: (´・∀・｀)?

tests/test_full_text_search.py
prazuma: done

app/full_text_search.py
```diff
@@ -25,6 +25,8 @@
 # This index contains person name and location.
 PERSON_LOCATION_FULL_TEXT_INDEX_NAME = 'person_location_information'
 
+REPEAT_COUNT_FOR_RANK = 5
```
gimite: Add a comment for this constant.

app/full_text_search.py
```diff
@@ -38,6 +40,20 @@ def make_or_regexp(query_txt):
     regexp = '|'.join([re.escape(word) for word in query_words if word])
     return re.compile(regexp, re.I)
 
+def create_sort_expression():
+    """
+    Creates SortExpression's for ranking.
+    Args:
```
gimite: Remove this Args section because it no longer takes arguments.

app/full_text_search.py
```diff
@@ -38,6 +40,20 @@ def make_or_regexp(query_txt):
     regexp = '|'.join([re.escape(word) for word in query_words if word])
     return re.compile(regexp, re.I)
 
+def create_sort_expression():
```
gimite: Should be named create_sort_expressions.

app/full_text_search.py
```diff
@@ -130,6 +172,8 @@ def create_jp_name_fields(**kwargs):
                         name=field+'_romanized_by_jp_name_dict',
                         value=romanized_japanese_name)
                 )
+                fields.extend(create_fields_for_rank(
+                    field+'romanized_by_jp__name_dict', romanized_jp_name))
```
gimite: s/jp__name_dict/jp_name_dict/ (double underscores)

app/full_text_search.py

app/full_text_search.py
```diff
@@ -113,6 +135,26 @@ def search(repo, query_txt, max_results):
...
+        value: field value
+    Returns:
+        array of appengine_search.TextField(name=field_name, value=value)
+           (length: REPEAT_COUNT_FOR_RANK)
+    """
+    if not value:
+        return []
+
+    fields = []
+    for x in range(REPEAT_COUNT_FOR_RANK):
```
gimite: Use xrange instead of range. They have the same effect in this context, but range(5) actually creates an array [0, 1, 2, 3, 4, 5] while xrange doesn't.

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py
```diff
@@ -113,6 +135,26 @@ def search(repo, query_txt, max_results):
...
+        fields.append(
```
gimite: It's up to you, but how about putting it into a single field (instead of 5 fields) but joining with a whitespace (e.g., "yamada yamada yamada yamada yamada")? It looks a bit simpler and can avoid adding too many fields.

app/full_text_search.py

app/full_text_search.py
```diff
@@ -113,6 +135,26 @@ def search(repo, query_txt, max_results):
     return results
 
 
+def create_fields_for_rank(field_name, value):
+    """
+    Creates fields for ranking.
```
gimite: Document why you repeat the value 5 times here.

app/full_text_search.py

app/full_text_search.py
```diff
@@ -171,10 +214,15 @@ def create_document(record_id, repo, **kwargs):
     fields.append(appengine_search.TextField(name='record_id', value=record_id))
 
     # Add name and location romanized by unidecode
+    name_fields = ['given_name', 'full_name', 'family_name', 'alternate_names']
     for field in kwargs:
         romanized_value = script_variant.romanize_word(kwargs[field])
-        fields.append(
-            appengine_search.TextField(name=field, value=romanized_value))
+
+        if field in name_fields:
```
gimite: Add a comment why you call create_fields_for_rank for name_fields, but not for other fields.

app/full_text_search.py
prazuma: i tried, but it doesn't work...(´・ω･`)

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py
```diff
@@ -107,15 +128,41 @@ def search(repo, query_txt, max_results):
...
+        array of appengine_search.TextField(name=field_name, value=value)
+           (length: REPEAT_COUNT_FOR_RANK)
+    """
+    if not value:
+        return []
+
+    fields = []
+    for x in xrange(REPEAT_COUNT_FOR_RANK):
+        fields.append(
+            appengine_search.TextField(name=field_name+'_for_rank_'+str(x),
```
gimite: Would be cleaner to write '%s_for_rank_%d' % (field_name, x) instead.

app/full_text_search.py
```diff
@@ -107,15 +128,41 @@ def search(repo, query_txt, max_results):
 
         if regexp.search(names) or regexp.search(romanized_jp_names):
             index_results.append(id)
-
+    logging.info(index_results)
```
gimite: Remove this line.

app/full_text_search.py
```diff
@@ -107,15 +128,41 @@ def search(repo, query_txt, max_results):
...
     results = []
     for id in index_results:
         result = model.Person.get(repo, id, filter_expired=True)
         if result:
             results.append(result)
+            logging.info(id)
```
gimite: Remove these loggings.

app/full_text_search.py
```diff
@@ -171,6 +218,8 @@ def create_romanized_name_fields(romanize_method, **kwargs):
                 appengine_search.TextField(
```
gimite: Do you need to keep this field in addition to "for_rank" ones? Didn't you remove it in the code before merging?

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

gimite:
LGTM. Merged.

PR #172, opened 2015-09-14
Fix error message on connecting to the appserver

prazuma:


tools/remote_api.py
```diff
@@ -98,7 +98,7 @@ def get_login():
                 raise
     except Exception, e:
         if isinstance(e, urllib2.HTTPError):
-            print >>sys.stderr, 'HTTP error %d from URL: %s' % (e.code, e.url)
+            print >>sys.stderr, e.read()
```
gimite: I found that "elif isinstance(e, urllib2.URLError):" block below does something similar, but it also outputs hostport. Can you try simply removing this if-block? HTTPError is subclass of URLError, so it should be automatically handled by the block below.

prazuma: done

gimite:
LGTM. Merged. Thanks for the fix!

PR #175, opened 2015-09-20
Update Readme.md with Deploy instruction

groschovskiy:


gimite:
Thanks for the request, and sorry for late.

There is actually a detailed doc which explains how to set up the environment:
https://github.com/google/personfinder/wiki/GettingStarted

So I made another pull request #188 to point to the doc from README.

PR #176, opened 2015-09-24
Rank same kanji higher than different kanji with the same reading#67

prazuma:
- if there are records record1:[name=菊地真], record2:[name=菊地眞], get results(1st: 菊地真、2nd: 菊地眞) when search by "菊地 真"
- demo: http://pfrank-1078.appspot.com/global/home.html


app/full_text_search.py
```diff
@@ -213,6 +242,12 @@ def create_romanized_name_fields(romanize_method, **kwargs):
         if romanized_name:
             fields.extend(create_fields_for_rank(field, romanized_name))
             romanized_names_list.append(romanized_name)
+        if script_variant.has_kanji(kwargs[field]):
```
gimite: You may not limit to kanji. You may always add this _non_romanized field and rank exact matches higher than others. What do you think?

app/full_text_search.py
```diff
@@ -109,12 +113,35 @@ def search(repo, query_txt, max_results):
...
+
+    person_location_index_results_with_kanji = person_location_index.search(
+        appengine_search.Query(
+            query_string=and_query_with_kanji, options=options)
+    )
+
     index_results = []
-    regexp = make_or_regexp(query_txt)
+    regexp = make_or_regexp(romanized_query)
+    for document in person_location_index_results_with_kanji:
```
gimite: The code here is very similar to the code at line 145. Can you define a function instead?

In general, repeating the same code multiple times is considered bad. It makes harder to read, and you need to update both codes when you need to modify the code. Please consider defining a function instead before you copy&paste multiple lines of code.

app/script_variant.py
```diff
@@ -35,6 +35,8 @@ def has_kanji(word):
     """
     # [\u3400-\u9fff]: kanji
     # TODO: update this range (some of kanji characters out of this range.)
+    if not word:
```
gimite: Do you need this code? I believe it anyway just returns False if word is empty.

tests/test_full_text_search.py
```diff
@@ -254,6 +269,12 @@ def test_search_by_name_only(self):
         assert set([r.record_id for r in results]) == \
             set(['haiti/0225'])
 
+        # Check rank order
+        # (same kanji higher than different kanji with the same reading)
+        results = full_text_search.search('haiti', u'菊地 真', 5)
```
gimite: Add test with [菊池 眞] too. If we test with just one of them, there is possibility that ['haiti/0829', 'haiti/1829'] are always ranked in this order regardless of the query.

app/full_text_search.py

app/full_text_search.py
prazuma: fixed

app/script_variant.py

tests/test_full_text_search.py
prazuma: done

app/full_text_search.py
```diff
@@ -54,7 +54,7 @@ def create_sort_expressions():
     return [appengine_search.SortExpression(
         expression='_score',
         direction=appengine_search.SortExpression.DESCENDING,
-        default_value=0.0
+        default_value=1.0
```
gimite: Why did you change this to 1.0?

app/full_text_search.py
```diff
@@ -244,9 +267,10 @@ def create_romanized_name_fields(romanize_method, **kwargs):
     for field_name in kwargs:
         romanized_names = romanize_method(kwargs[field_name])
         for index, romanized_name in enumerate(romanized_names):
-            fields.extend(create_fields_for_rank('%s_%d' %
-                                                 (field_name, index),
+            fields.extend(create_fields_for_rank('%s_%s_%d' %
+                                                 (field_name, romanize_method_name, index),
```
gimite: Over 80 chars.

app/full_text_search.py
```diff
@@ -244,9 +267,10 @@ def create_romanized_name_fields(romanize_method, **kwargs):
     for field_name in kwargs:
         romanized_names = romanize_method(kwargs[field_name])
         for index, romanized_name in enumerate(romanized_names):
-            fields.extend(create_fields_for_rank('%s_%d' %
-                                                 (field_name, index),
+            fields.extend(create_fields_for_rank('%s_%s_%d' %
```
gimite: Add a comment why you add both of romanized and non-romanized fields.

app/full_text_search.py
```diff
@@ -244,9 +267,10 @@ def create_romanized_name_fields(romanize_method, **kwargs):
...
+                                                 (field_name, romanize_method_name, index),
                                                  romanized_name))
+            fields.extend(create_fields_for_rank('%s_%d' % (field_name, index), kwargs[field_name]))
```
gimite: Do you need to add this field (with the same value) len(romanized_names) times?

app/full_text_search.py
```diff
@@ -244,9 +267,10 @@ def create_romanized_name_fields(romanize_method, **kwargs):
     for field_name in kwargs:
```
gimite: You can write

    for field_name, field_value in kwargs.iteritems():

to get the value instead.

app/full_text_search.py
```diff
@@ -125,6 +150,11 @@ def search(repo, query_txt, max_results):
     # Remove double quotes so that we can safely apply enclose_in_double_quotes().
     query_txt = re.sub('"', '', query_txt)
     romanized_query = create_query_txt(query_txt)
+    kanji_words_in_query_txt = script_variant.find_kanji_word(query_txt)
```
gimite: If you want to apply the logic for non-kanji words as well, you should extract all words here instead of kanji words.

app/full_text_search.py
```diff
@@ -101,6 +101,31 @@ def create_query_txt(query_txt):
     return enclose_in_parenthesis(romanized_query)
 
 
+def get_person_ids_from_results(romanized_query, results_list):
+    """
+    Returns person record_id (a part of romanized_query matches person name) list.
```
gimite: "Returns record_id of persons whose name contains at least one word in romanized_query."

Also, please mention that:

- Why you need to apply this regexp
- It also removes dups (i.e., If results_list contains multiple results with the same index_results, it returns just one of them.)

app/full_text_search.py
```diff
@@ -125,6 +150,11 @@ def search(repo, query_txt, max_results):
     # Remove double quotes so that we can safely apply enclose_in_double_quotes().
     query_txt = re.sub('"', '', query_txt)
     romanized_query = create_query_txt(query_txt)
+    kanji_words_in_query_txt = script_variant.find_kanji_word(query_txt)
+    double_quote_kanji_words_in_query_txt = ' '.join(
+        enclose_in_double_quotes(word) for word in kanji_words_in_query_txt)
+    romanized_query_with_kanji = romanized_query + ' ' +\
```
gimite: Do you need to add romanized_query? Would it work if you search just with double_quote_kanji_words_in_query_txt?

app/full_text_search.py
```diff
@@ -144,27 +174,20 @@ def search(repo, query_txt, max_results):
...
-            if field.name == 'names_romanized_by_romanize_word_by_unidecode':
-                names = field.value
-            if field.name == 'record_id':
-                id = field.value
-            if field.name == 'names_romanized_by_romanize_japanese_name_by_name_dict':
-                romanized_jp_names = field.value
-
-        if regexp.search(names) or regexp.search(romanized_jp_names):
-            index_results.append(id)
+    person_location_index_results_with_kanji = person_location_index.search(
```
gimite: Add a comment why you need to perform this search in addition to person_location_index_results above.

app/full_text_search.py
prazuma: fixed

app/full_text_search.py

app/full_text_search.py
prazuma: removed romanized_query

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py
prazuma: fixed

app/full_text_search.py
```diff
@@ -281,6 +317,17 @@ def create_romanized_location_fields(romanize_method, **kwargs):
             )
     return fields
 
+def create_non_romanized_fields(**kwargs):
+    """
+    Creates non romanized fields for ranking same kanji higher than
```
gimite: Now this is not limited to kanji, so you should instead say e.g., "to rank exact matches higher than non-exact matches with the same romanization". Also it would be nice to add example here.

app/full_text_search.py
```diff
@@ -101,6 +101,37 @@ def create_query_txt(query_txt):
     return enclose_in_parenthesis(romanized_query)
 
 
+def get_person_ids_from_results(romanized_query, results_list):
+    """
+    Returns person record_id of persons
+    whose name contains at least one word in romanized_query.
+    We need regexp because check romanized_query matches
```
gimite: Please fix this comment as discussed offline.

app/full_text_search.py
gimite: It is not removed yet.

app/full_text_search.py
```diff
@@ -149,22 +186,19 @@ def search(repo, query_txt, max_results):
...
-            if field.name == 'names_romanized_by_romanize_word_by_unidecode':
-                names = field.value
-            if field.name == 'record_id':
-                id = field.value
-            if field.name == 'names_romanized_by_romanize_japanese_name_by_name_dict':
-                romanized_jp_names = field.value
-
-        if regexp.search(names) or regexp.search(romanized_jp_names):
-            index_results.append(id)
+    # For ranking same kanji higher than different kanji with the same reading.
```
gimite: Now this is not limited to kanji, so you should instead say e.g., "To rank exact matches higher than non-exact matches with the same romanization".

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py
```diff
@@ -101,6 +114,38 @@ def create_query_txt(query_txt):
     return enclose_in_parenthesis(romanized_query)
 
 
+def get_person_ids_from_results(romanized_query, results_list):
+    """
+    Returns person record_id of persons
+    whose name contains at least one word in romanized_query.
+    We need regexp to check romanized_query matches
+    at least a part of person name.
+    To protect users' privacy, we should not return records matches location only.
```
gimite: Please fix English grammar here as chatted offline.

app/full_text_search.py
```diff
@@ -144,27 +190,21 @@ def search(repo, query_txt, max_results):
...
-            if field.name == 'record_id':
-                id = field.value
-            if field.name == 'names_romanized_by_romanize_japanese_name_by_name_dict':
-                romanized_jp_names = field.value
-
-        if regexp.search(names) or regexp.search(romanized_jp_names):
-            index_results.append(id)
+    # To rank exact matches higher than non-exact matches with the same romanization.
+    non_romanized_and_query = non_romanized_query + ' AND (repo: ' + repo + ')'
+    person_location_index_results_with_kanji = person_location_index.search(
```
gimite: Since it's no longer limited to kanji, please rename this to e.g., "non_romanized_person_location_index_results".

app/full_text_search.py
```diff
@@ -241,15 +281,17 @@ def create_romanized_name_fields(romanize_method, **kwargs):
...
-            fields.extend(create_fields_for_rank('%s_%d' %
-                                                 (field_name, index),
+            fields.extend(create_fields_for_rank('%s_%s_%d' %
+                                                 (field_name,
+                                                  romanize_method_name, index),
                                                  romanized_name))
         romanized_names_list.extend(romanized_names)
 
-    full_name_fields, romanized_full_names = create_full_name_without_space_fields(
+    full_name_fields, romanized_full_names =\
```
gimite: Insert white space between "=" and "\". Or put "(" instead here and close it at the end of the statement.

app/full_text_search.py

app/full_text_search.py

app/full_text_search.py

gimite:
LGTM. Please merge it after running tools/all_tests.

Issue #179, opened 2015-09-28
Support romanizing "山田太郎" without space just by "山田" and "太郎" in the dictionary


gimite:
For full-text search with Japanese dictonary.

PR #186, opened 2015-09-29
Support romanizing "山田太郎" without space just by "山田" and "太郎" in the dictionary #179

prazuma:


app/full_text_search.py
```diff
@@ -214,10 +230,34 @@ def create_romanized_name_fields(romanize_method, **kwargs):
             fields.extend(create_fields_for_rank(field, romanized_name))
             romanized_names_list.append(romanized_name)
 
-    full_name_fields, romanized_full_names = create_full_name_without_space_fields(
-        romanize_method, kwargs['given_name'], kwargs['family_name'])
-    fields.extend(full_name_fields)
-    romanized_names_list.extend(romanized_full_names)
+    splited_full_names = []
```
gimite: Past participle (過去分詞) of "split" is "split", not "splited".

app/full_text_search.py
```diff
@@ -185,21 +186,36 @@ def create_full_name_without_space_fields(romanize_method, given_name, family_na
...
         romanized_name_list.append(full_name_family_given)
     return fields, romanized_name_list
 
 
+def split_full_name_without_space(romanize_method, full_name):
+    names = []
+    for index in xrange(len(full_name)):
+        family_name = full_name[:index]
+        given_name = full_name[index:]
+        romanized_family_name = romanize_method(family_name)
```
gimite: Please merge with the master branch first. Then this romanize_method now returns an array instead, so please support it.

app/full_text_search.py
```diff
@@ -214,10 +230,34 @@ def create_romanized_name_fields(romanize_method, **kwargs):
...
             romanized_names_list.append(romanized_name)
 
-    full_name_fields, romanized_full_names = create_full_name_without_space_fields(
-        romanize_method, kwargs['given_name'], kwargs['family_name'])
-    fields.extend(full_name_fields)
-    romanized_names_list.extend(romanized_full_names)
+    splited_full_names = []
+    full_name = kwargs['full_name']
+    if full_name and not (' ' in full_name):
+        splited_full_names = split_full_name_without_space(
```
gimite: We want to apply this for search queries too. So how about implementing this in script_variant.romanize_japanese_name_by_name_dict instead?

app/full_text_search.py
```diff
@@ -185,21 +186,36 @@ def create_full_name_without_space_fields(romanize_method, given_name, family_na
...
+            name='no_space_full_name_%d_romanized_by_%s_2' % (
+                index, romanize_method_name),
             value=full_name_family_given))
         romanized_name_list.append(full_name_family_given)
     return fields, romanized_name_list
 
 
+def split_full_name_without_space(romanize_method, full_name):
+    names = []
+    for index in xrange(len(full_name)):
```
gimite: We don't have reason to split at position 0, so you can start from index=1.

app/full_text_search.py

app/full_text_search.py
prazuma: merged

app/full_text_search.py

app/script_variant.py
```diff
@@ -44,6 +44,18 @@ def has_kanji(word):
     return re.match(ur'([\u3400-\u9fff])', word)
 
 
+def find_word_in_dict(word):
```
gimite: This function also romanizes, not just "find". Also the function name should contain "japanese" because it uses Japanese dictionary. How about e.g., "romanize_single_japanese_word_by_name_dict"? Please also add a comment for the function.

tests/test_script_variant.py
```diff
@@ -8,13 +8,11 @@
 
 class ScriptVariantTests(unittest.TestCase):
     def test_romanize_japanese_name_by_name_dict(self):
-        assert script_variant.romanize_japanese_name_by_name_dict(
-            u'雪歩') == [u'YUKIHO']
+        results = script_variant.romanize_japanese_name_by_name_dict(u'天海')
```
gimite: Can you also add a test with a single letter name?

tests/test_script_variant.py
```diff
@@ -8,13 +8,11 @@
...
+        assert len(results) > 1
```
gimite: Can you also test that results contain any of the expected romanizations e.g., "TENKAI"?

app/script_variant.py

app/script_variant.py
```diff
@@ -57,12 +69,21 @@ def romanize_japanese_name_by_name_dict(word):
...
 
-    if word in JAPANESE_NAME_DICTIONARY:
-        yomigana_list = JAPANESE_NAME_DICTIONARY[word]
-        return [jautils.hiragana_to_romaji(yomigana)
-                for yomigana in yomigana_list]
-
-    return [word]
+    words = []
+    for index in xrange(1, len(word)):
+        first_part = word[:index]
```
gimite: Please add a comment why you split words.

app/script_variant.py
```diff
@@ -57,12 +69,21 @@ def romanize_japanese_name_by_name_dict(word):
...
+        last_part = word[index:]
+        romanized_first_parts = find_word_in_dict(first_part)
+        romanized_last_parts = find_word_in_dict(last_part)
+        for romanized_first_part in romanized_first_parts:
+            for romanized_last_part in romanized_last_parts:
+                if (romanized_first_part != first_part) and (
```
gimite: Nit: It would be better to put a single bracket for the whole expression in these two lines. Two brackets are redundant.

tests/test_script_variant.py
prazuma: done

app/script_variant.py
```diff
@@ -57,12 +69,21 @@ def romanize_japanese_name_by_name_dict(word):
...
+                        romanized_last_part != last_part):
+                    words.append(romanized_first_part+romanized_last_part)
+                    words.append(romanized_last_part)
```
gimite: I believe you don't have to (and shouldn't) add these "only last name" and "only first name". If we do this, searching for "山田太郎" also returns search result for just "yamada", right?

tests/test_script_variant.py
prazuma: done

app/script_variant.py

app/script_variant.py

app/script_variant.py

app/script_variant.py
```diff
@@ -44,9 +44,9 @@ def has_kanji(word):
     return re.match(ur'([\u3400-\u9fff])', word)
 
 
-def romanize_japanese_name_by_name_dict(word):
+def romanize_single_japanese_word_by_name_dict(word):
     """
-    This method romanizes japanese name by using name dictionary.
+    This method romanzies japanese name by using name dictionary.
```
gimite: s/romanizies/romanizes/

Also you should mention that this method doesn't support romanizing full names using first/last names in the dictionary. Otherwise it's not clear how this method is different from romanize_japanese_name_by_name_dict().

app/script_variant.py
```diff
@@ -65,6 +65,40 @@ def romanize_japanese_name_by_name_dict(word):
...
+    (because there are multiple ways to read the same kanji name in japanese)
+    Returns:
+        [romanized_jp_name, ...]
+    """
+    if not word:
+        return ['']
+
+    words = []
+    for index in xrange(1, len(word)):
+        # split word because query word may not contains white space
```
gimite: This is still not clear. Even if the query doesn't contain white space, it would work fine if the query is "山田". Please explain what exactly is the case you want to handle, maybe with an example.

app/script_variant.py
```diff
@@ -65,6 +65,40 @@ def romanize_japanese_name_by_name_dict(word):
     return [word]
 
 
+def romanize_japanese_name_by_name_dict(word, is_create=True):
```
gimite: Add document of is_create argument. Also, it may be better to name it "for_index".

app/script_variant.py
```diff
@@ -65,6 +65,40 @@ def romanize_japanese_name_by_name_dict(word):
...
+        romanized_first_parts = romanize_single_japanese_word_by_name_dict(
+            first_part)
+        romanized_last_parts = romanize_single_japanese_word_by_name_dict(
+            last_part)
+        for romanized_first_part in romanized_first_parts:
+            for romanized_last_part in romanized_last_parts:
+                if romanized_first_part != first_part and\
+                   romanized_last_part != last_part:
+                    words.append(romanized_first_part+romanized_last_part)
+                    if is_create:
```
gimite: Please add a comment why you add these fields only if is_create is true.

tests/test_script_variant.py
gimite: If the test below passes, len(results) is obviously more than one. So you can remove this test.

tests/test_script_variant.py
gimite: Can you also add a test with full name without white spaces? "天海" is also interpreted as full name in a sense, but you should add a case which is more obviously full name.

app/script_variant.py
```diff
@@ -65,6 +65,40 @@ def romanize_japanese_name_by_name_dict(word):
...
+        # split word because query word may not contains white space
+        first_part = word[:index]
+        last_part = word[index:]
+        romanized_first_parts = romanize_single_japanese_word_by_name_dict(
+            first_part)
+        romanized_last_parts = romanize_single_japanese_word_by_name_dict(
+            last_part)
+        for romanized_first_part in romanized_first_parts:
+            for romanized_last_part in romanized_last_parts:
+                if romanized_first_part != first_part and\
```
gimite: Add a white space before "\".

app/script_variant.py

app/script_variant.py
```diff
@@ -65,6 +65,40 @@ def romanize_japanese_name_by_name_dict(word):
...
+                   romanized_last_part != last_part:
```
gimite: Indent by 8 spaces from "if" at the line 92.

app/script_variant.py

app/script_variant.py
```diff
@@ -122,7 +156,7 @@ def romanize_word(word):
 
     romanized_words = []
     if has_kanji(word):
-        romanized_words = romanize_japanese_name_by_name_dict(word)
+        romanized_words = romanize_japanese_name_by_name_dict(word, is_create=False)
```
gimite: You set is_create=False here, which means that you assume this method is used for search queries. But it is not obvious from the method document. Can you do either:

- Add is_create parameter to normalize_word
- Rename the method to romanize_query or something, and document that this method is for search queries.

Note that you should do one of above, not both :)

app/script_variant.py

app/script_variant.py

app/script_variant.py

app/script_variant.py

app/script_variant.py

tests/test_script_variant.py
prazuma: done

tests/test_script_variant.py
prazuma: done

app/script_variant.py
```diff
@@ -65,6 +69,46 @@ def romanize_japanese_name_by_name_dict(word):
...
+    Returns:
+        [romanized_jp_name, and romanized_jp_name(split_word), ...]
+    """
+    if not word:
+        return ['']
+
+    words = []
+    for index in xrange(1, len(word)):
+        # split word because query word may not contains white space
+        # e.g., if the query (e.g., "山田") doesn't contain white space,
```
gimite: We may return "yamata" for "山田" with this change, but it's more like a side effect. What is your main motivation to implement this split? Please document it here.

app/script_variant.py
```diff
@@ -65,6 +69,46 @@ def romanize_japanese_name_by_name_dict(word):
...
 
+def romanize_japanese_name_by_name_dict(word, for_index=True):
+    """
+    This method romanizes japanese name by using name dictionary.
+    If word isn't found in dictionary, this method doesn't
+    apply romanize.
+    This method can return multiple romanizations.
+    (because there are multiple ways to read the same kanji name in japanese)
+    Args:
+        for_index: this method is called for indexing or not
```
gimite: Can you also mention when this method is called with for_index=False?

app/script_variant.py
```diff
@@ -65,6 +69,46 @@ def romanize_japanese_name_by_name_dict(word):
...
+        romanized_first_parts = romanize_single_japanese_word_by_name_dict(
+            first_part)
+        romanized_last_parts = romanize_single_japanese_word_by_name_dict(
+            last_part)
+        for romanized_first_part in romanized_first_parts:
+            for romanized_last_part in romanized_last_parts:
+                if romanized_first_part != first_part and \
+                        romanized_last_part != last_part:
+                    words.append(romanized_first_part+romanized_last_part)
+                    # To search by the query which doesn't contains white sapce,
```
gimite: This comment looks wrong. If the query is a full name without white space e.g., "やまだたろう", we don't need these fields, right? You need to explain:

- Why you need to add these fields in index
- Why you must not add these fields for search queries

It may be easier to explain using an example.

app/script_variant.py
```diff
@@ -122,7 +166,9 @@ def romanize_word(word):
 
     romanized_words = []
     if has_kanji(word):
-        romanized_words = romanize_japanese_name_by_name_dict(word)
+        # Add for_index parameter to normalize_word
```
gimite: I mean, you should add for_index parameter to normalize_word (not romanize_japanese_name_by_name_dict).

app/script_variant.py

app/script_variant.py

app/script_variant.py

app/script_variant.py

app/script_variant.py
```diff
@@ -65,6 +69,51 @@ def romanize_japanese_name_by_name_dict(word):
...
+        [romanized_jp_name, and romanized_jp_name(split_word), ...]
+    """
+    if not word:
+        return ['']
+
+    words = []
+    for index in xrange(1, len(word)):
+        # split word because query word may not contains white space
+        # to support romanizing fullname without space.
+        # e.g., if the query (e.g., "山田太郎") doesn't contain white space,
```
gimite: It should be something like:

> If the query is a full name without white space (e.g., "山田太郎"), and the first/last name is in the dictionary, but the full name is not in the dictionary, it can still return romanization "yamadataro".

app/script_variant.py
gimite: Line 90 and line 91 says similar thing. You can merge the two lines by e.g.,:

> Splits the word to support romanizing fullname without space.

app/script_variant.py
gimite: Combining these two lines, how about:

> Set this to True for indexing purpose.
> Set this to False when you want to romanize query text.

app/script_variant.py
```diff
@@ -65,6 +69,51 @@ def romanize_japanese_name_by_name_dict(word):
...
+        romanized_first_parts = romanize_single_japanese_word_by_name_dict(
+            first_part)
+        romanized_last_parts = romanize_single_japanese_word_by_name_dict(
+            last_part)
+        for romanized_first_part in romanized_first_parts:
+            for romanized_last_part in romanized_last_parts:
+                if romanized_first_part != first_part and \
+                        romanized_last_part != last_part:
+                    words.append(romanized_first_part+romanized_last_part)
+                    # if record contains a full name with kanji (e.g., "山田太郎"),
```
gimite: Can you add an example of a set of the query and the record where you need these fields for indexing?

Also, can you add an example of a set of the query and the record where it would be problematic if you add these fields for query?

app/script_variant.py

app/script_variant.py

app/script_variant.py
prazuma: doen

app/script_variant.py

gimite:
LGTM after making some modification. Merged.

PR #187, opened 2015-10-01
Combine Japanese name dictionary and location dictionary for romanization #183

prazuma:


app/script_variant.py
```diff
@@ -31,9 +31,20 @@ def read_dictionary(file_name):
         return None
     return dictionary
 
+def combine_name_location_dictionary(name_dict, location_dict):
```
gimite: How about putting a single file which both contains names and locations? It seems simpler.

tests/test_script_variant.py
```diff
@@ -7,21 +7,16 @@
 
 
 class ScriptVariantTests(unittest.TestCase):
-    def test_romanize_japanese_name_by_name_dict(self):
-        assert script_variant.romanize_japanese_name_by_name_dict(
+    def test_romanize_japanese_word(self):
+        assert script_variant.romanize_japanese_word(
```
gimite: Add a test for a word which is in location dictionary, but not in name dictionary, to make sure we don't miss location dictionary. '中野' is not good because it's probably also in name dictionary.

app/script_variant.py

tests/test_script_variant.py
prazuma: done

gimite: It looks this is not done yet.

app/script_variant.py
```diff
@@ -31,9 +31,7 @@ def read_dictionary(file_name):
         return None
     return dictionary
 
-JAPANESE_NAME_DICTIONARY = read_dictionary('japanese_name_dict.txt')
```
gimite: Can you delete japanese_name_dict.txt and jp_location_dict.txt which are no longer used?

tests/test_script_variant.py
prazuma: done

app/script_variant.py

gimite:
LGTM. Merged.

PR #190, opened 2015-10-14
Proposal fix for Issue #140 Automatically call setup_datastore()

clobrano:
Add button to call setup datastore in the admin page


app/resources/admin.html.template
```diff
@@ -36,6 +36,12 @@
   <input type=submit value="Delete (with confirmation)">
 </form>
 
+<h2>Setup datastore</h2>
+<form method="post" class="admin">
```
gimite: How about showing this button in all pages including the start page:
https://google.org/personfinder/global/home.html
instead of the real content, if the database is not initialized yet?

If we show this button only in the admin page, it would be hard to notice. All requests go through Main handler in main.py, so I believe you can hook all requests there.

clobrano: I did this way because I thought only admins were allowed to perform this step.
However, to get things clearer, what do you mean by "all pages"? When database is not initialized yet, I can only see this:

selection_010

If this is the normal view, I could add the button/link in the sidebar near home. Would be that ok?

clobrano:
@gimite could you please review this?

Thank you

gimite:
Sorry for late. I was a bit busy recently.

I don't think of a case that uninitialized Person Finder instance gets end users' traffic. So I believe you don't need to restrict the operation to admins.

There is almost nothing you can do with uninitialized instance except for initializing it. So how about showing only the "Set up datastore" button instead of the real pages, when the user accesses any pages including these pages?:
https://google.org/personfinder/global/home.html
https://google.org/personfinder/japan/

clobrano:
> Sorry for late. I was a bit busy recently.

No worries, I just thought you could have miss the notification.

> There is almost nothing you can do with uninitialized instance except for initializing it. So how about showing only the "Set up datastore" button instead of the real pages, when the user accesses any pages including these pages?:
> https://google.org/personfinder/global/home.html
> https://google.org/personfinder/japan/

With uninitialized instance, the link https://google.org/personfinder/japan/ goes to a simple 404 "not found page", do you want a new 404 page with a link to setup datastore?

gimite:
It may be rare to directly try to access https://google.org/personfinder/japan/ for a new instance. So showing 404 would be fine. Or, you can hook all requests (well, except for the request to initialize the database) in main.py and show the button regardless of the URL.

clobrano:
Hi @gimite 

I added a HTML fragment at the top of each served page with a warning and a link to inizialize the datastore. Please let me know if that is what you expected.

Best regards

app/main.py
```diff
@@ -494,9 +494,22 @@ def initialize(self, request, response):
...
+        if 0 == len(env.repo_options) and env.action != 'setup_datastore':
+            get_vars = lambda: {'env': env}
+            content = resources.get_rendered('setup_datastore.html', env.lang,
+                    (env.repo, env.charset), get_vars)
+            response.out.write(content)
+
         if not env.action and not env.repo:
             # Redirect to the default home page.
             self.redirect(env.global_url + '/' + HOME_ACTION)
+        elif env.action == "setup_datastore":
```
gimite: Having two end points to set up datastore looks weird. (One defined here, another defined in admin.py.) How about linking to the end point defined in admin.py from setup_datastore.html? Anyway it should be POST, not GET.

app/main.py
```diff
@@ -494,9 +494,22 @@ def initialize(self, request, response):
...
         request, response, env = self.request, self.response, self.env
+
+        # If the Person Finder instance has not been initialized yet,
+        # prepend to any served page a warning and a link to initialize
+        # the datastore.
+        if 0 == len(env.repo_options) and env.action != 'setup_datastore':
+            get_vars = lambda: {'env': env}
+            content = resources.get_rendered('setup_datastore.html', env.lang,
+                    (env.repo, env.charset), get_vars)
+            response.out.write(content)
```
gimite: In this case, I believe we can just show setup_datastore.html without the actual content i.e., "return" just after this line. Because there's nothing to do for uninitialized repositories other than initializing it. But you can leave it for now. I can change it later.

app/resources/repo-menu.html.template
```diff
@@ -19,4 +19,6 @@
       {% if option.test_mode %}({% trans "in test mode" %}){% endif %}</a><br>
   {% endfor %}
   </ul>
+{% else %}
```
gimite: Can you revert this part? I don't think we need this because we also have a header.

app/resources/setup_datastore.html.template
```diff
@@ -0,0 +1,17 @@
+{# Copyright 2010 Google Inc.  Licensed under the Apache License, Version   #}
```
gimite: Update the year to 2015.

app/main.py
```diff
@@ -515,7 +528,8 @@ def serve(self):
                 env.action, env.lang, (env.repo, env.charset), get_vars)
             if content is None:
                 response.set_status(404)
-                response.out.write('Not found')
+                page_not_found = 'Page "%s" could not be found' % env.action
```
gimite: env.action is a kind of technical detail, which doesn't always make sense to end users. Just saying "Not found" sounds enough for me.

gimite:
Thanks! I tested locally and it worked nicely. I added some line comments about its implementation.

app/main.py
```diff
@@ -494,9 +494,22 @@ def initialize(self, request, response):
 
     def serve(self):
         request, response, env = self.request, self.response, self.env
+
+        # If the Person Finder instance has not been initialized yet,
+        # prepend to any served page a warning and a link to initialize
+        # the datastore.
+        if 0 == len(env.repo_options) and env.action != 'setup_datastore':
```
gimite: len(env.repo_options) is not a good way to check if the instance is initialized. env.repo_options only returns active repositories, so it can often be empty even in production.

Can you define a new config, named "initialized" or something, which indicates that the instance is initialized? To set the config, you can add it to here:
https://github.com/google/personfinder/blob/master/tools/setup_pf.py#L62
To read the config, you can write config.get("initialized").

clobrano:
@gimite sorry for having misspelled your nickname in commit title ;)

app/resources/setup_datastore.html.template
```diff
@@ -0,0 +1,17 @@
...
+{# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. #}
+{# See the License for the specific language governing permissions and      #}
+{# limitations under the License.                                           #}
+
+{# Template fragment for the uninitialized instance warning.                #}
+
+{% load i18n %}
+
+<div>
+  <p>This instance of Person Finder <b>has not been initialized yet</b>. Please go to the <a href="{{env.global_url}}/admin">admin page</a> and click on the <b>setup datastore</b> button.
```
gimite: How about just directly put "Setup datastore" button here? Then you probably don't need to put the button in the admin page.

clobrano:
That could be easier in fact

app/main.py
```diff
@@ -494,6 +496,20 @@ def initialize(self, request, response):
...
     def serve(self):
         request, response, env = self.request, self.response, self.env
+
+        # If the Person Finder instance has not been initialized yet,
+        # prepend to any served page a warning and a link to the admin
+        # page where the datastore can be initialized.
+        if not config.get('initialized'):
+            if request.get('operation') == 'setup_datastore':
+                setup_pf.setup_datastore()
+                self.redirect(env.global_url + '/')
```
gimite: You should "return" after this line to skip the rest of the process.

app/resources/setup_datastore.html.template
```diff
@@ -0,0 +1,21 @@
...
+{# See the License for the specific language governing permissions and      #}
+{# limitations under the License.                                           #}
+
+{# Template fragment for the uninitialized instance warning.                #}
+
+{% load i18n %}
+
+<div>
+<form method="post" class="main">
+  <input type=hidden name=operation value="setup_datastore">
```
gimite: Can you enclose all attribute values (hidden, operation) in double-quote (" ")?

app/main.py
```diff
@@ -35,8 +35,10 @@
 import resources
 import utils
 import user_agents
+import setup_pf
 
 
+log = logging.getLogger('main')
```
gimite: Is this used?

app/resources/setup_datastore.html.template
clobrano: Sure, done

app/main.py
clobrano: Sorry, that escaped my check.

app/resources/setup_datastore.html.template
```diff
@@ -0,0 +1,21 @@
...
+
+{# Template fragment for the uninitialized instance warning.                #}
+
+{% load i18n %}
+
+<div>
+<form method="post" class="main">
+  <input type="hidden" name="operation" value="setup_datastore">
+  Please initialize this Person Finder instance
+  <input type=submit value="Setup datastore">
```
gimite: Enclose "submit" here with double-quote as well.

gimite:
Looks good to me. Merged. Thanks for the patience and sorry for taking so long! This is very useful.

clobrano:
No worries, every mark was right. Thank you! (looking for something else to
do ;))

Issue #192, opened 2015-10-26
Migrate to remote API using OAuth


gimite:
The one we are using, based on ClientLogin, will be shut down:
https://cloud.google.com/appengine/docs/python/tools/remoteapi#using_the_remote_api_in_a_local_client

PR #201, opened 2015-11-06
Migrate to remote API using OAuth. #192

gimite:


app/app.yaml
```diff
@@ -19,13 +19,19 @@ api_version: 1
 threadsafe: false
 
 handlers:
+# Remote API handlers. Note that login: admin must not be specified because the
+# handler performs its own authentication. /personfinder/remote_api is defined
+# for google.org/personfinder, which only accepts paths starting with
+# /personfinder.
+- url: /remote_api
```
skywhale: assuming wildcard didn't work?

gimite: It would work too, but I thought that it may be better to be specific rather than allowing it to match random paths e.g., /japan/view/remote_api or something. But I don't have strong opinion.

skywhale:
Thanks for fixing this!!! LGTM

app/app.yaml
skywhale: makes sense.

skywhale:
Merged.

PR #202, opened 2015-11-07
Updated buttons to be rectangular. Updated UI for mark duplicates but…

Jessie-Newman:
Buttons are now rectangular. "Mark duplicate" buttons have new-ui style. "Mark duplicate" page uses fields-table css styling. The comparison table in the "mark duplicates" page is not new-ui but it didn't conform to the old UI either.

### DIFF
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -493,7 +493,8 @@ td.submit {
   border: 2px solid #e6e6e6;
 }
 
-.seek.results.border .query.seek {
+.seek.results.border .query.seek,
+.seek.results.border .query.provide{
   border: 0px;
   margin: 0px;
   padding: 0px;
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -102,8 +102,7 @@ pre {  /* code samples */
   font-size: 12px;
 }
 
-.button input,
-.button-cancel input {
+.button input {
   border-style: none;
   color: white;
   display: block;
@@ -111,15 +110,15 @@ pre {  /* code samples */
   padding: 5px 15px;
 }
 
-.button-cancel input:disabled {
+.button input:disabled {
   background: #9e9e9e;
 }
 
-.button input {
+.button-blue input {
   background: #4285f4;  
 }
 
-.button-cancel input {
+.button-red input {
   background: #db4437;
 }
 
@@ -130,8 +129,7 @@ pre {  /* code samples */
   margin-right: -10px;
 }
 
-.button-box .button,
-.button-box .button-cancel {
+.button-box .button {
   margin-left: 10px;  
   margin-right: 10px;
 }
@@ -570,8 +568,7 @@ img.icon {
 
 }
 
-.note.fields-table .button input,
-.note.fields-table .button-cancel input {
+.note.fields-table .button input {
   margin: 10px auto;
 }
 
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -104,12 +104,15 @@ pre {  /* code samples */
 
 .button input,
 .button-cancel input {
-  border-radius: 28px;
   border-style: none;
   color: white;
   display: block;
-  font-size: 20px;
+  font-size: 16px;
   padding: 5px 15px;
+}
+
+.button-cancel input:disabled {
+  background: #9e9e9e;
 }
 
 .button input {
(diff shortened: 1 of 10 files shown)


app/resources/css-default.template
```diff
@@ -491,6 +493,22 @@ td.submit {
   border: 2px solid #e6e6e6;
 }
 
+.seek.results.border .query.seek,
+.seek.results.border .query.provide{
+  border: 0px;
```
kevinalle: you don't need "px" for value 0

app/resources/results.html.template
```diff
@@ -82,214 +82,233 @@
...
-                {{result.home_state}}&nbsp;&nbsp;
-              {% endif %}
-              {% if result.home_postal_code %}
-                {{result.home_postal_code}}&nbsp;&nbsp;
-              {% endif %}
-          </div>
+        {% endifchanged %}
+        <div class="resultItem" data-url="{{result.view_url}}" data-status="{{result.latest_status}}"
+            onclick="on_result_item_click(event, this);">
+          {# No whitespace inside .resultImageContainer. #}
```
kevinalle: I think you can achieve this with {% spaceless %}
https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#spaceless

app/resources/results.html.template
```diff
@@ -82,214 +82,233 @@
...
       </div>
+      {% if env.enable_dup_mode %}
+        <div id='dup_form' style='display:none'>
+          {% trans "Select up to 3 records to mark as duplicate:" %}
+          <form action="{{env.repo_path}}/multiview" method="get" onsubmit="mark_dup()">
+            {{env.hidden_input_tags_for_preserved_query_params|safe}}
+            <input type='hidden' name='id1' id='id1' value=''>
+            <input type='hidden' name='id2' id='id2' value=''>
+            <input type='hidden' name='id3' id='id3' value=''>
+            <div class='button-cancel'>
```
kevinalle: The submit button has class button-cancel? seems odd..

app/resources/results.html.template
Jessie-Newman: I'm not actually sure what's going on here (I didn't edit this section, just indented it).
Is this stating that we shouldn't put whitespace into the div? Adding whitespace doesn't seem to cause any problems.

app/resources/results.html.template
Jessie-Newman: button-cancel makes the button red. I thought it would be appropriate for "Mark records as duplicate" to be red as it indicates there's an error in the data.

Would you prefer me to change it to blue? Or to chose a different name for the CSS class?

app/resources/css-default.template
Jessie-Newman: Done.

app/resources/results.html.template
gimite: Hm I don't know the intention of the comment either. Anyway it should be fine since you just indented it.

Jessie-Newman: Not sure if the email went through but I've responded to all the comments.

Jessie-Newman: Friendly ping :) Still waiting on responses to some of my comments.

kevinalle: omg, so sorry. I haven't received emails from this at all. I have to fix my filters.

I meant you can use {% spaceless %} instead of closing the html tags in a separate line. But fine to leave it this way if it already was.

app/resources/results.html.template
kevinalle: I'm fine with the button being the color you think is best. But maybe the css could have a different name, yeah.

app/resources/results.html.template
gimite: FYI You can receive Github notification to a non-default address only for projects in "google" organization here: https://github.com/settings/notifications

I now vaguely remember that this space removal is probably to avoid weird rendering in some old feature phone browsers.

app/resources/results.html.template
Jessie-Newman: Instead of button and button-cancel there's now button (has things like padding, etc), button-red, and button-blue.

app/resources/results.html.template
Jessie-Newman: Leaving as is

Jessie-Newman: Thanks for reviewing. Responded to comments

kevinalle: Cool. LGTM!

gimite:
Thanks! Merged to new-ui branch.

@Jessie-Newman Do you have any remaining changes you want to make for short term? Otherwise we will review the latest UI, merge it to the master branch, and push it to production.

Jessie-Newman:
I don't have anything else.

gimite:
Got it. Thanks!

2016年1月21日(木) 0:18 Jessie Newman <notifications@github.com>:

PR #57, opened 2015-06-11
Avoid setting an arbitrary URL as the target of /reveal.

gimite:
Fix for #44 .

PR #213, opened 2015-11-12
Fix an issue that /reveal can still redirect users to other domains.

gimite:
I tried to fix it in #57 but it was incomplete. It's still possible to specify target=.malicious.com to redirect users to google.org.malicious.com.

### DIFF
--- a/app/reveal.py
+++ b/app/reveal.py
@@ -102,9 +102,16 @@ class Handler(BaseHandler):
         if captcha_response.is_valid or self.env.test_mode:
             signature = sign(self.params.content_id)
             # self.params.target contains only the path part of the URL e.g.,
-            # "/test/view?...". Puts our own scheme and netloc to make it an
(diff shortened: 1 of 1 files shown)


app/reveal.py
```diff
@@ -102,9 +102,16 @@ def post(self):
...
-            # the target.
+            # "/test/view?...".
+            #
+            # - We verify that self.params.target starts with '/'.
+            # - We put our own scheme and netloc to make it an absolute URL.
+            #
+            # These two are required to allow only URLs in our domain as the
+            # target, avoiding this to be used as a redirector to aid phishing
+            # attacks.
+            if not self.params.target or self.params.target[0] != '/':
```
skywhale: self.params.target.startswith('/') is more intuitive?

skywhale:
LGTM

app/reveal.py
gimite: Done.

gimite:

PR #215, opened 2015-11-12
Implement zero-rating mode except for reCaptcha and ToS page. #7

gimite:
I disabled all services I can think of which loads external resource, except for reCaptcha and ToS. But I'm not really sure these are all. I will check it with production after submitting this.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -395,6 +395,19 @@ def setup_env(request):
         # support UTF-8 but don't render UTF-8 symbols such as u'\xab'.
         env.back_chevron = u'<<'
 
+    env.enable_maps = (
+        env.enable_javascript
+        and not env.config.zero_rating_mode
+        and env.maps_api_key)
+    env.enable_analytics = (
+        env.enable_javascript
+        and not env.config.zero_rating_mode
+        and env.analytics_id)
+    env.enable_translate = (
+        env.enable_javascript
+        and not env.config.zero_rating_mode
+        and env.config.translate_api_key)
+
     # Repo-specific information.
     if env.repo:
         # repo_url is the root URL for the repository.
--- a/app/resources/admin.html.template
+++ b/app/resources/admin.html.template
@@ -899,6 +899,36 @@
   </fieldset>
 
   <fieldset>
+    <legend>Zero-rating</legend>
+
+    <div class="config">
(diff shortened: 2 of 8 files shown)


app/main.py
```diff
@@ -395,6 +395,19 @@ def setup_env(request):
         # support UTF-8 but don't render UTF-8 symbols such as u'\xab'.
         env.back_chevron = u'<<'
 
+    env.enable_maps = (
+        env.enable_javascript
+        and not env.config.zero_rating_mode
+        and env.maps_api_key)
+    env.enable_analytics = (
```
skywhale: I don't have full context here, but external api calls are totally prohibited with zero-rating mode, or is this to save bandwidth?

app/resources/admin.html.template
```diff
@@ -899,6 +899,36 @@
...
 
   <fieldset>
+    <legend>Zero-rating</legend>
+
+    <div class="config">
+      <p>Sometimes mobile carriers "zero-rate" our site so that users can see our site without normal data access fee. It is often applied per domain name. But when users access Person Finder, it also loads resource from other domains e.g., Google Maps, Google Analytics, Google Translate by default.</p>
+
+      <p>If you enable zero-rating mode, it disables all those features which loads resource from external domains, so that the users from the carriers can access Person Finder completely for free.</p>
+
+      <p>CAVEAT: Even if you enable zero-rating mode, the photos attached to person records or notes may be loaded from external domains, if the author has specified a URL of the photo instead of uploading it.</p>
```
skywhale: Do you want to file a feature request for routing such images through personfinder domain?

app/main.py
skywhale: Sorry, I just read your description below. My question answered. Do we currently use Analytics for understanding user behavior? I assume not, and it's fine to disable this for zero-rating users?

skywhale:
LGTM. Thank you!

app/main.py
gimite: We do use Analytics to analyze the usage. So there's a trade off between enabling zero-rating mode and giving up analysis via Analytics, which should be decided per crisis.

app/resources/admin.html.template
gimite: Good point. Done: https://github.com/google/personfinder/issues/218

Issue #214, opened 2015-11-12
Repository admin page rendering is broken if one of the config contains "</script>"


gimite:
We need to escape ```"</script>"```.
https://google.org/personfinder/test/admin
is currently broken.

PR #216, opened 2015-11-13
Fix a bug that repository admin page rendering is broken if one of the config contains "</script>". #214

gimite:
We need to escape ```"</script>"``` to avoid terminating ```<script>``` tag unintentionally.
"escapejs" filter escapes it to "\u003C/script\u003E\u0022".

Also moved the constants outside the function to avoid parsing it for every function call.

### DIFF
--- a/app/resources/admin.html.template
+++ b/app/resources/admin.html.template
@@ -42,7 +42,11 @@
 
 //TODO: move to object-oriented!!
 
-  var LANGUAGE_EXONYMS = {{language_exonyms_json|safe}};
+  var LANGUAGES = JSON.parse(
+      "{{view_config_json.language_menu_options|default:'[]'|escapejs}}");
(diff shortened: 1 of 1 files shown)


app/resources/admin.html.template
```diff
@@ -42,7 +42,11 @@
 
 //TODO: move to object-oriented!!
 
-  var LANGUAGE_EXONYMS = {{language_exonyms_json|safe}};
+  var LANGUAGES = JSON.parse(
```
skywhale: (optional) It may be less error-prone to use a single variable to pass json data from the controller to the view. e.g. view_config_json, language_exonyms_json and everything else packed in json_data.

var JSON_DATA = JSON.parse("{{json_data|escapejs}}");

skywhale:
LGTM. Thanks for the fix :)

app/resources/admin.html.template
gimite: Added a TODO.

gimite:

Issue #211, opened 2015-11-10
Remove get_secret()


gimite:
In main.py:

    # TODO(kpy): Make these global config settings and get rid of get_secret().
    env.analytics_id = get_secret('analytics_id')
    env.maps_api_key = get_secret('maps_api_key')

PR #217, opened 2015-11-13
Removed get_secret(). #211

aacn500:
Replaced get_secret() with config.get(). I fixed the relevant test by using config.set(), and I can't see anywhere else that is affected by the change.
First time working with git and open source in general, apologies if I missed something obvious!


gimite:
Thanks for the feedback! Can you also remove "Secret" class in https://github.com/google/personfinder/blob/master/app/model.py#L752 ?

aacn500:
Should reset_database() in https://github.com/google/personfinder/blob/master/tools/setup_pf.py#L40 still keep what used to be Secrets (analytics_id and maps_api_key), or is that no longer required functionality?

gimite:
It seems reset_database() is used in server test. Can you simply remove "Secret" from reset_database() and try running the server tests by this command?:

    tools/server_tests

Also, I found some other parts which still refer to secrets. Can you remove these too?:

- get_secret() in app/utils.py
- tests/server_test_cases/secret_tests.py (You can replace secrets by config here as well)
- import in app/reveal.py (I believe you can just delete it because it looks not used)

aacn500:
These commits pass all server tests for me, and I can see no more references to the Secret class when running ```git grep -in secret```.

gimite:
Thanks! Looks good to me.

Can you read through:
https://github.com/google/personfinder/blob/master/CONTRIBUTING.md
and sign this online form?:
https://cla.developers.google.com/about/google-individual

It is required by our company to accept patches from you.

gimite:
As my personal note: I must migrate secrets into configs in our production before pushing this change to production.

aacn500:
CLA has been submitted.

gimite:
Merged. Thanks for  your contribution!

> As my personal note: I must migrate secrets into configs in our production before pushing this change to production.

Done.

PR #220, opened 2015-11-19
Make ToS page a proxy to the actual ToS page, to make it zero-rating mode friendly. #7

gimite:


app/resources/tos.html.template
```diff
@@ -1,14 +1,40 @@
...
+
+    {{article_html|safe}}
+
+  {% else %}{% if tos_url %}
+
+    <h1>Terms of Service</h1>
+
+    {# In case we fail to fetch/parse the ToS page (e.g., the page has       #}
+    {# changed its layout), we fall back to show a link to the page instead. #}
+    <p><a href="{{tos_url}}">Click here to see the terms of service (external site)</a></p>
```
skywhale: And also add auto redirect?

app/tos.py
```diff
@@ -0,0 +1,68 @@
...
+        # just redirecting to the ToS page because we need to host the content
+        # in our domain in zero-rating mode. See "Zero-rating" section of the
+        # admin page (app/resources/admin.html.template) for details of
+        # zero-rating mode.
+        try:
+            if self.config.tos_url:
+                res = urllib2.urlopen(self.config.tos_url)
+                tos_html = res.read()
+                doc = lxml.etree.HTML(tos_html)
+                articles = doc.xpath("//div[@class='maia-article']")
```
skywhale: Can you provide the URL of the ToS page? I was wondering why just a redirection isn't good enough. The logic here is rather complex and a pain to maintain. ToS pages are in reality rarely visited, and I don't think a hop of redirection is bad.

app/resources/tos.html.template
gimite: It may require payment, so I think it's better not to redirect automatically. What do you think?

app/tos.py
gimite: Here it is:
https://www.google.com/policies/terms/?hl=en

The motivation is, as explained in the comment, to make it zero-rating mode friendly. If we just automatically redirect, users may need to pay fee unexpectedly when google.org is zero-rated, which is bad.

skywhale: sorry that I somehow completely missed that.

app/resources/tos.html.template
skywhale: I see. That makes sense.

BTW should we l10n this page? Also, "external site" is probably not good enough for most users. maybe add a more friendly message e.g. "clicking on the link may incur additional cost from your mobile carrier as it takes you to an external site." Or, do you think it's not necessary since this is only for the supposedly-minor error case?

gimite: I'm not really sure. Let's discuss both i18n and wording in the next meeting with @nmakida .

I understand your point, but I feel saying "additional cost" may be also confusing, because it gives impression that we require special cost compared to other normal sites. Truth is that the same data fee as other sites are applied, but it's not free as google.org.

Also, I also embed " (external site)" text to every external links at app/tos.py:56. Here we cannot embed too long text.

skywhale: Based on the conversation, I think the current state is good as is, at least for now. I'm going to LGTM.

skywhale:
LGTM. Wording etc, we may still want to follow up with nmakida@, though.

gimite:
Thanks, then I will merge it for now. I'll make a follow-up wording changes when needed.

Issue #221, opened 2015-11-30
[BUG] view.Handler.post lacks error-checking for person


nilsbarth:
In:
app/view.py
view.Handler.post is missing the error-checking for person that is present in view.Handler.get.

Specifically, it should include the following lines from `get`:
```
        # Check the request parameters.
        if not self.params.id:
            return self.error(404, 'No person id was specified.')
        try:
            person = Person.get(self.repo, self.params.id)
        except ValueError:
            return self.error(404,
                _("This person's entry does not exist or has been deleted."))
        if not person:
            return self.error(404,
                _("This person's entry does not exist or has been deleted."))
```

Currently `post` has:
```
        person = Person.get(self.repo, self.params.id)
        if person.notes_disabled:
```
...which fails if person is `None`.

(Easy to fix.)

PR #224, opened 2015-12-01
Add an error check when the user try to add a note for a non-existing person record. #221

gimite:


tests/scrape.py
```diff
@@ -170,8 +170,17 @@ def fetch(url, data='', agent=None, referrer=None, charset=None, verbose=0,
...
-        data = urlencode(data)
+
+    if not data:
+        data_str = ''
+    elif isinstance(data, str):
+        data_str = data
+    elif isinstance(data, unicode):
+        data_str = data.encode('utf-8')
+    elif isinstance(data, dict):
+        data_str = urlencode(data)
```
skywhale: could it also be a dict of unicodes? I haven't followed through the code.

app/view.py
```diff
@@ -182,6 +182,9 @@ def post(self):
                        '"believed_dead".'))
 
         person = Person.get(self.repo, self.params.id)
+        if not person:
```
skywhale: Nice! It's a bit surprising that he weren't checking the condition.

skywhale:
LGTM. Thanks for the bug fix!

tests/scrape.py
gimite: Reading the code, urlquote(), used in urlencode(), supports both str and unicode. Added a comment.

gimite:
Thanks. Merged.

Issue #223, opened 2015-11-30
[BUG] Write should implement get (even if as error), or gate


nilsbarth:
The api.Write class in:
app/api.py
...does not implement get, so a GET request that is dispatched to it results in a webapp2 error.

Better would be to explicitly handle this, like throwing an NotImplementedError or similar.

PR #225, opened 2015-12-01
Return HTTP status 405 instead of 500 for unimplemented HTTP methods. #223

gimite:


skywhale:
LGTM. Thanks for the fix. It's interesting that it's not the default behavior of webapp.RequestHandler.

gimite:
Thanks. Merged.

> LGTM. Thanks for the fix. It's interesting that it's not the default behavior of webapp.RequestHandler.

I agree. These methods are simply undefined in webapp.RequestHandler.

PR #231, opened 2015-12-09
Add xpath() method to the document returned by scrape module powered by lxml

gimite:


tests/scrape.py
```diff
@@ -1030,6 +1034,46 @@ def enclosing(self, tagname=None, content=None, enders=[], **attrs):
         c = content is not None and ' with content %r' % content or ''
         raise ScrapeError('no %s found%s%s' % (tag, a, c))
 
+class Document(Region):
+    """A document returned as an HTTP response.
+    """
+
+    def __init__(self, content_str, headers, charset=None):
```
skywhale: Please document that not passing charset will result in xpath() not working. I'd assume that the charset will be automatically determined by default.

skywhale: Please document the behavior when charset=None

skywhale:
LGTM. Thank you so much!

tests/scrape.py
gimite: Done. I also refactored the code to guess the charset from content-type header in Document.__init__ which was done in fetch().

gimite:
Thanks. Merged.

PR #234, opened 2015-12-11
Support lxml element in scrape.Session.follow().

gimite:


tests/scrape.py
```diff
@@ -1143,4 +1167,10 @@ def getnumber(text):
                 continue
     raise ScrapeError('no number found in %r' % text)
 
+def get_all_text(elem):
+    """Returns all texts in the subtree of the element.
```
skywhale: sorry for being inconsistent from the last review, but I actually think that we should call out the type of the element. don't need to say it's etree.Element. "an element returned by xpath() or cssselect()" is probably good enough.

tests/scrape.py
```diff
@@ -353,19 +353,43 @@ def back(self):
...
+                raise ScrapeError('link %r not found' % anchor)
+            if not link.get('href', ''):
+                raise ScrapeError('link %r has no href' % link)
+            href = link['href']
+
+        else:
+            if isinstance(anchor, basestring) or type(anchor) is RE_TYPE:
+                link = None
+                for l in (context or self.doc).cssselect('a'):
+                    if get_all_text(l) == anchor:
```
skywhale: should we also handle RE_TYPE?

PR #235, opened 2015-12-11
Change helper methods in person_note_tests.py to use cssselect().

gimite:
Except for forms.


skywhale:
Nice :) LGTM

I wish we could omit strip() here and there...

gimite:
Yeah that's a bit annoying, but I couldn't find a option to strip automatically in lxml. One option is to define a wrapper function.

gimite:
Thanks. Merged.

PR #236, opened 2015-12-14
Fix a bug that scrape.Session.follow doesn't work with compiled regexp.

gimite:
Also fixed/improved some method docs.


skywhale:
LGTM. Merged.

PR #242, opened 2016-01-28
Fix server test failures caused by the new UI change.

gimite:
They were broken mainly because the form to add notes was separated from the "view" page.

### DIFF
--- a/app/add_note.py
+++ b/app/add_note.py
@@ -42,12 +42,14 @@ class Handler(BaseHandler):
             return self.error(404, _('No person id was specified.'))
         try:
             person = Person.get(self.repo, self.params.id)
+        # TODO(ichikawa) Consider removing this "except" clause.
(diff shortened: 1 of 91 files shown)


app/add_note.py
```diff
@@ -42,12 +42,14 @@ def get(self):
             return self.error(404, _('No person id was specified.'))
         try:
             person = Person.get(self.repo, self.params.id)
+        # TODO(ichikawa) Consider removing this "except" clause.
```
skywhale: I also checked the code and I don't think it will raise ValueError. Why not just remove it?

skywhale:
LGTM. Thanks for the massive cleanup!!

app/add_note.py
gimite: Will do in a separated pull request because it's not related.

PR #249, opened 2016-04-19
Add Spanish localization support for SMS interface.

gimite:


app/api.py
```diff
@@ -654,26 +665,44 @@ def post(self):
...
+                    % {'url': 'google.org/personfinder/%s?ui=light' % repo})
             responses.append(
-                'All data entered in Person Finder is available to the public '
-                'and usable by anyone. Google does not review or verify the '
-                'accuracy of this data google.org/personfinder/global/tos')
-        elif self.config.enable_sms_record_input and add_self_m:
-            name_string = add_self_m.group(1).strip()
+                _('All data entered in Person Finder is available to the '
+                  'public and usable by anyone. Google does not review or '
+                  'verify the accuracy of this data '
```
skywhale: you intentionally left out the ending period?

skywhale:
LGTM

app/api.py
gimite: I don't know why. Maybe we wanted to keep it as shorter as possible :) I prefer to keep it as is in this change because it was like this before this change.

gimite:

PR #250, opened 2016-04-19
Detect dev app server correctly even if it's not localhost.

gimite:


app/utils.py
```diff
@@ -474,6 +474,9 @@ def strip_url_scheme(url):
     _, netloc, path, query, segment = urlparse.urlsplit(url)
     return urlparse.urlunsplit(('', netloc, path, query, segment))
```
skywhale: 2 blank lines between top-level functions

skywhale:
LGTM

app/utils.py
gimite: Done.

gimite:

PR #261, opened 2016-07-07
Add a side bar to admin pages.

gimite:
#121

This provides consistent way to navigate between admin pages.

s6yjixv5eqo

### DIFF
--- a/app/admin_resources.py
+++ b/app/admin_resources.py
@@ -139,10 +139,12 @@ class Handler(utils.BaseHandler):
             anchor = lang + ': ' + const.LANGUAGE_EXONYMS.get(lang, '?')
             crumbs.append((anchor, (bundle_name, name, lang)))
         last = crumbs[-1][1]
-        return '<div class="nav">%s</div>' % (' &gt; '.join(
-            '<a class="%s" href="%s">%s</a>' %
-            (args == last and 'sel', self.get_admin_url(*args), html(anchor))
-            for anchor, args in crumbs))
+        anchors = (
+            ['<a href="%s">Admin page</a>' % self.get_url('admin')] +
+            ['<a class="%s" href="%s">%s</a>' %
+                (args == last and 'sel', self.get_admin_url(*args), html(anchor))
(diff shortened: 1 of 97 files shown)


app/admin_resources.py
```diff
@@ -139,10 +139,12 @@ def format_nav_html(self, bundle_name, name, lang):
...
             crumbs.append((anchor, (bundle_name, name, lang)))
         last = crumbs[-1][1]
-        return '<div class="nav">%s</div>' % (' &gt; '.join(
-            '<a class="%s" href="%s">%s</a>' %
-            (args == last and 'sel', self.get_admin_url(*args), html(anchor))
-            for anchor, args in crumbs))
+        anchors = (
+            ['<a href="%s">Admin page</a>' % self.get_url('admin')] +
+            ['<a class="%s" href="%s">%s</a>' %
+                (args == last and 'sel', self.get_admin_url(*args), html(anchor))
```
kevinalle: args == last and 'sel'
might set class="False" on the element, no?

app/main.py
```diff
@@ -38,6 +39,35 @@
...
+        try:
+            return [
+                utils.Struct(
+                    repo=repo,
+                    url=utils.get_repo_url(self.request, repo) + '/admin')
+                for repo in sorted(model.Repo.list())]
+        except:
+            # Logs the exception here because exceptions thrown during template
+            # variable evaluation is silently ignored.
+            logging.exception('exception')
```
kevinalle: Are you sure you want to log the string 'exception'? and not the actual exception?

kevinalle:
Please see comments. Otherwise LGTM!

app/admin_resources.py
gimite: Good catch, though it was a bug before my change :) Fixed.

app/main.py
gimite: logging.exception() logs the current exception by default. Added a comment.

gimite:
Thanks! Merged.

I also needed to fix some tests.

PR #270, opened 2016-08-02
Add a link to privacy policy.

gimite:


gimite:
Sorry I changed where to put the link based on @nmakida 's ask. Can you review again?

app/resources/base.html.template
```diff
@@ -119,7 +119,16 @@
             {% trans "To delete your data, access Person Finder from your desktop." %}
             <br>
           {% endif %}
+
           {% trans "PLEASE NOTE: All data entered is available to the public and usable by anyone.  Google does not review or verify the accuracy of this data." %}
+
+          {% if config.privacy_policy_url %}
+            {# TODO(ichikawa) Proxy this in the same way as TOS. #}
+            {% blocktrans with config.privacy_policy_url as privacy_policy_url and env.target_attr|safe as target_attr %}
+              Learn more about Google
```
kevinalle: this doesn't need to be translated?

gimite: It should be translated, so it's inside {%blocktrans%}. I used {%blocktrans%} instead of {%trans%} because it contains some placeholders.

kevinalle:
Ah, I see.
LGTM

gimite:
Thanks! Merged.

PR #274, opened 2016-08-10
Automatically cd to the root directory on tests.

gimite:
Because some tests assumes the current directory is the root directory.


Youki:
LGTM

gimite:
Thanks! Merged.

PR #276, opened 2016-08-15
Edit the structure of the form

Stephanie1125:
edit the structure of the form to make the required-to-fill-in part show before the non-required part (prepare for the expansion function)

### DIFF
--- a/app/resources/add_note_base.html.template
+++ b/app/resources/add_note_base.html.template
@@ -15,127 +15,29 @@
 {% if not markdup %}
   <input type="hidden" name="add_note" value="yes">
   <input type="hidden" name="id" value="{{params.id}}">
-{% endif %}
-
-<div class="fields-table note" role="complementary">
-  <div class="header section">
-    <h1>
-      {% if not markdup %}
-        {% if person.full_name %}
-          {% blocktrans with person.full_name as full_name %}Status of {{full_name}} {% endblocktrans %}
-        {% else %}
-          {% trans "Status of Person" %}
-        {% endif %}
-      {% else %}
-        {% trans "Mark records as duplicate" %}
-      {% endif %}
-    </h1>
-  </div>
-  {% if not markdup %}
-    <div class="section">
-      <h2>Status</h2>
-      <div class="status subsection">
-        <h3>
-          {% trans "Status of this person:" %}
-        </h3>
-        <div class="field">
-          <select id="status" name="status">
-            {% for status in env.status_options %}
-              <option value="{{status.value}}"
-                  {% ifequal params.status status.value %}
-                    selected='selected'
-                  {% endifequal %}>
-                {{status.text}}
-              </option>
-            {% endfor %}
-          </select>
-        </div>
-        <div class="end-multi-columns"></div>
-      </div>
-
-      <div class="author_made_contact subsection">
-        <h3 class="mandatory">
-          {% trans "Have you personally talked with this person AFTER the disaster? (required)" %}
-        </h3>
-        <div class="field">
-          <input type="radio" name="author_made_contact"
-              value="yes" id="author_made_contact_yes"
-              onchange="update_contact()"
-              onclick="update_contact()"
-              {% ifequal params.author_made_contact "yes" %}
-                checked="checked"
-              {% endifequal %}>
-          <label class="option" for="author_made_contact_yes">
-            {% trans "Yes" %}
-          </label>
-        </div>
-        <div class="field">
-          <input type="radio" name="author_made_contact"
-              value="no" id="author_made_contact_no"
-              onchange="update_contact()"
-              onclick="update_contact()"
-              {% ifnotequal params.author_made_contact "yes" %}
-                checked="checked"
-              {% endifnotequal %}>
-          <label class="option" for="author_made_contact_no">
-            {% trans "No" %}
-          </label>
-        </div>
-        <div class="end-multi-columns"></div>
-      </div>
-
-      <div class="location subsection">
-        <h3>
-          {% trans "Last known location" %}
-        </h3>
-        <div class="hint">
-          {% if env.enable_javascript and env.maps_api_key %}
-            {% trans "Type an address or open the map below and indicate the location by clicking on the map." %}
-          {% else %}
-            {% trans "Type an address." %}
-          {% endif %}
-        </div>
-        <div class="field">
-            <textarea name="last_known_location"
-                id="clickable_map_location_field"
-                class="long-text-input"
-                rows=2>{{params.last_known_location}}</textarea>
-        </div>
-        <div class="end-multi-columns"></div>
-        {% if env.enable_javascript and env.maps_api_key %}
-          <div class="field map-container">
-            <div>
-              <span id="clickable_map_show_link">
-                <a href='#' onclick="toggleClickableMap('clickable_map'); return false;">
-                  {% trans "Show Map" %}</a>
-              </span>
-              <span id="clickable_map_hide_link" style="display: none">
-                <a href='#' onclick="toggleClickableMap('clickable_map'); return false;">
-                  {% trans "Hide Map" %}</a>
-                &nbsp;&nbsp;
-                <a href='#' id="switch_to_latlng_link"
-                    onclick="toggleLatLngAndAddress('clickable_map'); return false;">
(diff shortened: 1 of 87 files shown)


gimite:
Can you also attach a screenshot of the new form UI?

Stephanie1125:
Photos for new forms

i have information about someone
adding notes for person

app/resources/add_note_base.html.template
```diff
@@ -15,25 +15,11 @@
...
-      {% else %}
-        {% trans "Mark records as duplicate" %}
-      {% endif %}
-    </h1>
-  </div>
+<div class="status section">
   {% if not markdup %}
-    <div class="section">
-      <h2>Status</h2>
+    <center><h2>Status</h2></center>
```
gimite: Define a CSS rule in css-default.template to centerize h2 element instead of using ```<center>```. Make it a rule for e.g., ".fields-table h2" so that it won't affect h2 elements in other places.

In principle, HTML file should only contain the content and its structure. And CSS file should define the look-and-feel. In that principle, usage of tags like ```<center>```, ```<font>```, etc. are discouraged.

app/resources/add_note_base.html.template
```diff
@@ -129,37 +134,13 @@
...
-  </div>
   
-  {% if not markdup %}
-
-    
+  {% if not markdup %}    
     {% if env.enable_photo_upload %}
       <div class="photo section">
-        <h2>Photo</h2>
+        <h3>Photo</h3>
```
gimite: Do you have any reason to make it h3 instead of h2?

app/resources/create.html.template
```diff
@@ -28,7 +28,7 @@
     {{env.hidden_input_tags_for_preserved_query_params|safe}}
     <input type="hidden" name="role" value="{{params.role}}">
     <div class="create" role="form">
-
+    <div class="column centered">
       <div class="column start">
```
gimite: I believe you no longer need this "column start" element. Can you try deleting it?

"column start" is used together with "column end" to define a multi-column layout.

app/resources/create.html.template
```diff
@@ -40,15 +40,10 @@
               {% endifequal %}
             </h1>
           </div>
-          <div class="name section">
+          <div class="identifyinfo section">
```
gimite: Maybe just "identity" is OK. Or it should be "identifying-info".

You should avoid abbreviations unless it's commonly used e.g., info (for readability). And you should put a dash "-" between words for class names. (HTML class/id names in Person Finder is very inconsistent, so you would see other namings too :P)

app/resources/create.html.template
```diff
@@ -40,15 +40,10 @@
...
             <h2>
-              {% trans "Identifying information" %}
+            <center>{% trans "Identifying information" %}</center>
             </h2>
-
-            <div class="subsection">
-              <h3>
-                {% trans "Name" %}
-              </h3>
               <div class="hint">
```
gimite: Align tags in the same level at the same column. e.g., The previous ```<h2>``` and this ```<div>``` are now in the same level, so unindent this ```<div>``` (and the following tags) to the same column as ```<h2>```.

app/resources/create.html.template
```diff
@@ -99,8 +94,15 @@
                   </span>
                 </div>
               {% endif %}
+
+          <div class="name section">
```
gimite: How about using "subsection" here instead of "section"? This is a section inside another section, so it sounds natural to make it "subsection". Let me know if it's not rendered as expected.

Basic expected hierarchy is:
```
fields-table
  section
    subsection (optional)
      field
        label
        value
```

app/resources/create.html.template
```diff
@@ -377,37 +378,12 @@
...
-                </option>
-                {% endfor %}
-              </select>
-            </div>
-            <div class="end-multi-columns"></div>
           </div>
 
-          <div class="source section">
-            <h2>
+          <div class="sourcerecord section">
```
gimite: Would be fine to keep it "source" here. Otherwise make it "source-of-record". "sourcerecord" sounds rather like a record of a source.

app/resources/create.html.template
```diff
@@ -559,15 +572,11 @@
...
       </div>
+      </div>
 
-      {% ifequal params.role "provide" %}
-        <div class="column end">
-          {% include "add_note_base.html.template" %}
-        </div>
-      {% endifequal %}
-
+      <div class="column end"></div>
```
gimite: There is no longer a right column, so you can delete this.

app/resources/create.html.template
```diff
@@ -559,15 +572,11 @@
...
       <div class="end-multi-columns"></div>
 
-    </div>
+      </div>
```
gimite: Fix the indent (unindent two white spaces).

app/resources/create.html.template
```diff
@@ -541,9 +551,12 @@
                 </span>
               </div>
             {% endif %}
-            {% include "subscribe.html.template" %}
             <div class="end-multi-columns"></div>
-          </div>
+            </div>
```
gimite: Fix the indent here as well.

app/resources/add_note_base.html.template
Stephanie1125: The <h3> part is actually hiding inside the expanding part as a subtitle(like the following photo), and I am using the <h2> part for the center title like Identify information/Source of record etc...

note with expand2

gimite:
Is it intentional that there are double horizontal lines sometimes?:
https://screenshot.googleplex.com/YdsKex4hZPu

If not, can you fix it?

gimite:
Looks good, thanks! Made some comments.

app/resources/add_note_base.html.template
gimite: Oh I see. Then looks good.

app/resources/create.html.template
Stephanie1125: already fixed this indent

app/resources/create.html.template
Stephanie1125: already fixed this indent

app/resources/create.html.template
Stephanie1125: already deleted this

app/resources/create.html.template
Stephanie1125: already kept it "source" here

app/resources/create.html.template
Stephanie1125: already changed it to just "identify"

app/resources/create.html.template
Stephanie1125: already unindent this to the same column as h2

app/resources/create.html.template
Stephanie1125: changed all small subsection part to subsection instead of section (name subsection/ address subsection ... etc)

app/resources/create.html.template
Stephanie1125: already deleted it

app/resources/create.html.template
Stephanie1125: (for below comments) already fixed the double horizontal lines too :)

Stephanie1125:
form after edited:

i have info edit

note ediy

Stephanie1125:
Please checked it again! Thank you !

app/resources/create.html.template
```diff
@@ -572,11 +571,9 @@
 
         </div>
       </div>
-      </div>
 
-      <div class="column end"></div>
       <div class="end-multi-columns"></div>
```
gimite: You probably don't need this div element because it's no longer multi-column. Can you try to delete it and let me know if it breaks the layout?

Stephanie1125: nothing changed when i deleted the <div class="end-multi-columns"></div> 
so i delete all <div class="end-multi-columns"></div> in create.html

gimite: Well, end-multi-columns in the end of each set of fields are required per documented here:
https://github.com/google/personfinder/blob/master/app/resources/css-default.template#L243
So only delete the last one I pointed, and keep the remaining ones.

Stephanie1125: I see! okay. I fixed it and only delete the last one.

gimite:
Thanks! LGTM. Merged to new-ui branch. Send pull request against this branch for new UI changes.

Issue #96, opened 2015-07-06
Simplify "I'm this person" flow


skywhale:
Can we simplify the flow for creating a person record for _myself_? (How popular is this use case?)

In particular, auto-fill the following fields:
  Author's name: Person's name
  Status: I'm this person
  Have you personally talked with this person: YES

skywhale:
Let's revisit as part of UI revamp.

Stephanie1125:
I believe this is fixed in the new UI
![info](https://cloud.githubusercontent.com/assets/15961861/18830644/8899735a-841c-11e6-88c8-19060034dd46.png)

gimite:
Yes. Can you put the pull request ID which fixed this issue here and mark this as closed?

PR #279, opened 2016-08-18
Hide some part of the form if the person is inputing his/her own info…

Stephanie1125:
**The goal is to Hide some part of the form if the person is inputing his/her own information**

So inside the form user can choose 

+ I want to input information about this person
form

+ I am this person. (input my own information) **(Some not necessary part of the form is hidden)**
form2

also edit the place if user select 
+ This record is copied from another source
source

One remaining unsolved problem is that when you hide some part of the code,
it will appear a warning message to ask the user to fill all the required part (because the required part in the source section is hidden) :(

I have two idea to solve this problem:
1. define params.own_info to add {% if params.own_info %} to set the author_name automatically to the person's full_name.
2. edit the requred field to set the author name to not required. 

Do you have any better idea?

### DIFF
--- a/app/create.py
+++ b/app/create.py
@@ -58,8 +58,13 @@ class Handler(BaseHandler):
         else:
             if not self.params.given_name:
                 return self.error(400, _('Name is required.  Please go back and try again.'))
-        
-        if not self.params.own_info:
+
+        if self.params.own_info == 'yes':
+            self.params.author_name = self.params.given_name
+            self.params.status = 'is_note_author' 
+            self.params.author_made_contact = 'yes'
+
+        else:
             if not self.params.author_name:
                 if self.params.clone:
                     return self.error(400, _('The Original author\'s name is required.  Please go back and try again.'))
--- a/app/create.py
+++ b/app/create.py
@@ -58,11 +58,13 @@ class Handler(BaseHandler):
         else:
             if not self.params.given_name:
                 return self.error(400, _('Name is required.  Please go back and try again.'))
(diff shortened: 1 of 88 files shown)


Stephanie1125:
Please check it :)
Thank you very much

app/create.py
```diff
@@ -58,11 +58,13 @@ def post(self):
         else:
             if not self.params.given_name:
                 return self.error(400, _('Name is required.  Please go back and try again.'))
-        if not self.params.author_name:
-            if self.params.clone:
-                return self.error(400, _('The Original author\'s name is required.  Please go back and try again.'))
-            else:
-                return self.error(400, _('Your name is required in the "Source" section.  Please go back and try again.'))
+        
+        if not self.params.own_info:
```
kevinalle: I think this will only check whether the own_info was provided. You should check whether it's set to yes or no

Stephanie1125: I see! 
Already edited it to ```if self.params.own_info == "no":```

kevinalle: As discussed in the meeting, you'll have to fill in author_name automatically if inputting own info.
So maybe a better option is to do something like this above:

if self.params.own_info == "yes":
  self.params.author_name = self.params.given_name

..etc

Stephanie1125: okay!

Stephanie1125:
Please check it again :)
Thank you very much

kevinalle:
Cool. LGTM

Stephanie1125:
fixed #96

Issue #280, opened 2016-08-24
Kanji Handling problems(Chinese and Japanese)


matthew-z:
Currently full_text_search will try to romanise kanji, but it has some problems.

A few Chinese kanji has a different pronunciation and romaji when it is a part of a family name. For example, "曾" could be romanised into "Ceng" in most cases, but it must be romanised into "Zeng" when it is used as a family name. This transformation is quite fixed, so a dictionary about it can tackle it easily. In additional, since Hong Kong and Taiwan people use a different sachem to romanise Chinese kanji from Mainland Chinese, the romanise method could be improved for them.

Furthermore, the romanise method cannot handle some rare Japanese family names in some cases. For example, 小鳥遊 should be TAKANASHI instead of  ODORIYU or KOTORIYU

gimite:
Thanks for the feedback!

For Chinese, we currently rely on unidecode library. Do you know of any other library which supports the person name specific romanization, or a list of exceptions available somewhere? Or maybe can you make one?

We have special handling for Japanese, based on a dictionary here (big):
https://raw.githubusercontent.com/google/personfinder/master/app/japanese_name_location_dict.txt
We believe that it covers most of Japanese names. Unfortunately it missed 小鳥遊, so I'm adding it at #281. It's quite rare name, though...

matthew-z:
No problem, I have added some supports for romanising Chinese family name at #282.

PR #282, opened 2016-08-24
to romanize chinese name correctly

matthew-z:
#280


matthew-z:
Thank you for the message.

Actually the code tries to figure out if the input word is a valid full Chinese name: if the word doest start with a valid Chinese surname or the remained part is too long, then the code will use the unidecode instead of the dict. 

Although some corner cases may exist: when a person's given name looks like a full name, then the justification may be wrong. However, I think it is not a problem since Mainland Chinese rarely split their name into two parts when they input Chinese.

Of course I think using the two methods together would be a better solution. Therefore, I just committed some codes to let them appear together. i.e. 
>>> romanize_search_query(u"曾诚")
>>> [u'\u66fe\u8bda', 'Zeng Cheng', 'Ceng Cheng']

Any comments?

gimite:
I understand that **if** you write Chinese full name, you don't put whitespace between the family name and the given name e.g., You don't write "习 近平".

But I heard that it's possible to search for people by just the family name or the given name e.g., Search for [习] or [近平] to find 习近平. I heard from @Stephanie1125 whom you probably know, so I believe it's correct :)

So I believe this should also work:

    >>> romanize_search_query(u"曾")
    >>> [u'\u66fe', 'Zeng', 'Ceng']

Can you maybe remove the word length condition so that it also works for just family names?

Also, how did you create chinese_family_name_dict.txt? Did you list them up yourself? Do you know if it's a comprehensive list, or does it only cover major ones?

matthew-z:
OK, I have removed the length condition.

The dict was found in [family_name_handle](https://github.com/wanghuafeng/family_name_handle/blob/master/data/all_familyname_word_pinyin_484.txt)          I was wondering if we need to cite it explicitly.

It seems that the data is obtained from wikipedia so I think it is not comprehensive. However, according to some statistics, the major 500 surnames can cover about 99% population in China. I think it is enough at this stage, and `unidecode` may produce a correct result when encountering a rare surname.

gimite:
Do you know what is the license of the original data (e.g., Apache License)? If not, maybe can you ask the original author what the license is?

matthew-z:
The author did not mention anything about license, so I opened an issue to ask him.

matthew-z:
Hi

The author have uploaded a MIT license.

app/script_variant.py
```diff
@@ -168,6 +215,14 @@ def romanize_search_query(word):
         hiragana_word = jautils.normalize(word)
         romanized_words.append(jautils.hiragana_to_romaji(hiragana_word))
 
-    romanized_word = unidecode(word)
-    romanized_words.append(romanized_word.strip())
+    # if the name is a Chinese name and the chinese_romanize can produce
+    # a different result, append the result to the romanzied_words with
+    # unidecode results together
+    unidecode_romanize_word = unidecode(word).strip()
+    if is_chinese_person_name(word):
```
gimite: It looks waste to split Chinese names twice (in is_chinese_person_name and romanize_chinese_person_name). How about removing this check, because romanize_chinese_person_name already has a logic when it fails to split?

app/chinese_family_name_dict.txt
```diff
@@ -0,0 +1,484 @@
+怀	huai
```
gimite: Can you add a comment where this data comes from? Maybe add comment lines starting with "#", and add a logic to skip lines starting with "#".

app/script_variant.py
```diff
@@ -3,7 +3,6 @@
 import jautils
 
 from unidecode import unidecode
-
```
gimite: Keep this blank line, which separates third-party modules and standard modules.

app/script_variant.py
```diff
@@ -145,6 +145,53 @@ def romanize_word_by_unidecode(word):
...
+    """
+    This method tries to split a chinese name into two parts: surname and lastname
+    Args:
+        word: a chinese name string
+    Returns:
+        surname, lastname if it is a valid chinese name
+        else None, None
+    """
+    word = word.replace(" ", "")
+    if len(word) != len(re.findall(ur"[\u4e00-\u9fa5]", word)):
```
gimite: Use code range \u3400-\u9fff instead, which is used here:
https://github.com/google/personfinder/blob/bcec681718ae454ced01d00fe839026b18ab80ad/app/text_query.py#L44
It looks your range misses some characters.

Even \u3400-\u9fff doesn't look like a comprehensive range, but we can take a look later.

Also, can you simplify this to e.g.;

```python
if not re.search(ur'^[\u3400-\u9fff]$', word):
```

app/script_variant.py
```diff
@@ -145,6 +145,53 @@ def romanize_word_by_unidecode(word):
...
+        return None, None
+
+    if word[:2] in CHINESE_FAMILY_NAME_DICTIONARY:
```
gimite: Optional: You may use for-loop to group these two.

app/script_variant.py
```diff
@@ -145,6 +145,53 @@ def romanize_word_by_unidecode(word):
...
+    
+    Return:
+        Boolean
+    """
+    surname, lastname  = split_chinese_name(word)
+    return surname is not None and lastname is not None
+
+def split_chinese_name(word):
+    """
+    This method tries to split a chinese name into two parts: surname and lastname
```
gimite: Use family_name and given_name, which are terminology used in Person Finder.

"lastname" usually means surname in English, so it's confusing.

app/script_variant.py
```diff
@@ -145,6 +145,53 @@ def romanize_word_by_unidecode(word):
...
+def romanize_chinese_person_name(word):
+    """
+    This method romanizes a chinese person name including surname and last name.
+
+    Returns:
+        romanized_cn_name
+    """
+
+    surname, lastname = split_chinese_name(word)
+    if not surname or not lastname:
```
gimite: We want to apply this logic even if the query is just a family name. So avoid checking "not lastname".

app/script_variant.py
```diff
@@ -145,6 +145,53 @@ def romanize_word_by_unidecode(word):
...
+        return word[0], word[1:]
+
+    return None, None
+
+def romanize_chinese_person_name(word):
+    """
+    This method romanizes a chinese person name including surname and last name.
+
+    Returns:
+        romanized_cn_name
```
gimite: Should be "Romanized Chinese name". Avoid using abbreviations e.g., "cn".

app/script_variant.py
```diff
@@ -145,6 +145,53 @@ def romanize_word_by_unidecode(word):
...
+
+    if word[:2] in CHINESE_FAMILY_NAME_DICTIONARY:
+        return word[:2], word[2:]
+
+    elif word[0] in CHINESE_FAMILY_NAME_DICTIONARY:
+        return word[0], word[1:]
+
+    return None, None
+
+def romanize_chinese_person_name(word):
```
gimite: Can you add this method to here as well?:
https://github.com/google/personfinder/blob/bcec681718ae454ced01d00fe839026b18ab80ad/app/full_text_search.py#L372

It is needed to apply this romanization for indexing.

app/script_variant.py
```diff
@@ -145,6 +145,53 @@ def romanize_word_by_unidecode(word):
...
+    """
+    This method romanizes a chinese person name including surname and last name.
+
+    Returns:
+        romanized_cn_name
+    """
+
+    surname, lastname = split_chinese_name(word)
+    if not surname or not lastname:
+        return []
```
gimite: It is a bit weird that this method returns a string or an empty list. Can you return None here instead?

app/script_variant.py
```diff
@@ -145,6 +145,53 @@ def romanize_word_by_unidecode(word):
...
+
+    romanized_surname_list = list(CHINESE_FAMILY_NAME_DICTIONARY[surname])
+    return romanized_surname_list[0].capitalize() + unidecode(lastname).strip()
```
gimite: Maybe you mean upper() instead of capitalize()? I don't think we want to capitalize (e.g., "hoge" -> "Hoge") here.

matthew-z: I noticed that original unidecode will output capitalised results so I added `capitalize` here to avoid potential problems. Perhaps it is not necessary.

gimite: Can you return a list with a single string here to be consistent with other romanization methods?

gimite:
> The author have uploaded a MIT license.

Sounds good! Made some comments to the code.

app/script_variant.py
gimite: Oh I see. Then can you remove capitalize()? The matching should be case insensitive, so case should not matter.

gimite:
Also, it's optional, but is very appreciated if you put some tests here:
https://github.com/google/personfinder/blob/master/tests/test_full_text_search.py

matthew-z:
Thank you for the code review. 

I will fix these problems then.

app/script_variant.py
matthew-z: OK!

app/script_variant.py
matthew-z: OK


BTW, I suppose I don't need to change the search options [returned_fields] (https://github.com/google/personfinder/blob/bcec681718ae454ced01d00fe839026b18ab80ad/app/full_text_search.py#L186) or [get_person_ids_from_results](https://github.com/google/personfinder/blob/bcec681718ae454ced01d00fe839026b18ab80ad/app/full_text_search.py#L139).

right?

app/script_variant.py
matthew-z: OK, Thank you

app/script_variant.py
matthew-z: OK

app/script_variant.py
matthew-z: No problem

app/script_variant.py
matthew-z: OK

app/script_variant.py
matthew-z: OK

app/chinese_family_name_dict.txt
matthew-z: OK

app/script_variant.py
gimite: Oh good point, you should update them as well, I believe.

It would be better to define a list of method names somewhere and use it in all places. You can do it now, or I can do it later.

matthew-z:
Hi

I have made some changes according to the code review.

BTW, I found that [create_romanized_query_txt](https://github.com/google/personfinder/blob/bcec681718ae454ced01d00fe839026b18ab80ad/app/full_text_search.py#L98) and  [make_or_regexp](https://github.com/google/personfinder/blob/bcec681718ae454ced01d00fe839026b18ab80ad/app/full_text_search.py#L32) may cause many problems in handling Chinese name and Japanese name, for example:

A Chinese person: 姚明,  romanize: family name 姚 -> Yao, given name 明-> Ming. 

Usually mainland people and Hong Kong people put their family name on the right hand side when they romanise their name. On passport, usually the english version name would be:   'Ming Yao'. 

Note that Taiwan people usually use a different order: `Family_name given_name` for english version of name.

If we stored his info in this way 
```
        self.p17 = model.Person.create_original_with_record_id(
            'haiti',
            'haiti/0915',
            full_name=u'Ming Yao',
            entry_date=TEST_DATETIME)
```

And we may call the `full_text_search `using his Chinese name: 

`full_text_search.search('haiti', u'姚明', 5)`

then the `and_query` will become `u'("姚明" OR "Yao Ming") AND (repo: haiti)'`. since `"Yao Ming"` is an atomic term, and then the search API will return nothing as the index only has "Ming Yao". Furthermore, `make_or_regexp` also has the same problem: it assumes that the document field and query are always in the same order.

Additionally, even in handling Japanese name, if the document only has a romaji full_name and the query are kanji without space, then the result will also be nothing as the romanisation of query does not try to split the kanji name without space into a family name and a given name.

app/script_variant.py
matthew-z: I suppose `get_person_ids_from_results` is used to handle location/name matching problem right. In this sense, I suppose that we don't have to modify the `get_person_ids_from_results` as long as the query will be romanized by `script_variant.romanize_search_query` in [make_or_regexp](https://github.com/google/personfinder/blob/bcec681718ae454ced01d00fe839026b18ab80ad/app/full_text_search.py#L32). what do you think?

gimite: No, we should also use fields with all romanizations applied in the index in get_person_ids_from_results. Otherwise it doesn't cover the case where the record in the database is in kanji, and the query is in romanized alphabet.

gimite:
> BTW, I found that create_romanized_query_txt and make_or_regexp may cause many problems in handling Chinese name and Japanese name (Google/master branch has the same problem), for example:

That's good point. I agree. I created an issue #289 for it with some potential solutions. It would be great if you also work on it, but please send another pull request for it because it's a separated issue.

app/full_text_search.py
```diff
@@ -370,7 +370,8 @@ def create_document(person):
     # Applies two methods because kanji is used in Chinese and Japanese,
     # and romanizing in chinese and japanese is different.
     romanize_methods = [script_variant.romanize_word_by_unidecode,
-                        script_variant.romanize_japanese_word]
+                        script_variant.romanize_japanese_word,
+                        script_variant.romanize_chinese_name]
```
gimite: Looks like it is expected that romanize_methods here returns a list. Can you update romanize_chinese_name to return a list (with zero or one string) instead of a string?

matthew-z: No problem, I have pushed a new commit to solve this problem

matthew-z: Then I will try to add some test cases in test_full_text_search

app/script_variant.py
matthew-z: Yes, it will cause a problem indeed.

I will fix this.

matthew-z:
Hi,

I have fixed the returned fields issues in search, and added some test cases in test_full_text_search.

Furthermore, I added a global variable to control the applied romanzie methods so that we don't need to write the return fields manually.

app/full_text_search.py
```diff
@@ -114,7 +118,7 @@ def create_romanized_query_txt(query_txt):
     return enclose_in_parenthesis(romanized_query)
 
 
-def get_person_ids_from_results(romanized_query, results_list):
+def get_person_ids_from_results(romanized_query, results_list, returned_fields):
```
gimite: Can you rename returned_fields to e.g., romanized_name_fields? "returned" doesn't make much sense in this method.

app/full_text_search.py
```diff
@@ -130,19 +134,15 @@ def get_person_ids_from_results(romanized_query, results_list):
...
-                if field.name == 'names_romanized_by_romanize_word_by_unidecode':
-                    names = field.value
-                if field.name == 'record_id':
-                    id = field.value
-                if field.name == 'names_romanized_by_romanize_japanese_word':
-                    romanized_jp_names = field.value
-            
-            if id in index_results:
+            fields = {field.name:field.value for field in document.fields}
+            id = fields.get('record_id', None)
```
gimite: I believe record_id always exists, so you can simply write fields['record_id'].

app/full_text_search.py
```diff
@@ -130,19 +134,15 @@ def get_person_ids_from_results(romanized_query, results_list):
...
+            romanized_fields = (value for name, value in fields.items()
```
gimite: Can you rename romanized_fields to romanized_names?

app/full_text_search.py
```diff
@@ -204,7 +208,7 @@ def search(repo, query_txt, max_results):
 
     results_list = [non_romanized_person_location_index_results,
                     person_location_index_results]
-    index_results = get_person_ids_from_results(query_txt, results_list)
+    index_results = get_person_ids_from_results(query_txt, results_list, returned_fields)
```
gimite: Can you pass returned_name_fields here instead of returned_fields? We don't want to include 'record_id' here.

app/full_text_search.py
```diff
@@ -369,9 +373,7 @@ def create_document(person):
 
     # Applies two methods because kanji is used in Chinese and Japanese,
     # and romanizing in chinese and japanese is different.
-    romanize_methods = [script_variant.romanize_word_by_unidecode,
-                        script_variant.romanize_japanese_word,
-                        script_variant.romanize_chinese_name]
+    romanize_methods = ROMANIZE_METHODS
```
gimite: Can you directly use ROMANIZED_METHODS below instead of defining another variable, for simplicity?

tests/test_full_text_search.py
```diff
@@ -347,6 +365,24 @@ def test_search_by_name_only(self):
         assert set([r.record_id for r in results]) == \
             set(['haiti/0523'])
 
+        # Search by Special Chinese Family Name
+        # Where Documents are written in English
```
gimite: Would be better to say "record" instead of "Documents" here. Same for below.

app/full_text_search.py
matthew-z: I think 'record_id' is necessary here as we need to remove duplicates in 'get_person_ids_from_results' according to 'record_id'.

matthew-z: Sorry, I misunderstood it.

We don't need to add 'record_id' here.

matthew-z:
Thanks for reviewing.

I just fixed these problems, could you please have a look?

gimite:
Looks good. Merged. Thanks!

matthew-z:
Thank you!

Issue #189, opened 2015-10-06
Update the message to let users know that they can search by person name + location


gimite:
Update "Enter the person's name, parts of the name, or mobile phone number."

After we launch full text search.

Stephanie1125:
Is update message needed for new UI too?
![search](https://cloud.githubusercontent.com/assets/15961861/18864854/3a1aa980-84d3-11e6-965e-76f78b9e7173.png)

gimite:
Let's leave it now and ask @nmakida to confirm. We may remove the message itself.

Stephanie1125:
Okay! Let's discuss about it this Friday

PR #283, opened 2016-08-30
implement full text search for name/location boxes

Stephanie1125:
So here is some change in this pull request:
+ Edit the full_text_search.py to implement two boxes search(one for name, the other for location)
+ Edit the old structure and added filter for location field. 
+ Edited the search function inside the results.py.

Instead of sending a query_txt (string type), a query_list is sent into the search function in results.py.
query_list is a list contain [query_name, query_location]  and this is used in full_text_search.py because we are sending two query (query_name and query location), this is for distinguish between name and location queries.
if there is no location input, the query_list[1] will be ""(empty string).
and the query_name (query_list[0]) plays the role as the query_txt we have before (non-edit version: only contain the name part) 
I also edit the unit test/test_fulltext_search and server_test/person_note_tests.py and tools/all_tests are passed.

+ Here is the new ui for searching: <br>
search


app/full_text_search.py
```diff
@@ -14,37 +14,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
 import re
-
+import logging
```
kevinalle: sort them alphabetically

app/full_text_search.py
```diff
@@ -114,47 +98,74 @@ def create_romanized_query_txt(query_txt):
     return enclose_in_parenthesis(romanized_query)
 
 
-def get_person_ids_from_results(romanized_query, results_list):
+def is_query_match(query_txt, values, romanized_values):
+    """
+    To check if the query match the field value
```
kevinalle: Checks if the query matches the field value

app/full_text_search.py
```diff
@@ -114,47 +98,74 @@ def create_romanized_query_txt(query_txt):
...
+    Args:
+        query_txt: Search query
+        values: field value for unidecode method
+        romanized_values: romanized field value
+    Returns:
+        Boolean
+    """
+    if not query_txt:
+        return True
```
kevinalle: empty matches everything?

app/full_text_search.py
```diff
@@ -114,47 +98,74 @@ def create_romanized_query_txt(query_txt):
...
+    romanized_query_list = (script_variant.romanize_search_query(query_txt))
+
+    for name in romanized_query_list:
+        words = name.split(" ")
+        for word_index, word in enumerate(words):
+            if not re.search(word, values + " " + romanized_values, re.I):
+                break
+            if word_index == len(words) - 1:
+                return True
+    return False
```
kevinalle: Mm, all this function is weird. Can you explain what it does? (add a comment)

app/full_text_search.py
```diff
@@ -114,47 +98,74 @@ def create_romanized_query_txt(query_txt):
...
+
+
+def get_person_ids_from_results(romanized_query_list, results_list):
     """
     Returns person record_id of persons
```
kevinalle: The explanation is useful. If it's no longer accurate, please update it, but don't remove it

app/full_text_search.py
Stephanie1125: ok!

app/full_text_search.py
Stephanie1125: ok!

app/full_text_search.py
Stephanie1125: yes, empty location query will match everything.
Inside the search function, it check if the name query is empty or not, if it is empty then it will show error message so this is just for empty location query.

app/full_text_search.py
Stephanie1125: This is for creating compiled regular expression for AND search. We want to make sure that every part of the name and location query match the name and location field. Unfortunately, there is no such symbol: "&" in the re.search so we combine the string and see if the query contain part of the field. Here is an example: 
If the user input the location query: tokyo
Inside the field there have: tokyo japan
And this is valid search and also you can get the result.
However, if user input the query: "tokyo taiwan"
This this not valid and will return None because it doesn't match the location field.
screenshot from 2016-08-30 15 02 55

For the old version of full text search, we use OR search because it only has one input box and we don't know which part of the query match the name field. (and in this cause if you input tokyo taiwan as location query it will work because the tokyo part match the location field and this will be a little bit weird and if we use the OR, the user can type random things into the search boxs and still get the result) ---> the goal is to make the location query more accurate to higher the accuracy of finding the right person.

app/full_text_search.py
Stephanie1125: ok!

app/main.py
```diff
@@ -594,4 +594,4 @@ def head(self):
         self.response.clear()
 
 if __name__ == '__main__':
-    webapp.util.run_wsgi_app(webapp.WSGIApplication([('.*', Main)]))
+    webapp.util.run_wsgi_app(webapp.WSGIApplication([('.*', Main)]))
```
kevinalle: add newline at end of file

app/resources/query_form.html.template
```diff
@@ -94,7 +94,7 @@
             {% endif %}
           </span>
         </div>
-      {% if config.use_family_name and not config.family_name_first %}
+       {% if config.use_family_name and not config.family_name_first %}
```
kevinalle: remove extra space

app/results.py
```diff
@@ -83,10 +91,10 @@ def search(self, query_txt):
         if not results:
             if config.get('enable_fulltext_search'):
                 results = full_text_search.search(self.repo,
-                                                  query_txt, MAX_RESULTS)
+                                                  query_dict, MAX_RESULTS)
             else:
                 results = indexing.search(self.repo,
-                                          TextQuery(query_txt), MAX_RESULTS)
+                                          TextQuery(query_txt), MAX_RESULTS)
```
kevinalle: remove space at end of line

tests/server_test_cases/person_note_tests.py
```diff
@@ -4959,7 +4959,7 @@ def test_config_use_alternate_names(self):
         assert '_test_alternate_given' not in d.text
         assert '_test_alternate_family' not in d.text
 
-        self.go('/pakistan/results?query=_test_given+_test_family')
+        self.go('/pakistan/results?role=seek&query_name=_test_given+_test_family&query_location=')
```
kevinalle: add a test with a real query_location

app/full_text_search.py
```diff
@@ -114,47 +99,85 @@ def create_romanized_query_txt(query_txt):
...
+    Returns:
+        Boolean
+    """
+    if not query_txt:
+        return True
+
+    romanized_query_list = (script_variant.romanize_search_query(query_txt))
+
+    for name in romanized_query_list:
+        words = name.split(" ")
```
kevinalle: you don't need to recompute words in every loop. put this line before the for

kevinalle: Name this search_terms

app/full_text_search.py
```diff
@@ -114,47 +99,85 @@ def create_romanized_query_txt(query_txt):
...
+        for word_index, word in enumerate(words):
+            if not re.search(word, values_romanized_unidecode + " " +
+                    values_romanized_jp_words, re.I):
+                break
+            if word_index == len(words) - 1:
+                return True
+    return False
```
kevinalle: This logic is a little hard to follow. at least add comments.
# A query matches a record if all search_terms appear in the record

app/main.py
Stephanie1125: ok!

app/resources/query_form.html.template
Stephanie1125: ok!

app/results.py
Stephanie1125: ok!

app/full_text_search.py
Stephanie1125: ok!

app/full_text_search.py
Stephanie1125: ok!

tests/server_test_cases/person_note_tests.py
Stephanie1125: ok!

app/full_text_search.py
```diff
@@ -99,22 +99,25 @@ def create_romanized_query_txt(query_txt):
     return enclose_in_parenthesis(romanized_query)
 
 
-def is_query_match(query_txt, values_romanized_unidecode,
-                   values_romanized_jp_words):
+def search_terms(query_txt, values_romanized_unidecode,
```
kevinalle: ah, sorry, I meant reaname the "words" variable to search_terms. The name of this function was ok!

app/full_text_search.py
```diff
@@ -99,22 +99,25 @@ def create_romanized_query_txt(query_txt):
...
     """
+    # empty matches everything
     if not query_txt:
         return True
 
     romanized_query_list = (script_variant.romanize_search_query(query_txt))
 
+    # A query matches a record if all search_terms appear in the record
     for name in romanized_query_list:
         words = name.split(" ")
```
kevinalle: rename this variable to search_terms

kevinalle:
Cool!
Just that one comment, and LGTM

app/full_text_search.py
Stephanie1125: Changed it to 
```
    for search_terms in romanized_query_list:
        words = search_terms.split(" ")
```

Stephanie1125:
maybe update new message for #189 ?

PR #285, opened 2016-09-06
Confirmation message when submitting the form

Stephanie1125:
Show confirmation message before submitting the form.

Here is the iphone look of the confirmation message:
iphone look

### DIFF
--- a/app/resources/forms.js
+++ b/app/resources/forms.js
@@ -367,3 +367,13 @@ function make_form_expandable(expandedSection) {
     return false;
   });
 }
+
+function form_confirmation() {
+   var personName = $('given_name').value + ' ' + $('family_name').value;
+   if (confirm('Do you want to create a record for ' + personName + '?\n' +
+               'Press OK to submit the form.\n' +
+               'Press Cancel to return to the form.'))
+      return validate_fields();
+   else
+      return false;
+}
--- a/app/resources/forms.js.template
+++ b/app/resources/forms.js.template
@@ -369,11 +369,10 @@ function make_form_expandable(expandedSection) {
   });
 }
 
-// Check validation of fields before asking comfirmation
-// Ask for confirmation when submitting a form
+// Check validation of fields and ask for confirmation when submitting a form
 // familyNameFirst: true or false
 function confirm_submit(familyNameFirst) {
-  var personName
+  var personName;
 
   if (familyNameFirst)
     personName = $('family_name').value + ' ' + $('given_name').value;
--- a/app/resources/forms.js.template
+++ b/app/resources/forms.js.template
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+{% load i18n %}
 
 function $(id) {
   return document.getElementById(id);
@@ -368,12 +369,18 @@ function make_form_expandable(expandedSection) {
   });
 }
 
-function form_confirmation() {
-   var personName = $('given_name').value + ' ' + $('family_name').value;
-   if (confirm('Do you want to create a record for ' + personName + '?\n' +
-               'Press OK to submit the form.\n' +
-               'Press Cancel to return to the form.'))
-      return validate_fields();
-   else
-      return false;
+// Ask for confirmation when submitting a form
+// familyNameFirst: true or false
+function confirm_submit(familyNameFirst) {
+  var personName = $('given_name').value + ' ' + $('family_name').value;
+
+  if (familyNameFirst)
+    personName = $('family_name').value + ' ' + $('given_name').value;
+
+  if (validate_fields())
+    return (confirm('{% trans "Do you want to create a record for the following person?" %}'
+      + '\n' + personName));
+  else 
+    return false;
 }
+
--- a/app/resources/forms.js.template
+++ b/app/resources/forms.js.template
@@ -369,16 +369,19 @@ function make_form_expandable(expandedSection) {
   });
 }
 
+// Check validation of fields before asking comfirmation
 // Ask for confirmation when submitting a form
 // familyNameFirst: true or false
 function confirm_submit(familyNameFirst) {
-  var personName = $('given_name').value + ' ' + $('family_name').value;
+  var personName
 
   if (familyNameFirst)
     personName = $('family_name').value + ' ' + $('given_name').value;
+  else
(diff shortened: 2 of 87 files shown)


app/resources/forms.js
```diff
@@ -367,3 +367,13 @@ function make_form_expandable(expandedSection) {
     return false;
   });
 }
+
+function form_confirmation() {
+   var personName = $('given_name').value + ' ' + $('family_name').value;
```
gimite: We change the order of given/family name with config.family_name_first per repository. Use it to change the order like this:
https://github.com/google/personfinder/blob/master/app/resources/create.html.template#L55

app/resources/forms.js
```diff
@@ -367,3 +367,13 @@ function make_form_expandable(expandedSection) {
...
+   if (confirm('Do you want to create a record for ' + personName + '?\n' +
+               'Press OK to submit the form.\n' +
+               'Press Cancel to return to the form.'))
+      return validate_fields();
```
gimite: Isn't it better to validate first and then confirm? Otherwise the user would first be asked for confirmation, and then encounters the error.

app/resources/forms.js
```diff
@@ -367,3 +367,13 @@ function make_form_expandable(expandedSection) {
     return false;
   });
 }
+
+function form_confirmation() {
+   var personName = $('given_name').value + ' ' + $('family_name').value;
+   if (confirm('Do you want to create a record for ' + personName + '?\n' +
```
gimite: Internationalize this message i.e., make the message ready to be translated into other languages.

You can do that by writing e.g.,
```js
  confirm('{% trans "Do you want to create a record for this person?:" %}');
```
It's actually tricky to translate a message with a placeholder like personName here in JavaScript. Here, I would recommend *escape* from the issue by breaking the line into two:

```
Do you want to create a record for this person?
  Hiroshi Ichikawa
```

app/resources/forms.js
```diff
@@ -367,3 +367,13 @@ function make_form_expandable(expandedSection) {
...
+               'Press OK to submit the form.\n' +
```
gimite: I believe these lines are obvious and make the message too verbose. What do you think?

app/resources/forms.js
```diff
@@ -367,3 +367,13 @@ function make_form_expandable(expandedSection) {
     return false;
   });
 }
+
+function form_confirmation() {
```
gimite: It's usually preferred to use a verb for function names. e.g., confirm_submit, ask_for_form_confirmation

Also, add a short comment to explain what this function does.

Stephanie1125: ok!

app/resources/forms.js
Stephanie1125: i removed some lines

app/resources/forms.js
Stephanie1125: ok!

app/resources/forms.js
Stephanie1125: ok!

app/resources/forms.js
Stephanie1125: ok! thank you for the explanation !

app/resources/forms.js.template
```diff
@@ -368,12 +369,18 @@ function make_form_expandable(expandedSection) {
...
-               'Press Cancel to return to the form.'))
-      return validate_fields();
-   else
-      return false;
+// Ask for confirmation when submitting a form
+// familyNameFirst: true or false
+function confirm_submit(familyNameFirst) {
+  var personName = $('given_name').value + ' ' + $('family_name').value;
+
+  if (familyNameFirst)
```
gimite: It would be easier to read if you use:
```js
var personName = familyNameFirst ? ... : ...;
```
or
```js
var personName;
if (familyNameFirst)
  ...
else
  ...
```

app/resources/forms.js.template
```diff
@@ -368,12 +369,18 @@ function make_form_expandable(expandedSection) {
...
 
-function form_confirmation() {
-   var personName = $('given_name').value + ' ' + $('family_name').value;
-   if (confirm('Do you want to create a record for ' + personName + '?\n' +
-               'Press OK to submit the form.\n' +
-               'Press Cancel to return to the form.'))
-      return validate_fields();
-   else
-      return false;
+// Ask for confirmation when submitting a form
```
gimite: Can you document that it also performs validation of fields?

app/resources/forms.js.template
```diff
@@ -368,12 +369,18 @@ function make_form_expandable(expandedSection) {
...
+// familyNameFirst: true or false
+function confirm_submit(familyNameFirst) {
+  var personName = $('given_name').value + ' ' + $('family_name').value;
+
+  if (familyNameFirst)
+    personName = $('family_name').value + ' ' + $('given_name').value;
+
+  if (validate_fields())
+    return (confirm('{% trans "Do you want to create a record for the following person?" %}'
```
gimite: I should have pointed this earlier, but I feel that it's better if the message looks more like a confirmation e.g., "Are you sure to submit a record for the following person?"

What do you think?

Current message may be interpreted to create something new, independent of the record the user has just input.

app/resources/forms.js.template
Stephanie1125: ok!

app/resources/forms.js.template
Stephanie1125: Yes,
maybe using the word "create" is not very good.

Changed it :)

app/resources/forms.js.template
Stephanie1125: ok!

app/resources/forms.js.template
```diff
@@ -369,16 +369,19 @@ function make_form_expandable(expandedSection) {
   });
 }
 
+// Check validation of fields before asking comfirmation
```
gimite: Optional: It's a bit redundant to say "ask confirmation" twice here. Maybe you can combine the two sentences by "Check validation of fields and ask for ..."

app/resources/forms.js.template
```diff
@@ -369,16 +369,19 @@ function make_form_expandable(expandedSection) {
...
 // Ask for confirmation when submitting a form
 // familyNameFirst: true or false
 function confirm_submit(familyNameFirst) {
-  var personName = $('given_name').value + ' ' + $('family_name').value;
+  var personName
```
gimite: Missing semi-colon.

gimite:
LGTM. Merged.

PR #287, opened 2016-09-08
Making changes for ifconfig/easy_install

nomadmtb:
The original Dockerfile would not build correctly on the latest version of Docker.

In the phusion/baseimage ifconfig is not available for use.  However the ip command is available so I migrated the IP addressing command to use ip over ifconfig inside of the docker/gae-run-app.sh and docker/setup_datastore.sh scripts.

In addition, there was a build error when trying to run easy_install to get the pytest module.  I corrected the command and the image builds successfully.


gimite:
Thanks! Merged.

nomadmtb:
Thank you!

PR #290, opened 2016-09-15
Add title bar and side menu

Stephanie1125:
Add title bar and side menu

+ haiti repo page

+ fixed some form width to make it the same

+ side menu look (click on the gray part to close the side menu)

+ home page (will think about new design for this page)


app/resources/forms.js.template
```diff
@@ -375,4 +375,14 @@ function make_form_expandable(expandedSection) {
   });
 }
 
+// Show the side menu
+function menuOpenNav() {
+  document.getElementById("mySidenav").style.width = "250px";
+  document.getElementById("overLay").style.width = "100%";
```
gimite: It would be more natural to change visibility of this element than changing width.

app/resources/forms.js.template
```diff
@@ -375,4 +375,14 @@ function make_form_expandable(expandedSection) {
...
+}
 
+// Close the side menu
+function menuCloseNav() {
+  document.getElementById("mySidenav").style.width = "0";
+  document.getElementById("overLay").style.width = "0";
+}
```
gimite: Add a line break in the end.

app/resources/forms.js.template
```diff
@@ -375,4 +375,14 @@ function make_form_expandable(expandedSection) {
   });
 }
 
+// Show the side menu
+function menuOpenNav() {
```
gimite: Generally the function name should be a verb. e.g., showSideMenu() would be a better name. Same for below.

Stephanie1125:
+ Here is the link of the demo:
https://pf-liu.appspot.com/global/home.html
+ The admin page look fine as following:
screenshot from 2016-09-16 10 04 41

app/resources/base.html.template
```diff
@@ -136,27 +133,34 @@
...
-        {% if env.ui != "light" %}
-          <a href="https://github.com/google/personfinder"
-              {{env.target_attr|safe}}>{% trans "Developers" %}</a>
-          <span class="link-separator">&#xb7;</span>
-        {% endif %}
-        <a href="{{env.global_url}}/tos" {{env.target_attr|safe}}
-            >{% trans "Terms of Service" %}</a>
-      {% endblock footer %}
-    </div>
+      <div id="mySidenav" class="sidenav">
```
Stephanie1125: ok!

app/resources/base.html.template
```diff
@@ -136,27 +133,34 @@
...
+        <span class="google-logo"><img src="https://material.google.com/static/images/nav_google_logo.svg" alt=""/></span>
```
Stephanie1125: I see!

app/resources/base.html.template
```diff
@@ -136,27 +133,34 @@
...
+          <a href="{{env.global_url}}/tos" {{env.target_attr|safe}}
+              >{% trans "Terms of Service" %}</a>
+          {% if env.show_language_menu %}
+            {% include "language-menu.html.template" %}
+          {% endif %}
+        {% endblock footer %}
+      </div>
+    </div>
     {% include "analytics.html.template" %}
+    <div id="overLay" class="overlay" onclick="menuCloseNav()"></div>
```
Stephanie1125: ok!

app/resources/css-default.template
```diff
@@ -60,8 +61,8 @@ h3 {
 }
 
 .header {
-  border-bottom: 1px solid #dadada;
-  padding: 1em 0;
+  padding: 0;
```
Stephanie1125: i see!

app/resources/css-default.template
```diff
@@ -139,12 +140,46 @@ pre {  /* code samples */
   margin-right: 10px;
 }
 
+/* Title bar */
+
+.title-bar {
+  background-color: #FB8C00;
```
Stephanie1125: i see!

app/resources/css-default.template
```diff
@@ -139,12 +140,46 @@ pre {  /* code samples */
...
+  display: block;
+  color: white;
+}
+
+.title-bar button {
+  background-color: transparent;
+  width: 64px;
+  height: 64px;
+  border: none;
+  padding: 0;
```
Stephanie1125: ok! this is set as default!

app/resources/css-default.template
```diff
@@ -139,12 +140,46 @@ pre {  /* code samples */
...
 /* Language menu */
 
 div.languages {
-  float: {{end}};
-  margin: 10px 0;
-  padding: 0;
+  padding: 15px 8px;
+  text-decoration: none;
+  color: black;
+  display: block;
```
Stephanie1125: I see!

app/resources/css-default.template
```diff
@@ -1136,3 +1177,41 @@ table.api-keys td.empty {
     padding: 5px;
   }
 }
+
+/* Side menu */
+
+.sidenav {
+  height: 100%;
+  width: 0;
+  position: fixed;
```
Stephanie1125: i see!

app/resources/css-default.template
```diff
@@ -1136,3 +1177,41 @@ table.api-keys td.empty {
...
+  border-top: 1px solid #dadada;
+}
+
+.google-logo img {
+  padding: 12px 8px;
+  height: 32px;
+}
+
+.overlay {
+  position:fixed;
```
Stephanie1125: ok!

app/resources/css-default.template
```diff
@@ -1136,3 +1177,41 @@ table.api-keys td.empty {
...
+  top:0;
+  left:0;
+  height: 100%;
+  width: 0;
+  z-index: 2;
+  background: rgba(0,0,0,0.5);
+  overflow-x: hidden;
```
Stephanie1125: ok!

app/resources/css-default.template
```diff
@@ -1136,3 +1177,41 @@ table.api-keys td.empty {
...
+}
```
Stephanie1125: ok!

app/resources/app-base.html.template
```diff
@@ -13,7 +13,8 @@
 {% block title %}{{block.super}}{% endblock title %}
 
 {% block logo %}
-  <a href="{{env.global_url}}"><img src="{{env.global_url}}/logo.png" alt=""></a>
+  <button class="menu" onclick="menuOpenNav()"><img src="{{env.global_url}}/menu.png" alt="Menu"></button>
```
Stephanie1125: ok!

app/resources/forms.js.template
Stephanie1125: i see!

app/resources/forms.js.template
Stephanie1125: ok!

Stephanie1125:
Here is the admin page:
admin2
admin3

app/resources/css-default.template
```diff
@@ -1206,12 +1194,12 @@ table.api-keys td.empty {
...
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
   z-index: 2;
   background: rgba(0,0,0,0.5);
-  overflow-x: hidden;
+  display: none;
 }
```
Stephanie1125: Already added a new line.

gimite: Looks like you are still missing the line break. The red circle on Github indicates that the line misses a line break.

app/resources/css-default.template
gimite: Looks like this is still "fixed".

app/resources/forms.js.template
```diff
@@ -376,13 +376,13 @@ function make_form_expandable(expandedSection) {
...
+  document.getElementById("sidenav").style.maxWidth = "250px";
 }
 
 // Close the side menu
-function menuCloseNav() {
-  document.getElementById("mySidenav").style.width = "0";
-  document.getElementById("overLay").style.width = "0";
-}
\ No newline at end of file
+function closeSideMenu() {
```
gimite: Nitpick, but it's better to use paired names for this kind of paired functions. Here you should use either "show/hide" or "open/close".

app/resources/forms.js.template
```diff
@@ -376,13 +376,13 @@ function make_form_expandable(expandedSection) {
 }
 
 // Show the side menu
-function menuOpenNav() {
-  document.getElementById("mySidenav").style.width = "250px";
-  document.getElementById("overLay").style.width = "100%";
+function showSideMenu() {
+  document.getElementById("overlay").style.display = "block";
+  document.getElementById("sidenav").style.maxWidth = "250px";
```
gimite: Can you add a comment that you change maxWidth instead of display here for animation?

app/resources/admin-base.html.template
```diff
@@ -18,6 +18,10 @@
   admin
 {% endblock %}
 
+{% block logo %}
+    <button class="title-bar-button" id="menu-button" onclick="showSideMenu()"><img src="{{env.global_url}}/menu.png" alt="Menu"></button>Google Person Finder
```
gimite: This line should have 2-space indent instead of 4.

app/resources/app-base.html.template
```diff
@@ -13,8 +13,15 @@
 {% block title %}{{block.super}}{% endblock title %}
 
 {% block logo %}
-  <button class="menu" onclick="menuOpenNav()"><img src="{{env.global_url}}/menu.png" alt="Menu"></button>
-  <a href="{{env.global_url}}">Google Person Finder</a>
+  {% if env.repo and env.action and env.action != "start" %}
+    {% if env.rtl %}
+	  <button class="title-bar-button"><a href="{{env.start_url}}"><img src="{{env.global_url}}/forward_arrow.png" alt=""></a></button>Google Person Finder
```
gimite: Unindent this line by 2 spaces.

gimite: "Google Person Finder" part is common among all cases. Can you put it outside of these if-statements so that you write it just once?

app/resources/admin-base.html.template
```diff
@@ -33,57 +37,43 @@
...
+  </form>
+
+  {% if env.repo %}
+    <a href="{{env.repo_url}}/admin">Repository settings</a>
+    <a href="{{env.repo_url}}/admin/dashboard">Dashboard</a>
+    <a href="{{env.repo_url}}/admin/review">Review notes</a>
+    <a href="{{env.repo_url}}/admin/api_keys/list">Repository API keys</a>
+    <a href="{{env.repo_url}}/admin/delete_record">Delete a record</a>
+  {% endif %}
+{% endblock footer %}
```
gimite: Add a line break in the end of the line.

app/resources/base.html.template
```diff
@@ -75,92 +75,85 @@
...
-
     </div>
 
+    {% if env.repo %}
+      <div class="subtitle-bar">
+        {{env.repo_title}}
+      </div>
+    {% endif %}
 
     {% if env.repo_test_mode %}
```
gimite: I believe this should be part of "window" div.

app/resources/base.html.template
```diff
@@ -75,92 +75,85 @@
...
         {% trans "This repository is currently in test mode. While test mode is in effect, records that are over 24 hours old are deleted." %}
       </div>
     {% endif %}
 
-    {% block container %}
-      <div class="container">
-        <div class="content" role="main">
-          {% block content %}
-          {% endblock content %}
+    <div class=window>
```
gimite: Enclose the class name with double-quotes.

Also, "window" may be confusing because usually the whole page is called "window" in HTML. I don't have a good naming idea, but maybe something like "outer-container", "content-pane". (We actually want to say "everything except for the header and the side bar", but that would be too long.)

app/resources/base.html.template
```diff
@@ -75,92 +75,85 @@
...
 
-        {% block footer %}
-          {% if env.repo and env.ui == "light" %}
-            <a href="{{env.default_ui_url}}">{% trans "Desktop version" %}</a>
-            <span class="link-separator"></span>
-          {% endif %}
-          <a href="{{env.global_url}}" {{env.target_attr|safe}}
-              >{% trans "About Google Person Finder" %}</a>
+    <div id="sidenav" class="sidenav">
+      {% block footer %}
```
gimite: Optional: Now it's weird to call this "footer". It should rather be "sidenav". You can rename it now, or maybe just add a comment "TODO Rename this to sidenav" and do it later.

app/resources/css-default.template
```diff
@@ -16,18 +16,17 @@
 
 /* All pages */
 
-html {
-  padding: 15px;
+.window {
```
gimite: Can you move this below "body"? Roughly speaking, outer element or above element should come first.

app/resources/css-default.template
```diff
@@ -122,6 +116,7 @@ pre {  /* code samples */
 
 .button-blue input {
   background: #4285f4;
+  margin-top: 10px;
```
gimite: Why did you add margin here? Maybe you should also add it to .button-red as well? Why not adding margin to the bottom as well?

app/resources/css-default.template
```diff
@@ -143,43 +138,40 @@ pre {  /* code samples */
...
 }
 
 .title-bar button {
   background-color: transparent;
-  width: 64px;
   height: 64px;
   border: none;
-  padding: 0;
-  float: left;
+  width: 64px;
```
gimite: Nitpick, but "width" and "height" should be written together.

app/resources/css-default.template
```diff
@@ -143,43 +138,40 @@ pre {  /* code samples */
...
-button .menu img {
+button .title-bar-button img {
   height: 24px;
   width: 24px;
 }
 
+div.subtitle-bar {
+  font-size: 20px;
+  font-weight: normal;
+  margin-top: 0;
```
gimite: Shouldn't this be 0 by default?

app/resources/css-default.template
```diff
@@ -143,43 +138,40 @@ pre {  /* code samples */
...
+  box-shadow: 0 2px 1px -1px #a0a0a0;
+  background-color: #ffffff;
+  display: block;
```
gimite: Div should be display: block by default.

app/resources/app-base.html.template
gimite: Fill in the "alt" attribute. Maybe just say "Back".

app/resources/view.html.template
```diff
@@ -17,6 +17,16 @@
   {{person.full_name}} - {{block.super}}
 {% endblock %}
 
+{% if params.query or params.given_name or params.family_name %}
+  {% block logo %}
+    {% if env.rtl %}
+      <button class="title-bar-button"><a href="{{results_url}}"><img src="{{env.global_url}}/forward_arrow.png" alt=""></a></button>Google Person Finder
```
gimite: Fill in the "alt" attribute. Maybe just say "Back".

app/results.py
```diff
@@ -100,7 +100,8 @@ def search(self, query_dict):
             result.view_url = self.get_url('/view',
                                            id=result.record_id,
                                            role=self.params.role,
-                                           query=self.params.query,
+                                           query_name=self.params.query_name,
+                                           query_location=self.params.query_location,
```
gimite: Over 80 chars. You can either break the line after "=":
```python
  query_location=
      self.params.query_location,
```
Or break after ```get_url(```:
```python
result.view_url = self.get_url(
    '/view',
    ...
```

app/results.py
```diff
@@ -218,7 +221,7 @@ def get(self):
             else:
                 # Look for prefix matches.
                 results = self.search(query_dict)
-                results_url = self.get_results_url(self.params.query)
+                results_url = self.get_results_url(self.params.query_name, self.params.query_location)
```
gimite: Over 80 chars.

Hint: If you use Sublime, you can show a vertical bar at 80th column by following:
http://stackoverflow.com/a/25901060/4339963
(You probably don't want to auto-wrap, but just show the ruler.)

app/resources/forms.js.template
Stephanie1125: ok!

app/resources/forms.js.template
Stephanie1125: ok! changed it to hideSideMenu

app/resources/admin-base.html.template
Stephanie1125: ok

app/resources/app-base.html.template
Stephanie1125: ok

Stephanie1125: ok

app/resources/admin-base.html.template
Stephanie1125: ok

app/resources/base.html.template
Stephanie1125: ok~

app/resources/base.html.template
Stephanie1125: i see!

app/resources/css-default.template
Stephanie1125: ok!

app/resources/css-default.template
Stephanie1125: ok!

app/resources/css-default.template
Stephanie1125: ok

app/resources/app-base.html.template
Stephanie1125: ok

app/resources/view.html.template
Stephanie1125: ok

app/results.py
Stephanie1125: ok

app/results.py
Stephanie1125: ok

app/resources/css-default.template
Stephanie1125: ok

Stephanie1125:
Also changed the color for result page
if clicked before--> change color to dark blue.

change color

app/resources/forms.js.template
```diff
@@ -376,13 +376,15 @@ function make_form_expandable(expandedSection) {
 }
 
 // Show the side menu
+// Change the sidenav maxWidth for animation
```
gimite: This comment should be just before the line to change maxWidth. The comment here should just contain general comment for the entire function.

app/resources/app-base.html.template
```diff
@@ -15,13 +15,14 @@
 {% block logo %}
   {% if env.repo and env.action and env.action != "start" %}
     {% if env.rtl %}
-	  <button class="title-bar-button"><a href="{{env.start_url}}"><img src="{{env.global_url}}/forward_arrow.png" alt=""></a></button>Google Person Finder
+      <a href="{{env.start_url}}"><img class="title-bar-button" src="{{env.global_url}}/forward_arrow.png" alt="Forward"></a>
```
gimite: Alt attribute for this should be still "Back" because it's a button to go back (even though it's right arrow).

app/resources/css-default.template
```diff
@@ -141,27 +142,22 @@ pre {  /* code samples */
...
   color: white;
   background-color: #616161;
-}
-
-.title-bar button {
-  background-color: transparent;
   height: 64px;
-  border: none;
-  width: 64px;
+  line-height: 64px; /* vertical center text */
```
gimite: Maybe you mean "Vertically center the text"?

app/resources/css-default.template
```diff
@@ -632,6 +628,7 @@ img.icon {
 
 .view input.add-note {
   width: 100%;
+  margin-top: 10px;
```
gimite: Can you add a comment to explain the reason of this margin?

app/resources/view.html.template
```diff
@@ -17,15 +17,22 @@
   {{person.full_name}} - {{block.super}}
 {% endblock %}
 
-{% if params.query or params.given_name or params.family_name %}
-  {% block logo %}
+{% block logo %}
+  {% if params.query_name or params.given_name or params.family_name %}
     {% if env.rtl %}
-      <button class="title-bar-button"><a href="{{results_url}}"><img src="{{env.global_url}}/forward_arrow.png" alt=""></a></button>Google Person Finder
+      <a href="{{results_url}}"><img class="title-bar-button" src="{{env.global_url}}/forward_arrow.png" alt="Forward"></a>
```
gimite: Same as above for "alt" attribute.

app/resources/view.html.template
```diff
@@ -17,15 +17,22 @@
   {{person.full_name}} - {{block.super}}
 {% endblock %}
 
-{% if params.query or params.given_name or params.family_name %}
-  {% block logo %}
+{% block logo %}
+  {% if params.query_name or params.given_name or params.family_name %}
```
gimite: Do you have any reason to swap the order of "block logo" and this if-statement? The original order looks better because you don't need to put else-statement to repeat the original implementation here.

Or, you can keep the current order, but replace the content of the else-statement with {{block.super}}, which means that it uses the implementation written in the base template.

app/resources/view.html.template
Stephanie1125: ok

app/resources/app-base.html.template
Stephanie1125: ok

app/resources/css-default.template
Stephanie1125: yes

app/resources/forms.js.template
Stephanie1125: ok

app/resources/css-default.template
Stephanie1125: ok

app/resources/view.html.template
Stephanie1125: yes,
I found there will have some problems for the original order of the code when user want to create a new record for a person who already has some similar search records. if user create a new record for this person and click go back arrow, it will return to the search result error page instead of the front page. and for the current order, it will return to the front page.

edited else part to {{block.super}}

gimite: That sounds weird. Do you have any idea why? If you use {{block.super}}, shouldn't it be equivalent to the logic in original order?

If you have no idea then that's fine because it's anyway fixed, but I'm just curious.

Stephanie1125:
Screenshots for Arabic:
arabic
arabic2

app/resources/view.html.template
Stephanie1125: I think the reason is that "blocks are included regardless of any conditionals around them"
I found the answer from this page:
http://stackoverflow.com/questions/12088222/calling-block-inside-an-if-condition-django-template

app/resources/css-default.template
```diff
@@ -143,7 +143,7 @@ pre {  /* code samples */
   color: white;
   background-color: #616161;
   height: 64px;
-  line-height: 64px; /* vertical center text */
+  line-height: 64px; /* Vertically center the text */
```
gimite: Add one more space before "/*". I'm not sure if there is CSS style guide, but let's follow the general styling in Google to put two whitespaces before inline comment.

app/resources/css-default.template
```diff
@@ -628,7 +628,7 @@ img.icon {
 
 .view input.add-note {
   width: 100%;
-  margin-top: 10px;
+  margin-top: 10px; /* Vertically align blue buttom with record card */
```
gimite: Should be: "To vertically align the "Add note" button with the record card"

For spell correction and more clarification.

app/resources/view.html.template
```diff
@@ -20,16 +20,12 @@
...
     {% endif %}
   {% else %}
-    {% if env.rtl %}
-      <a href="{{env.start_url}}"><img class="title-bar-button" src="{{env.global_url}}/forward_arrow.png" alt="Forward"></a>
-    {% else %}
-      <a href="{{env.start_url}}"><img class="title-bar-button" src="{{env.global_url}}/back_arrow.png" alt="Back"></a>
-    {% endif %}
+    {{block.super}}
   {% endif %}
   Google Person Finder
```
gimite: Use {%trans%} tag to internationalize this. Same for another instance in app-base.

app/resources/view.html.template
gimite: Oh I see. That's tricky. Understood.

app/resources/view.html.template
Stephanie1125: ok

app/resources/css-default.template
Stephanie1125: i see

app/resources/css-default.template
Stephanie1125: ok

Stephanie1125:
Arabic screenshot:
ar3

https://screenshot.googleplex.com/qFGBt65v2X7
this is the google virtual keyboard element and it appear only in Arabic
ar4

It is defined in main.py 308 line
```
env.virtual_keyboard_layout = const.VIRTUAL_KEYBOARD_LAYOUTS.get(env.lang)
```

gimite:
> this is the google virtual keyboard element and it appear only in Arabic

Oh I see. Thanks for the finding. One of the image looks broken, though:
https://screenshot.googleplex.com/Edaggt0tP2F
Can you file an issue for it?

app/resources/css-default.template
```diff
@@ -628,7 +628,7 @@ img.icon {
 
 .view input.add-note {
   width: 100%;
-  margin-top: 10px; /* Vertically align blue buttom with record card */
+  margin-top: 10px;  /* To vertically align the "Add note" button with the record card */
```
gimite: Over 80 chars.

Stephanie1125: I checked the other css code and i found that there is one also over 80 chars
css

Should I fix it too?

gimite: Yeah it would be nice if you fix them too. For the case above, it would be more readable to break line with all commas:
```css
table.api-keys .contact,
table.api-keys .contact a,
table.api-keys .contact a:visited {
...
```

It's usually more readable to break with either all commas or no commas, than breaking with some of them.

Stephanie1125: okay~~ i see!

Stephanie1125:
Home page screenshot:

homepage
homepage2

gimite:
LGTM. Merged.

Issue #197, opened 2015-10-27
Layout of the repository home page looks a bit broken on RTL language + mobile


gimite:
https://screenshot.googleplex.com/h5rZ7kRi90s

Stephanie1125:
I think this is fixed in new UI

gimite:
Yeah thanks. Can you put the pull request ID (prefixed with "#") which fixed the bug here, and mark this as closed?

PR #292, opened 2016-09-23
Fixed Arabic align problem

Stephanie1125:
The two buttons are not align for Arabic. It looks a little bit weird with different screen size.
ar5


gimite:
Thanks for the fix! LGTM. Merged.

Stephanie1125:
fixed #197

Issue #256, opened 2016-05-09
HTTPException: Deadline exceeded in jp_mobile_carriers.py


gimite:
This exception happens occasionally:

HTTPException: Deadline exceeded while waiting for HTTP response from URL: http://dengon.docomo.ne.jp/inoticelist.cgi

For URL like this: http://google.org/personfinder/japan/results?role=seek&query=090xxxxxxx

We should handle the exception.

d268:
I'm looking into this bug. I think we can just add some try-catch around `urllib2.urlopen` inside the `look_up_number` function in `jp_mobile_carriers.py` ([here](https://github.com/google/personfinder/blob/master/app/jp_mobile_carriers.py#L150) and [here](https://github.com/google/personfinder/blob/master/app/jp_mobile_carriers.py#L157)).
My question is: Should we wrap the exception in our own customized exception and `raise`? Or should just we log about it and return? Current [comments](https://github.com/google/personfinder/blob/master/app/jp_mobile_carriers.py#L146) on the function say we should throw exception when failed to scrape, though. 
Shall we decide on how to handle the exception?

gimite:
We can either just throw the exception as is (as we currently do) or raise a customized exception. In either case, result.py should catch the exception and show a nicer error message, saying that we encountered an error connecting to our backend.

PR #298, opened 2016-10-03
Handle exceptions happened when waiting for HTTP response in jp_mobile_carriers.py

d268:
Handle exceptions happened when waiting/reading HTTP response from URL http://dengon.docomo.ne.jp/inoticelist.cgi inside jp_mobile_carriers.py 

See also #256 for more details.

### DIFF
--- a/app/jp_mobile_carriers.py
+++ b/app/jp_mobile_carriers.py
@@ -147,24 +147,31 @@ def look_up_number(number):
         Exception when failed to scrape.
     """
     # Scrape Docomo's gateway page and get a hidden time stamp param.
-    scrape = urllib2.urlopen(DOCOMO_URL).read()
-    hidden_param = DOCOMO_HIDDEN_RE.findall(scrape)[0]
-
-    # Encode the number and the above param as POST data
-    data = get_docomo_post_data(number, hidden_param)
-    encoded_data = urllib.urlencode(data)
-    # Scrape Docomo's answer on the number
-    scrape = urllib2.urlopen(DOCOMO_URL, encoded_data).read()
-
-    # Extract a further redirect url, if any.
-    url = extract_redirect_url(scrape)
-    if url:
-        return url
-    elif docomo_has_messages(scrape):
-        # Checks if Docomo has messages for the number, and returns the url
-        # for Docomo if it does.
-        return DOCOMO_URL + '?' + encoded_data
-
+    try:
+        scrape = urllib2.urlopen(DOCOMO_URL).read()
+    except (urllib2.URLError, httplib.HTTPException, socket.error) as e:
+        raise e
+    else:
+        hidden_param = DOCOMO_HIDDEN_RE.findall(scrape)[0]
+    
+        # Encode the number and the above param as POST data
+        data = get_docomo_post_data(number, hidden_param)
+        encoded_data = urllib.urlencode(data)
+        # Scrape Docomo's answer on the number
+        try:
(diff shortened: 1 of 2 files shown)


app/results.py
```diff
@@ -188,8 +188,12 @@ def get(self):
             # If a query looks like a phone number, show the user a result
             # of looking up the number in the carriers-provided BBS system.
             if self.config.jp_mobile_carrier_redirect:
-                if jp_mobile_carriers.handle_phone_number(self, query.query):
-                    return 
+                try:
+                    if jp_mobile_carriers.handle_phone_number(self, query.query):
+                        return
+                except Exception as e:
+                    logging.error('failed to scrape phone number')
```
gimite: Can you also show the error message to the user too? You can use self.error like this:
https://github.com/google/personfinder/blob/master/app/view.py#L175

Also, "scrape phone number" is confusing because it sounds like we collect phone numbers from some page. You should say e.g., "scrape the search result for the phone number".

app/jp_mobile_carriers.py
```diff
@@ -147,24 +147,31 @@ def look_up_number(number):
...
-        return url
-    elif docomo_has_messages(scrape):
-        # Checks if Docomo has messages for the number, and returns the url
-        # for Docomo if it does.
-        return DOCOMO_URL + '?' + encoded_data
-
+    try:
+        scrape = urllib2.urlopen(DOCOMO_URL).read()
+    except (urllib2.URLError, httplib.HTTPException, socket.error) as e:
+        raise e
```
gimite: These "except" clauses just throw the original exception, so I believe it's equivalent to the original code. Do you have any reason to add these clauses?

app/results.py
d268: Thanks for the pointer! I have further questions:
- Do we need a message with i18 translation support in this case as we show the message to users? (I guess that _("a message") means in the example). If not, then I'll just update with some clearer message (e.g., "failed to scrape the search result for the phone number").
- Should the return code be 500 (internal server error) in this case?

(Sorry for asking too many little details. I'm new to the codebase so please bear with me for now.)

gimite: > Do we need a message with i18 translation support in this case as we show the message to users? (I guess that _("a message") means in the example).

I18n the message shown to the end users using _("..."), but do not i18n the message for logging, which makes it hard to grep the log, for example.

> Should the return code be 500 (internal server error) in this case?

Yes.

> (Sorry for asking too many little details. I'm new to the codebase so please bear with me for now.)

No problem at all.

d268:
I've made code change to address previous comments. PTAL.

app/results.py
```diff
@@ -188,8 +188,13 @@ def get(self):
             # If a query looks like a phone number, show the user a result
             # of looking up the number in the carriers-provided BBS system.
             if self.config.jp_mobile_carrier_redirect:
-                if jp_mobile_carriers.handle_phone_number(self, query.query):
-                    return 
+                try:
+                    if jp_mobile_carriers.handle_phone_number(self, query.query):
+                        return
+                except Exception as e:
+                    logging.error('failed to scrape search result for the phone number.')
```
gimite: Over 80 chars.

FYI We are following the external version of Python Style Guide:
https://google.github.io/styleguide/pyguide.html
which is a bit different from the internal one.

gimite: Can you also log the stack trace of the exception to help debugging the reason?

Issue #206, opened 2015-11-09
Make it clear that location fields may be missing in the records, when a user search with a location


gimite:
When a user searches for [hiroshi tokyo] and gets no results, we should tell the user that it doesn't always mean that there are no records with name "hiroshi" in Tokyo. All location fields are optional, so it's possible that the record doesn't have a city name.

PR #299, opened 2016-10-04
If no search results for name+location show search results without location

Stephanie1125:
Fixed issue #206 
+ No search results for name+location but have search results for name
match

+ No search results for both name+location and name (Only show no search results for name+location)
no

+ Has match for name+location (just normally show the result)
normal


app/results.py
```diff
@@ -224,7 +225,21 @@ def get(self):
...
                 results_url = self.get_results_url(self.params.query_name,
                                                    self.params.query_location)
+                results_url_without_location = self.get_results_url(
+                                                    self.params.query_name, '')
                 third_party_query_type = ''
+                query_location = self.params.query_location
+
+            # If there is no results match for both name and location
+            # Check if there have results match for name
+            if not results:
```
kevinalle: All of this should be inside the above else.

app/resources/css-default.template
```diff
@@ -890,12 +896,24 @@ resultsList {
...
 .searchResults {
   list-style: none;
   margin: 0;
   padding: 0;
 }
 
+/* emphasized the messages of 
+  no search results for name + location 
+  but have search results for name */
+.bold-message {
```
Stephanie1125: ok~

Stephanie1125:
Thank you~ Merged :)

PR #300, opened 2016-10-06
Fixed some bugs and edit the query form (seek)

Stephanie1125:
Fixed issue #189
+ Remove the message and change "Name to Name/phone number" 
 (Edit the Text field for this because the "Name/phone number" is too long. )
search

+ **[fix bug]** Fixed the ios button
It was:
ios1
Now is:
ios2

+ **[fix bug]** Fixed the height of the text field changes by 1px when it's focused
+ **[fix bug]** Make sure that the (No results for "name + location" , but have results for "name") message only shown in seek form.

### DIFF
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -112,6 +112,7 @@ pre {  /* code samples */
   padding: 5px 15px;
   border-radius: 2px;
   box-shadow: 0 2px 5px 0 rgba(0, 0, 0, 0.26);
+  -webkit-appearance: none;  /* for ios button */
 }
 
 .button input:disabled {
@@ -200,7 +201,7 @@ input {
   box-sizing: border-box;
   font-family: arial, sans-serif;
   font-size: 13px;
-  margin-bottom: .5em;
+  margin-bottom: 8px;
   max-width: 100%;
 }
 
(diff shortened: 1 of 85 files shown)


app/resources/css-default.template
```diff
@@ -605,6 +607,16 @@ td.submit {
   border: 2px solid #e6e6e6;
 }
 
+.query.seek .field {
```
Stephanie1125: Should I add a comment for this? 
/* Edit field in seek query because label is too long (change line) */ ?

app/resources/css-default.template
```diff
@@ -605,6 +607,16 @@ td.submit {
...
+  display: block;
+  margin: 8px 20px;
+}
+
+.query.seek .label {
```
Stephanie1125: Should I add a comment for this too? 
/* Edit label in seek query because label is too long */ ?

app/resources/css-default.template
kevinalle: no. that comment is about why you are making the cl. it's not explaining anything about the style. no comment is ok

app/resources/css-default.template
kevinalle: no~

Stephanie1125: i see~~ thank you~~~

PR #301, opened 2016-10-07
Make some change of the results page and fixed some bug

Stephanie1125:
+ Fixed text field for ios by setting the border-radius to 0
It was:
textboxios
Now:
iostext

+ Move red button "Mark records as duplicate" to the end of the search results and remove the "possible duplicates found" because it shows the same meaning as "Mark records as duplicate"

+ Remove the warning icon and simplify the message when there is no results for "name+location" but have results for "name"
anylocation

+ Remove the expand-link in ui=light and rearrange the name section
It was:
light1
Now:
light2

+ Small mode works with new UI


app/resources/css-default.template
```diff
@@ -216,6 +217,7 @@ input[type="text"] {
 input:not([type]):focus,  /* type attribute is not specified */
 input[type="text"]:focus {
   border: 0;
+  border-radius: 0;
```
kevinalle: there is no need to repeat properties for x and x:focus, it should inherit them. In :focus add only what changes

Stephanie1125: i see! i removed them.

Stephanie1125:
Merged~

PR #305, opened 2016-10-10
merge master to new-ui

Stephanie1125:
create a branch (liuhsinwen-mergemaster) on new-ui 
merge master to branch liuhsinwen-mergemaster
then fix the conflict and
merge branch liuhsinwen-mergemaster back to new-ui


app/full_text_search.py
```diff
@@ -149,25 +150,21 @@ def get_person_ids_from_results(query_dict, results_list):
...
+
             # use set to faster the speed.
             # average time complexity: set-O(1) list-O(n)
             if record_id in added_results:
                 continue
 
-            if is_query_match(name_query_txt, names, romanized_jp_names) and \
-                    is_query_match(location_query_txt,
-                                   locations_romanized_unidecode, locations_romanized_jp):
+            romanized_names = [value for name, value in fields.items() if name in romanized_name_fields]
```
gimite: More than 80 chars. Same for the next line.

app/full_text_search.py
```diff
@@ -149,25 +150,21 @@ def get_person_ids_from_results(query_dict, results_list):
...
+            romanized_locations = [value for name, value in fields.items() if name in romanized_location_fields]
+
+            if is_query_match(name_query_txt, romanized_names) and \
```
gimite: Use:
```python
if (... and
        ...):
```
instead of backslash. We prefer using brackets to backslashes to break lines.

app/full_text_search.py
```diff
@@ -121,15 +123,14 @@ def is_query_match(query_txt, values_romanized_unidecode,
...
-                    values_romanized_jp_words, re.I):
+            if not re.search(word, " ".join(romanized_values), re.I):
                 break
             if word_index == len(words) - 1:
                 return True
     return False
 
 
-def get_person_ids_from_results(query_dict, results_list):
+def get_person_ids_from_results(query_dict, results_list, romanized_name_fields, romanized_location_fields):
```
Stephanie1125: More than 80 chars. I will fix this.

app/full_text_search.py
```diff
@@ -246,7 +249,9 @@ def search(repo, query_dict, max_results):
     results_list = [non_romanized_person_location_index_results,
                     person_location_index_results]
 
-    index_results = get_person_ids_from_results(query_dict, results_list)
+    index_results = get_person_ids_from_results(query_dict, results_list,
+                                                returned_name_fields, returned_location_fields)
```
Stephanie1125: More than 80 chars. I will fix this.

app/full_text_search.py
Stephanie1125: i will fix all the over 80 chars. Thank you~

app/full_text_search.py
Stephanie1125: i see

tests/test_full_text_search.py
```diff
@@ -362,17 +380,36 @@ def test_search_by_name_only(self):
...
         results = full_text_search.search('haiti',
                                           {'name': 'Shinjuku',
                                            'location': 'Rin Shibuya'}, 5)
-        assert not results         
+        assert not results
+
+        # Search by Special Chinese Family Name
+        # while records are written in English
+        results = full_text_search.search('haiti', {'name': u'单鱼'}, 5)
+        assert set([r.record_id for r in results]) == \
```
Stephanie1125: It seems like in many tests, we use ' \ ' to break the line, I am wondering is there any better way to break line without using ' \ ' in this case?
Is something like
```
assert (.... ==
         ....)
```
better?

gimite: Yeah that would be better. But it would be better to do it in a separated pull request because there should be many for existing code.

Stephanie1125: okay~ I see~

Stephanie1125:
Close this because this is the same work as #307

Issue #110, opened 2015-08-06
Clearer error message when running server_tests while running local dev_appserver


gimite:
It fails, probably because of the port conflict, but the error message is cryptic.

PR #306, opened 2016-10-14
Improve an app_server error on server_test.

githaruo:
This commit is for issue #110. 
https://github.com/google/personfinder/issues/110

Sorry, I've committed with bunch of old message files. Should I merge this first?

### DIFF
--- a/tests/server_tests.py
+++ b/tests/server_tests.py
@@ -50,7 +50,8 @@ class ProcessRunner(threading.Thread):
     READY_RE = re.compile('')  # this output means the process is ready
     ERROR_RE = re.compile('ERROR|CRITICAL')  # output indicating failure
     OMIT_RE = re.compile('INFO |WARNING ')  # don't bother showing these lines
-    BIND_RE = re.compile('BindError: Unable to bind (.*):(\d*)') # this output is for appserver's port error
+    # this output is for appserver's port error
+    BIND_RE = re.compile('BindError: Unable to bind (.*):(\d+)')
     debug = False  # set to True to see all log messages, ignoring OMIT_RE
 
     def __init__(self, name, args):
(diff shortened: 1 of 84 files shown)


tests/server_tests.py
```diff
@@ -159,6 +160,13 @@ def __init__(self, port, smtp_port):
             '--datastore_consistency_policy=consistent',
         ])
 
+    def flush_output(self):
+      """Flushes the buffered output from this subprocess to stderr."""
+      self.output, lines_to_print = [], self.output
+      if lines_to_print:
+        if self.BIND_RE.search('\n'.join(lines_to_print)):
+            sys.stderr.write('%s failed port 8081 already in use.\n' % self.name)
```
gimite: Can you also suggest to turn down the local Person Finder server or the server test if any, in the error message? It would help users like you to understand the reason of the error :)

Also, can you extract the error message "port 8081 already in use" from lines_to_print instead of hard-coding it? It might be possible that it outputs BindError for other ports, and this message would be confusing in that case.

tests/server_tests.py
```diff
@@ -159,6 +160,13 @@ def __init__(self, port, smtp_port):
             '--datastore_consistency_policy=consistent',
         ])
 
+    def flush_output(self):
+      """Flushes the buffered output from this subprocess to stderr."""
+      self.output, lines_to_print = [], self.output
```
gimite: Nit pick, but lines_to_print is not a good name in this context because we won't print these lines. Maybe "original_output"?

githaruo:
Update your commented parts. #306

tests/server_tests.py
```diff
@@ -162,10 +162,12 @@ def __init__(self, port, smtp_port):
...
     def flush_output(self):
       """Flushes the buffered output from this subprocess to stderr."""
-      self.output, lines_to_print = [], self.output
-      if lines_to_print:
-        if self.BIND_RE.search('\n'.join(lines_to_print)):
-            sys.stderr.write('%s failed port 8081 already in use.\n' % self.name)
+      self.output, original_output = [], self.output
+      if original_output:
+          matchOB = self.BIND_RE.search('\n'.join(original_output), re.MULTILINE);
+          if matchOB:
```
gimite: If this doesn't match (i.e., it fails with some other reason), shouldn't we output the all original_output as we originally did?

githaruo:
Fixed your suggesting part. #306

tests/server_tests.py
```diff
@@ -164,10 +165,18 @@ def flush_output(self):
       """Flushes the buffered output from this subprocess to stderr."""
       self.output, original_output = [], self.output
       if original_output:
-          matchOB = self.BIND_RE.search('\n'.join(original_output), re.MULTILINE);
-          if matchOB:
-              sys.stderr.write('%s failed %s port %s is already in use.\n\n' %
-                               (self.name, matchOB.group(1), matchOB.group(2)))
+          error = '\n'.join(original_output)
```
gimite: Calling this as "error" is a bit weird because it's an "output", not necessarily an error message. (It can be just some logs, I guess.) How about "original_output_text"?

gimite:
LGTM. Merged. Thanks!

tests/server_tests.py
githaruo: Agree with you updated that part as well. #306

PR #313, opened 2016-10-24
Clean up some duplicate code

### LINKED ISSUE #309: Should use /add_note instead of /view to handle POST request to add a note
Currently when you submit a note in /add_note page, it posts against /view URL:
https://github.com/google/personfinder/blob/master/app/resources/add_note.html.template#L53
There is the same logic in both add_note.Handler.post and view.Handler.post, but actually add_note.Handler.post is never used.

We should let the form in /add_note to post against /add_note instead of /view, and delete view.Handler.post.

Also see #157 .

Stephanie1125:
Fixed #309 
Clean up import codes in server test cases


app/add_note.py
```diff
@@ -60,7 +60,7 @@ def get(self):
                     standalone=standalone,
                     enable_notes_url=enable_notes_url)
 
-
+    # Add a note in person's record view page
```
gimite: Use """...""" style comment for method documents.

app/add_note.py
```diff
@@ -177,4 +177,4 @@ def post(self):
                                  context='add_note')
 
         # Redirect to this page so the browser's back button works properly.
```
gimite: Update the comment here because it's no longer "this page".

Stephanie1125: ok

gimite:
LGTM. Merged.

PR #314, opened 2016-10-27
Add more detail for note status for history statistics page

Stephanie1125:
Add more detail to the statistics page for note status.
statistics

### DIFF
--- a/app/add_note.py
+++ b/app/add_note.py
@@ -160,24 +160,6 @@ class Handler(BaseHandler):
             person.latest_status not in ['believed_alive', 'is_note_author']):
             UserActionLog.put_new('mark_alive', note, person.primary_full_name)
 
-        if note.status == 'is_note_author':
-            UsageCounter.increment_counter(self.repo, 'is_note_author')
-
-        if note.status == 'believed_alive':
-            UsageCounter.increment_counter(self.repo, 'believed_alive')
-
-        if note.status == 'believed_dead':
-            UsageCounter.increment_counter(self.repo, 'believed_dead')
-
-        if note.status == 'believed_missing':
-            UsageCounter.increment_counter(self.repo, 'believed_missing')
-
-        if note.status == 'information_sought':
-            UsageCounter.increment_counter(self.repo, 'information_sought')
-
-        if note.status == '':
-            UsageCounter.increment_counter(self.repo, 'unspecified')
-
         # Update the Person based on the Note.
         if person:
             person.update_from_note(note)
--- a/app/add_note.py
+++ b/app/add_note.py
@@ -160,6 +160,24 @@ class Handler(BaseHandler):
             person.latest_status not in ['believed_alive', 'is_note_author']):
             UserActionLog.put_new('mark_alive', note, person.primary_full_name)
 
+        if note.status == 'is_note_author':
+            UsageCounter.increment_counter(self.repo, 'is_note_author')
+
+        if note.status == 'believed_alive':
+            UsageCounter.increment_counter(self.repo, 'believed_alive')
+
+        if note.status == 'believed_dead':
+            UsageCounter.increment_counter(self.repo, 'believed_dead')
+
+        if note.status == 'believed_missing':
+            UsageCounter.increment_counter(self.repo, 'believed_missing')
+
+        if note.status == 'information_sought':
+            UsageCounter.increment_counter(self.repo, 'information_sought')
+
+        if note.status == '':
+            UsageCounter.increment_counter(self.repo, 'unspecified')
+
         # Update the Person based on the Note.
         if person:
             person.update_from_note(note)
--- a/app/admin_statistics.py
+++ b/app/admin_statistics.py
@@ -29,10 +29,10 @@ class Handler(utils.BaseHandler):
         all_usage = [self.get_repo_usage(repo) for repo in repos]
         note_status_list = []
         for note_status in NOTE_STATUS_TEXT:
-            if note_status == '':
-                continue
-            note_status_list.append(note_status)
-        note_status_list.append('unspecified')
+            if not note_status:
+                note_status_list.append('num_notes_unspecified')
+            else:
+                note_status_list.append('num_notes_' + note_status)
 
         self.render('admin_statistics.html',
                     all_usage=all_usage,
@@ -51,10 +51,8 @@ class Handler(utils.BaseHandler):
                       'num_notes': getattr(counters, 'note', 0)}
 
         for note_status in NOTE_STATUS_TEXT:
-            if note_status == '':
-                repo_usage[
-                'num_notes_unspecified'] = getattr(counters, 'unspecified', 0)
-            else:
-                repo_usage[
-                'num_notes_' + note_status] = getattr(counters, note_status, 0)
+            if not note_status:
+                note_status = 'unspecified'
+            repo_usage['num_notes_' + note_status] = (
+                getattr(counters, note_status, 0))
         return repo_usage
--- a/app/admin_statistics.py
+++ b/app/admin_statistics.py
@@ -16,6 +16,8 @@
 import model
 import utils
 
+from const import NOTE_STATUS_TEXT
+
 class Handler(utils.BaseHandler):
     """Show various statistics for each repository,
     including repositories which are no longer active."""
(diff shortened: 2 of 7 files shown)


app/add_note.py
```diff
@@ -160,6 +160,24 @@ def post(self):
             person.latest_status not in ['believed_alive', 'is_note_author']):
             UserActionLog.put_new('mark_alive', note, person.primary_full_name)
 
+        if note.status == 'is_note_author':
```
gimite: 1. Use for-loop to simplify this code. You can also get a list of statuses from this constant:
https://github.com/google/personfinder/blob/master/app/const.py#L259

    * If you find yourself repeating very similar code many times, it's a bad sign. See:
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

1. Consider putting this code in Note.put_new so that you can also remove another dup code in create.py.
1. Currently db.put() is called twice, one for "notes", once for "believed_alive" for a single note, right? Can you call db.put() just once for both?

app/admin_statistics.py
```diff
@@ -32,9 +32,16 @@ def get_repo_usage(self, repo):
...
-        {'repo': haiti, 'num_persons': 10, 'num_notes': 5}
+        {'repo': haiti, 'num_persons': 10, 'num_notes': 5, ...etc}
         """
         counters = model.UsageCounter.get(repo)
         return {'repo': repo,
-                'num_persons': counters.person_counter if counters else 0,
-                'num_notes': counters.note_counter if counters else 0}
+                'num_persons': getattr(counters, 'person', 0),
+                'num_notes': getattr(counters, 'note', 0),
+                'is_note_author': getattr(counters, 'is_note_author', 0),
```
gimite: Simplify here with for-loop as well.

Also, you should prefix these keys with "num_notes" or something, otherwise it's not clear these are number of notes or persons with the status.

app/resources/admin_statistics.html.template
```diff
@@ -36,4 +36,34 @@
...
+      <th>believed_missing</th>
+      <th>information_sought</th>
+      <th>unspecified</th>
+    </tr>
+  </thead>
+  <tbody>
+  {% for usage in all_usage %}
+    <tr>
+      <td id="{{usage.repo}}-note-statistics">{{usage.repo}}</td>
+      <td id="{{usage.repo}}-is_note_author">{{usage.is_note_author}}</td>
```
gimite: Use for-loop here as well.

tests/server_test_cases/admin_statistics_tests.py
```diff
@@ -68,3 +67,45 @@ def test_note_counter(self):
...
+                      family_name='_test_family_name',
+                      text='_note_text')
+        doc = self.go_as_admin('/global/admin/statistics')
+        assert 'japan' in doc.text
+        is_note_author_counter = doc.cssselect_one('#japan-is_note_author')
+        assert is_note_author_counter.text == '1'
+
+    def test_status_counter(self):
+        """Test of counter increment based on the given status_name"""
+        def increment_counter(status_name, amount):
```
gimite: Maybe better to name it increment_counter_and_assert because it asserts as well as increments counter.

tests/server_test_cases/admin_statistics_tests.py
```diff
@@ -68,3 +67,45 @@ def test_note_counter(self):
...
+                            author_made_contact='yes',
+                            status=status_name,
+                            text='_note_text',
+                            author_name='_note_author_name')
+            doc = self.go_as_admin('/global/admin/statistics')
+            assert 'haiti' in doc.text
+            assert status_name in doc.text
+            status_name_counter = doc.cssselect_one('#haiti-' + status_name)
+            assert status_name_counter.text == str(amount)
+        increment_counter('is_note_author', 3)
```
gimite: Can you put a blank line before this line?

tests/server_test_cases/admin_statistics_tests.py
Stephanie1125: ok

tests/server_test_cases/admin_statistics_tests.py
Stephanie1125: ok

app/admin_statistics.py
Stephanie1125: ok

app/resources/admin_statistics.html.template
Stephanie1125: ok

app/model.py
```diff
@@ -649,9 +653,13 @@ def put_new(self):
         a new note is created. Also, logs user actions is updated. We should
         never call this method against an existing record."""
         db.put(self)
-        UsageCounter.increment_counter(self.repo, 'note')
         UserActionLog.put_new('add', self, copy_properties=False)
-
+        for note_status in NOTE_STATUS_TEXT:
```
gimite: I believe you don't need this for-loop. You can just directly use self.status instead of note_status.

app/model.py
```diff
@@ -649,9 +653,13 @@ def put_new(self):
...
+            if self.status == note_status:
+                if note_status == '':
```
gimite: This would be simpler:
```python
if not note_status:
    note_status = 'unspecified'
UsageCounter.increment_counter(self.repo, ['note', note_status])
```

app/admin_statistics.py
```diff
@@ -35,13 +46,15 @@ def get_repo_usage(self, repo):
...
-                'believed_missing': getattr(counters, 'believed_missing', 0),
-                'information_sought': getattr(counters, 'information_sought', 0),
-                'unspecified': getattr(counters, 'unspecified', 0)
-                }
+        repo_usage = {'repo': repo,
+                      'num_persons': getattr(counters, 'person', 0),
+                      'num_notes': getattr(counters, 'note', 0)}
+
+        for note_status in NOTE_STATUS_TEXT:
+            if note_status == '':
```
gimite: This part can be simplified as well if you write ```note_status = 'unspecified'``` for this case.

app/admin_statistics.py
```diff
@@ -35,13 +46,15 @@ def get_repo_usage(self, repo):
...
+                repo_usage[
```
gimite: Should be formatted like:
```python
repo_usage['num_notes_unspecified'] = (
    getattr(counters, 'unspecified', 0))
```

app/admin_statistics.py
```diff
@@ -25,7 +27,16 @@ class Handler(utils.BaseHandler):
     def get(self):
         repos = sorted(model.Repo.list())
         all_usage = [self.get_repo_usage(repo) for repo in repos]
-        self.render('admin_statistics.html', all_usage=all_usage)
+        note_status_list = []
+        for note_status in NOTE_STATUS_TEXT:
+            if note_status == '':
+                continue
+            note_status_list.append(note_status)
```
gimite: Optional: You can prepend "num_notes_" here instead of in the template. Then it will also prepend "num_notes" to the "id" attribute, but I believe it's fine.

app/admin_statistics.py
```diff
@@ -25,7 +27,16 @@ class Handler(utils.BaseHandler):
     def get(self):
         repos = sorted(model.Repo.list())
         all_usage = [self.get_repo_usage(repo) for repo in repos]
-        self.render('admin_statistics.html', all_usage=all_usage)
+        note_status_list = []
+        for note_status in NOTE_STATUS_TEXT:
+            if note_status == '':
+                continue
```
gimite: Optional: Maybe more simpler to add 'unspecified' here instead of below?

app/model.py
```diff
@@ -22,18 +22,23 @@
...
 import pfif
 import prefix
-from const import HOME_DOMAIN
+from const import HOME_DOMAIN, NOTE_STATUS_TEXT
+from django.template.defaulttags import register
 
 # default # of days for a record to expire.
 DEFAULT_EXPIRATION_DAYS = 40
 
+@register.filter
```
gimite: Can you put this in utils.py instead of here?

app/model.py
```diff
@@ -22,18 +22,23 @@
...
+def get_value(dictionary, key):
+    """Django Template filter"""
```
gimite: Explain what it does.

app/model.py
Stephanie1125: ok

app/model.py
Stephanie1125: ok

app/admin_statistics.py
Stephanie1125: ok

app/model.py
Stephanie1125: ok

app/model.py
Stephanie1125: i see

app/admin_statistics.py
Stephanie1125: ok~

app/admin_statistics.py
Stephanie1125: ok~

app/admin_statistics.py
Stephanie1125: i see

gimite:
LGTM. Thanks!

Issue #255, opened 2016-04-21
google.org/personfinder/global/tos should be internationalized


gimite:
It looks google.org/personfinder/global/tos always redirects to the English version:
http://www.google.com/intl/en/policies/terms/

It should link to the URL which shows localized version automatically, or change the URL based on the lang URL parameter.

myoshimu:
The issue is not happening in [current production](https://google.org/personfinder/global/home.html), but I realized that the tos page in latest code shows 500 internal server error during local test with the log below:
====
：
  File "/Users/kmiki/pf/app/tos.py", line 73, in get
    tos_url=url)
UnboundLocalError: local variable 'url' referenced before assignment
INFO     2016-10-24 01:39:44,286 module.py:788] default: "GET /global/tos HTTP/1.1" 500 228
：
====

To fix this issue, I added the initial value=None when no config url value exists.

gimite:
The original bug no longer reproduces and another bug mentioned above was fixed.

PR #316, opened 2016-10-27
fixed 500 internal server error when no config_url exists for #255

myoshimu:
this is a fix for https://github.com/google/personfinder/issues/255#issuecomment-256622358


gimite:
Thanks for the fix! LGTM. Merged.

PR #200, opened 2016-11-02
Log number of searches/adds by SMS

myoshimu:


myoshimu:
Thanks @gimite, 
Implemented the GA measurement protocol to send SMS hits.

*Sorry for confusion, it looks like my recent fix for #255 is remained in app/tos.py.

gimite:
Can you pull the latest master branch (```git pull origin master```) and push again? Then the previous change should disappear from the pull request.

myoshimu:
done, thanks:)

app/api.py
```diff
@@ -27,6 +27,11 @@
 import StringIO
 import xml.dom.minidom
 
+#For GA MP
```
gimite: Can you write "Google Analytics Measurement Protocol" instead of "GA MP"? Avoid abbreviations/acronyms in general, unless it's well known and commonly used in the project. Same for below.

Also, put a whitespace between "#" and the comment. Person Finder follows this style guide:
https://google.github.io/styleguide/pyguide.html
(Note that this is slightly different from Google's internal Python style guide.)

app/api.py
```diff
@@ -27,6 +27,11 @@
...
+import uuid
```
gimite: Sort imports in alphabetical order.

app/api.py
```diff
@@ -791,3 +798,21 @@ def get_element_text(self, doc, tag_name):
             return text.encode('utf-8')
         else:
             return None
+
+    #Send hit to GA
+    def gamp(e,ec):
```
gimite: Use more descriptive name for the method name and parameter names (gamp, e, ec).

app/api.py
```diff
@@ -791,3 +798,21 @@ def get_element_text(self, doc, tag_name):
...
+        ga_cid = uuid.uuid4()
+        params = urllib.urlencode({
+            'v': 1,
+            'tid': 'UA-17770602-2',
```
gimite: Use self.env.analytics_id instead of 'UA-17770602-2' here. We make Analytics ID configurable via config. To test it in your local environment, you can set config "analytics_id" by following instruction here:
https://github.com/google/personfinder/wiki/DeveloperFaq#how-to-readwrite-config-values

You should also handle a case where self.env.analytics_id is None. In that case, you should not send Measurement Protocol.

app/api.py
```diff
@@ -791,3 +798,21 @@ def get_element_text(self, doc, tag_name):
             return text.encode('utf-8')
         else:
             return None
+
+    #Send hit to GA
+    def gamp(e,ec):
+        ga_cid = uuid.uuid4()
+        params = urllib.urlencode({
```
gimite: Can you add a comment to point to the document of Measurement Protocol so that people can understand what these parameters are?

app/api.py
```diff
@@ -791,3 +798,21 @@ def get_element_text(self, doc, tag_name):
...
+            'tid': 'UA-17770602-2',
+            'cid': ga_cid,
+            't': 'event',
+            'ec': ec,
+            'ea': 'SMS',
+            'dp': '/sms_action'
+        })
+        url = 'http://www.google-analytics.com/collect'
+        try:
+            f = urllib2.urlopen(url, params)
```
gimite: You don't need to assign the result to ```f``` because ```f``` is not used.

app/api.py
```diff
@@ -791,3 +798,21 @@ def get_element_text(self, doc, tag_name):
...
+        except urllib2.URLError:
+            logging.exception('Caught exception fetching url')
```
gimite: Can you make the message clearer to say that it failed to send Measurement Protocol?

app/api.py
```diff
@@ -799,20 +799,23 @@ def get_element_text(self, doc, tag_name):
         else:
             return None
 
-    #Send hit to GA
-    def gamp(e,ec):
+    # ga_send_hit sends hit to Google Analytics via Measurment Protocol.
```
gimite: Use doc string """...""" for method documents:
https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

app/api.py
```diff
@@ -799,20 +799,23 @@ def get_element_text(self, doc, tag_name):
...
+    # With Measurement Protocol, you can send data by making HTTP requests.
+    # You can find more on developer guide.
+    # https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide
+    def ga_send_hit(self, event_category):
         ga_cid = uuid.uuid4()
         params = urllib.urlencode({
             'v': 1,
-            'tid': 'UA-17770602-2',
+            'tid': self.env.analytics_id,
```
gimite: Handle case where self.env.analytics_id is None.

app/api.py
```diff
@@ -799,20 +799,23 @@ def get_element_text(self, doc, tag_name):
         else:
             return None
 
-    #Send hit to GA
-    def gamp(e,ec):
+    # ga_send_hit sends hit to Google Analytics via Measurment Protocol.
+    # With Measurement Protocol, you can send data by making HTTP requests.
+    # You can find more on developer guide.
+    # https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide
+    def ga_send_hit(self, event_category):
```
gimite: "ga" is still acronym, so use "google_analytics" instead. Would be nicer to make it more natural English too e.g., "send_hit_to_google_analytics".

app/api.py
```diff
@@ -799,16 +799,26 @@ def get_element_text(self, doc, tag_name):
...
-    # You can find more on developer guide.
-    # https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide
-    def ga_send_hit(self, event_category):
-        ga_cid = uuid.uuid4()
+    def send_hit_to_google_analytics(self, event_category):
+        """Sends hit to Google Analytics via Measurment Protocol.
+        With Measurement Protocol, you can send data by making HTTP requests.
+        You can find more on developer guide.
+        https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide
+        Args:
```
gimite: Put a blank line before "Args" section.

app/api.py
```diff
@@ -799,16 +799,26 @@ def get_element_text(self, doc, tag_name):
...
+            event_category: Google Analytics Event Category.
+        """
+
+        # Check Google Analytics property setting
+        try:
```
gimite: I believe ```self.env.analytics_id``` just returns None instead of throwing exception. So you should check if the returned value is None instead of using try-except.

Also, when ```self.env.analytics_id``` is none, just don't send Measurement Protocol request, instead of using a specific default ID. It would be consistent behavior with normal Google Analytics usage of Person Finder:
https://github.com/google/personfinder/blob/1390a96cfabc0d5458354760c202de74610abb13/app/resources/analytics.html.template#L13
And we don't want to log traffic to our Analytics ID for non-production Person Finder.

app/api.py
```diff
@@ -799,16 +799,26 @@ def get_element_text(self, doc, tag_name):
...
+    def send_hit_to_google_analytics(self, event_category):
+        """Sends hit to Google Analytics via Measurment Protocol.
+        With Measurement Protocol, you can send data by making HTTP requests.
+        You can find more on developer guide.
+        https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide
+        Args:
+            event_category: Google Analytics Event Category.
+        """
+
+        # Check Google Analytics property setting
```
gimite: This comment is too obvious, so you won't need it.

In general, it's recommended to add a document for each method/class, but otherwise you should write comment only if it's not obvious what the code does or why such code is there.

app/api.py
```diff
@@ -804,28 +804,26 @@ def send_hit_to_google_analytics(self, event_category):
...
+                'ec': event_category,
+                'ea': 'SMS',
+                'dp': '/sms_action'
+            })
+            url = 'http://www.google-analytics.com/collect'
+            try:
+                urllib2.urlopen(url, params)
+            except urllib2.URLError:
+                logging.exception('Caught exception when sending Google Analytics hit')
+        else:
```
gimite: Remove this else-clause. This method doesn't return any values, so you don't need to write "return None" explicitly.

gimite:
Thanks! LGTM. Did you confirm that tools/all_tests passes with your change? Feel free to merge yourself after confirming it.

myoshimu:
Thanks! Since the tools/all_tests passed, I'm merging this update.

Issue #318, opened 2016-11-21
Merge redundant i18n messages.


githaruo:
Merge and clean up translation setting files.

PR #321, opened 2016-12-09
Merge query_name to query and removed some query_location parts.

githaruo:
Should I also merge "query_location_original"? #318.

### DIFF
--- a/app/locale/af/LC_MESSAGES/django.po
+++ b/app/locale/af/LC_MESSAGES/django.po
@@ -183,15 +183,6 @@ msgstr ""
 #, python-format
 msgid ""
 "\n"
-"            No results found for: \"%(query)s\"\n"
-"            <p>\n"
-"            Results for \"%(query_name)s\" (any location):\n"
-"          "
-msgstr ""
-
-#, python-format
-msgid ""
-"\n"
 "            No results found for: \"%(query_name)s "
 "%(query_location_original)s\"\n"
 "            <p>\n"
@@ -203,6 +194,13 @@ msgstr ""
 msgid ""
 "\n"
 "            No results found for: %(query)s\n"
+"          "
+msgstr ""
+
+#, python-format
+msgid ""
+"\n"
+"            Results for \"%(query_name)s\" (any location):\n"
 "          "
 msgstr ""
 
--- a/app/locale/af/LC_MESSAGES/django.po
+++ b/app/locale/af/LC_MESSAGES/django.po
@@ -176,32 +176,31 @@ msgstr ""
 #, python-format
 msgid ""
 "\n"
-"            No results found for: \"%(query_name)s "
-"%(query_location_original)s\"\n"
-"          "
-msgstr ""
-
-#, python-format
-msgid ""
-"\n"
-"            No results found for: \"%(query_name)s "
+"            No results found for: \"%(query)s "
 "%(query_location_original)s\"\n"
 "            <p>\n"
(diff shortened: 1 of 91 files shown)


app/resources/results.html.template
```diff
@@ -114,7 +114,7 @@
       {% if results_based_on_input %}
         <div class='resultsTitle'>
           {% blocktrans %}
-            Search Results for: {{query_name}} {{query_location}}
+            Search Results for: {{query}}
```
gimite: If I understand correctly, it now outputs "Search result for: Hiroshi" instead of "Search result for: Hiroshi Tokyo" when I search for Hiroshi in Tokyo. Is that correct? That's not what we want. We still want to see "Search result for: Hiroshi Tokyo", but we do want to use the message template "Search result for: %(query)s".

I believe the best way to do that is to concatenate query_name and query_location in Python code and assign it to template parameter "query" (which is arguments of self.render).

Also, maybe you shouldn't rename query_name param to query because it may be confusing with the "query" template parameter mentioned above.

Same for other changes in this file

app/resources/results.html.template
```diff
@@ -91,17 +91,17 @@
       {% if results %}
         <h1>
           {% blocktrans %}
```
gimite: Break this blocktrans tag into two, splitting with <p> tag.

app/resources/results.html.template
```diff
@@ -91,7 +91,7 @@
...
```
gimite: Break this blocktrans tag into two, splitting with ```<p>``` tag.

app/results.py
```diff
@@ -261,5 +264,6 @@ def get(self):
                                query_location_original=
```
gimite: Looks like these query_location_original and query_location parameters, and also query_location *variable*, are no longer used. Can you delete them?

app/results.py
```diff
@@ -247,6 +247,9 @@ def get(self):
                         results_based_on_input = False
                         query_location = ''
 
+            concatenate_query = ('%s %s' % (self.params.query_name,
```
gimite: s/concatenate_query/concatenated_query/

app/locale/af/LC_MESSAGES/django.po
```diff
@@ -183,7 +183,8 @@ msgstr ""
 #, python-format
 msgid ""
```
gimite: This entry is still here. Can you try rerunning tools/update_messages to see if this entry disappears?

app/results.py
```diff
@@ -247,7 +247,7 @@ def get(self):
                         results_based_on_input = False
                         query_location = ''
```
gimite: Can you also delete this query_location *variable* which looks no longer used? It is also referred to at line 237.

app/results.py
```diff
@@ -245,7 +245,6 @@ def get(self):
                         results = self.search(query_dict)
                         # search result not based on the user input
                         results_based_on_input = False
-                        query_location = ''
```
gimite: Can you also remove line 237?

app/resources/results.html.template
```diff
@@ -91,8 +91,10 @@
...
-            No results found for: "{{query_name}} {{query_location_original}}"
+            No results found for: "{{query}}"
```
gimite: It looks inconsistent that the query is sometimes quoted (like this line) and sometimes not (line line 119). Can you clean them up to always quote the query?

gimite:
LGTM. Thanks! Merged.

PR #323, opened 2017-01-13
Switch from Django 1.4 to Django 1.9.

gimite:
Primary motivation is to support adding description to i18n messages.

### DIFF
--- a/app/django_setup.py
+++ b/app/django_setup.py
@@ -24,9 +24,10 @@ selected; ugettext for those after (ugettext is safe to use in all Handlers)."""
 
 __author__ = 'kpy@google.com (Ka-Ping Yee)'
 
+import django
 import django.conf
 import django.template
-import django.template.loader
+import django.template.loaders.base
 import django.utils.translation
 import os
 from django.utils.translation import activate, gettext_lazy, ugettext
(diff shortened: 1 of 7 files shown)


app/django_setup.py
```diff
@@ -51,24 +49,50 @@
...
-class TemplateLoader(django.template.loader.BaseLoader):
+# It's required to call this function when we use Django features outside
+# Django framework.
+django.setup()
+
+class TemplateLoader(django.template.loaders.base.Loader):
     """Our custom template loader, which loads templates from Resources."""
-    is_usable = True  # Django requires this flag
 
-    def load_template(self, name, dirs):
```
d268: I'm not familiar with Django, but the 1.9 doc says `load_template()` is deprecated (since 1.9) and custom loader should use `get_template()` and `get_contents()` instead. Is it necessary to override `get_contents()` as well?

gimite: Yes, it's required to override `get_contents()`. Otherwise it's considered as an old format:
https://github.com/django/django/blob/master/django/template/loaders/base.py#L104

It looks usually preferred to override fine-grained methods `get_contents()` and `get_template_sources()`. But it looks more effort is needed to adopt our logic to the two methods. So I overrode `get_template()` for now.

PR #328, opened 2017-01-26
Fix #275 : Load third party libraries with PIP

### LINKED ISSUE #275: Load third party libraries with PIP
We should probably follow "Using pip requirements files" section here to install third-party libraries:
https://cloud.google.com/appengine/docs/python/tools/using-libraries-python-27
instead of directly putting third party library *.py files under "app" directory.

AwesomeTurtle:
Here we go again ! 

For testing changes : 

```
$ pip install -r requirements.txt -t app/vendors
$ rm -r app/babel app/pytz app/simplejson app/xlrd
$ ./tools/gae run --port=19001 app
```

As we said in the latest PR, we let `recaptcha` and `unidecode` into `app/` because `recaptcha` is not original library and I can not find current version of `unidecode` (latest version produces errors)
#275

### DIFF
--- a/app/appengine_config.py
+++ b/app/appengine_config.py
@@ -1,7 +1,0 @@
-#!/usr/bin/python2.7
-
-import os
-
-from google.appengine.ext import vendor
-
-vendor.add(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'vendors'))
--- a/app/appengine_config.py
+++ b/app/appengine_config.py
@@ -4,4 +4,4 @@ import os
 
 from google.appengine.ext import vendor
 
-vendor.add(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'lib'))
+vendor.add(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'vendors'))
--- a/app/appengine_config.py
+++ b/app/appengine_config.py
@@ -1,0 +1,7 @@
+#!/usr/bin/python2.7
+
+import os
+
+from google.appengine.ext import vendor
+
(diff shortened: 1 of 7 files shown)


app/appengine_config.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/python2.7
```
gimite: Do you still need this file? It looks working without this file. I believe just fix_path.py is sufficient.

app/main.py
```diff
@@ -18,6 +18,7 @@
 
 import django_setup  # always keep this first
 
+import fix_path
```
gimite: Maybe it's simpler to directly insert the content of fix_path.py here? It's only used here (after you remove it from unit_tests.py) and it's just one line.

docker/Dockerfile
```diff
@@ -1,39 +1,39 @@
...
+	python2.7 \
+	libpython2.7-dev \
+	python-pip \
+	git \
+	time \
+	gettext \
+	&& apt-get clean \
+	&& rm -rf /var/lib/apt/lists/*
+
+RUN pip install --upgrade pip && pip install pytest django==1.4 lxml==3.7.2
```
gimite: Do we really need Django? I believe it's part of AppEngine SDK.

gimite:
Thanks, looks good! Merged. Thanks for working on it.

I made a follow-up commit to remove the libraries in app directory and install the libraries in app/vendors: https://github.com/google/personfinder/commit/355f9f0e62f0985167aa1039471dec2e8b505b95

PR #335, opened 2017-02-02
Fix #228 : Try removing the hack for lang=zh-HK

### LINKED ISSUE #228: Try removing the hack for lang=zh-HK
We have implemented a hack to treat lang=zh-HK as lang=zhhk to workaround a bug in Django. The bug  may be fixed by upgrading Django to 1.4. Let's try removing the hack and just use lang=zh-HK as is.

AwesomeTurtle:
```
$ mv app/locale/zhhk/ app/locale/zh-HK/
```

Then remove workarounds in `app/utils.py` and `app/const.py`

NB: Update `Dockerfile` by adding `cssselect` and `pillow`

#228


gimite:
Thanks!

Can you separate "remove zhhk hack" part and "fix Dockerfile" part into 2 pull requests? I believe they are independent; tests on Docker should be already failing without your change.

Also, can you make the change again after pulling our latest master? I have just submitted a change to django.po files. Sorry for the conflict.

AwesomeTurtle:
I wondering if github notified you about my last commit ?

gimite:
It didn't :( Thanks for the reminder. Maybe it's a bug of Github...

Looks good to me. Did you make sure that tools/all_tests passes?

AwesomeTurtle:
Yes, everything is alright. I just ran tests it once again.

gimite:
Thanks! Merged.

PR #336, opened 2017-02-02
Fix Dockerfile for development environment

AwesomeTurtle:
Add `cssselect` and `pillow` to Docker image. We need them for executing tests.


gimite:
Thanks! Merged.

Issue #291, opened 2016-09-21
RedirectTooLongError on searching/creating record with kanji


Stephanie1125:
This error appears when searching or creating a record using kanji.

```
 (/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.3/webapp2.py:1528)
Traceback (most recent call last):
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.3/webapp2.py", line 1511, in __call__
    rv = self.handle_exception(request, response, e)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.3/webapp2.py", line 1505, in __call__
    rv = self.router.dispatch(request, response)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.3/webapp2.py", line 1253, in default_dispatcher
    return route.handler_adapter(request, response)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.3/webapp2.py", line 1076, in __call__
    handler = self.handler(request, response)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.3/webapp2.py", line 503, in __init__
    self.initialize(request, response)
  File "/base/data/home/apps/s~pf-liu/1.395792257428236030/main.py", line 518, in initialize
    self.env = setup_env(request)
  File "/base/data/home/apps/s~pf-liu/1.395792257428236030/main.py", line 440, in setup_env
    env.admin = AdminEnv(request)
  File "/base/data/home/apps/s~pf-liu/1.395792257428236030/main.py", line 48, in __init__
    self.logout_url = users.create_logout_url(self.request.url)
  File "/base/data/home/runtimes/python27/python27_lib/versions/1/google/appengine/api/users.py", line 290, in create_logout_url
    raise RedirectTooLongError
RedirectTooLongError
```

PR #337, opened 2017-02-03
Fix an issue that "create" URL becomes too long and causes an error

gimite:
#291 

Fix the issue by using create_url instead of the URL which inherits all params (which includes all unspecified params). They should be equivalent.

Also refine the logic to generate create_url and result_url to avoid generating redundant "ui" param.

### DIFF
--- a/app/results.py
+++ b/app/results.py
@@ -127,8 +127,11 @@ class Handler(BaseHandler):
             'given_name': self.params.given_name,
             'family_name': self.params.family_name,
         }
+        # Links to the default UI from the small UI. Otherwise preserves "ui"
+        # param of the current page. Note that get_url() preserves "ui" param
+        # by default (i.e., when it's not specified in +params+).
         if self.env.ui == 'small':
-            params['ui'] = None  # Links to the default UI.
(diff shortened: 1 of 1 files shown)


app/results.py
```diff
@@ -120,22 +120,27 @@ def reject_query(self, query):
...
-            family_name=self.params.family_name)
+        params = {
+            'query_name': query_name,
+            'query_location': query_location,
+            'role': self.params.role,
+            'given_name': self.params.given_name,
+            'family_name': self.params.family_name,
+        }
+        if self.env.ui == 'small':
+            params['ui'] = None  # Links to the default UI.
```
kevinalle: This does not set params.ui to self.env.ui in the default case (it did before), is that intended?

gimite: get_url() preserves subset of URL parameters (including "ui") from the current URL by default. So it still preserves "ui" param unless it's "small". Updated the comment to clarify it.

It *removes* the param from the URL if None is explicitly passed. It's a bit confusing, though.

gimite:
Thanks, Merged.

Issue #248, opened 2016-04-18
Analyze why we make so many datastore requests


gimite:
"Trace" feature in the development console shows that it makes 64 datastore requests for a single request to load a static content:

![atzs12hqcrz](https://cloud.githubusercontent.com/assets/15363/14593621/b7fa35fa-0567-11e6-82bf-d4dca7643644.png)

Would be nice to analyze why, and may optimize it to reduce latency.

gimite:
Also it may be worth consider serving some contents which are likely not overridden by resource bundles as completely static contents (e.g., jquery), or set HTTP cache header appropriately.

gimite:
#346 may be one way to solve this.

Issue #346, opened 2017-03-03
DB to NDB migration


gimite:
https://cloud.google.com/appengine/docs/standard/python/ndb/db_to_ndb

Among others, NDB API provides automatic caching of entities (explained in the page below), which should make the app faster.
https://cloud.google.com/appengine/docs/standard/python/ndb/

waseem18:
Hi @gimite ,

I'm really interested in contributing to this project and would like to start by resolving this issue.

I'll go through the complete code first and then will start the migration. Will notify you my progress here!

gimite:
Great, thanks!

@githaruo FYI

waseem18:
@gimite I have gone through the code and the `db to ndb migration` documentation provided by google cloud. I have also started the migration just now and will complete it soon! :) :+1:

waseem18:
90% of the migration is done. Will send a PR by this weekend. :)

waseem18:
Migrating tests!

gimite:
Great progress, thanks! Looking forward.

waseem18:
@gimite I'm constantly working on this in my free time. Currently busy with my office work. Will make a PR within a week. :+1:

waseem18:
@gimite I'm looking into server tests now. As soon as I'm done with it, I'll push the code to a different branch here so that you can have a look and test it.

gimite:
Hello @waseem18 ,

What is your progress so far?

It would be great if you share your change so far e.g., as a pull request, even if it's still incomplete. Then we can start reviewing it, or maybe one of us can take over your work in case you have difficulty continuing the project.

PR #348, opened 2017-03-29
Adding a memcache feature for UX improvement.

githaruo:
#248 I've added a memcache feature for UX improvement. And I think someone working on DB to NDB also accelerate web page's speed.

### DIFF
--- a/app/admin.py
+++ b/app/admin.py
@@ -24,6 +24,8 @@ from utils import *
 import const
 import tasks
 
+from google.appengine.api import memcache
+
 
 class Handler(BaseHandler):
     # After a repository is deactivated, we still need the admin page to be
@@ -160,6 +162,7 @@ class Handler(BaseHandler):
                 break
 
         config.set_for_repo(repo, **self.__view_config_to_model_config(values))
+        memcache.flush_all()
         return True
 
     def __model_config_to_view_config(self, model_config):
--- a/app/config.py
+++ b/app/config.py
@@ -26,6 +26,7 @@ import logging
 import datetime
 import utils
 from datetime import timedelta
+from google.appengine.api import memcache
 
 
 class ConfigurationCache:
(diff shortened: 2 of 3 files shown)


gimite:
Thanks for making the change!

But I found that there's a race condition issue in this implementation.

Assume there's a config named "x" with initial value "a", and memcache is not populated. And there are two processes A and B which calls these methods at the same time:

Process A: config.get("x")
Process B: config.set(x="b")

Then it's possible that the two processes execute these operations in this order:

Process A: Read from memcache and get None
Process A: Read from datastore and get "a"
Process B: Write "b" to datastore
Process B: Write "b" to memcache
Process A: Write "a" to memcache

As a result, "b" is stored in datastore while "a" is stored in memcache, and this inconsistency is not resolved until 1 hour passes. This is something we want to avoid.

This article is discussing this issue:
https://cloud.google.com/appengine/articles/best-practices-for-app-engine-memcache#code_complexity_and_concurrency
And recommend to use NDB (which internally caches to memcache) rather than implementing memcache caching by ourselves:
> When using Memcache for data that is both read and modified, use the Python NDB Client Library. 

So it sounds like we should just wait for #346 and test again to see if we need any more memcache handling. What do you think? Sorry I should have found this earlier.

On the other hand, as we chatted before, there's another potential improvement for config; to put all configs for a repository to a single datastore entry, instead of storing them separately. I created #349 for it. If you are interested, it would be great if you work on it instead which is orthogonal to #346. (You might still want to wait until #346 is submitted before working on #349 to avoid code conflict, though.)

githaruo:
I didn't notice that your suggestion is right it's hard to handle from multiple changes by this design.

Another thing I tested this version's render speed by using my browser. There is >200% faster than current version (1 sec is the average of current response time, memcache version shows 400msec-500msec). I hope that the integration to NDB will perform as much as this and also #349 makes more faster than ever seen.

Let's stop this pull request.

gimite:
Thanks, that's a great news. Did you check by uploading it to AppEngine, or with the local dev appserver? Use AppEngine for this kind of measurement because the local dev appserver is implemented in a very different way and can show very different performance.

Let's measure it again after the NDB change is in. If it's not as fast as your memcache implementation, we may want to revisit using memcache directly.

githaruo:
Of course, I uploaded it on my test app engine appspot.com and tested both **dedicated memcache** and **shared memcache** (free version). Both were similar score.

gimite:
Great, thanks!

Issue #344, opened 2017-03-02
Move the language selector to the top


gimite:
We should probably move the language selector to the top:

![nmqvnhaifei](https://cloud.githubusercontent.com/assets/15363/23500037/df745a20-ff71-11e6-996d-8fa57b1bc698.png)

It should be inside the top gray header if possible. If it's hard, it may be also acceptable to put it below the gray and white headers.

Reason: For users who open Person Finder in foreign language, it should be easy for them to notice that they can change the language.

RajibHossen:
I will work on this issue.

gimite:
Great, thanks for your contribution!

RajibHossen:
Made a pull request changing codes according to your comments.

gimite:
As I mentioned in #356, I found an issue with #350's UI on mobile, decided to move it below the header bar, and made a change in #356. Sorry for back and forth.

PR #350, opened 2017-04-05
Issue #344: Move the language selector to the top

RajibHossen:
Moved the language selection section to the top navigation bar according to the comment in the issue page.

Here's the preview of the changes.
screencapture-localhost-8000-global-home-html-1491372561964
screencapture-localhost-8000-haiti-1491372577621

### DIFF
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -177,7 +177,7 @@ div.languages {
   text-decoration: none;
   color: black;
   text-align: {{end}};
-  float: right;
+  float: {{end}};
   padding-{{end}}: 20px;
 }
 
@@ -186,7 +186,7 @@ div.languages img {
 }
 
 div.languages select {
- background: white;
+  background: white;
 }
 
 /* Repository menu */
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -177,10 +177,15 @@ div.languages {
   text-decoration: none;
   color: black;
   text-align: {{end}};
+  float:right;
+  margin-right:5%;
 }
 
 div.languages img {
   vertical-align: text-top;
+}
+div.languages select{
+ background:white;
 }
 
 /* Repository menu */
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -177,15 +177,16 @@ div.languages {
   text-decoration: none;
   color: black;
   text-align: {{end}};
-  float:right;
-  margin-right:5%;
+  float: right;
+  padding-{{end}}: 20px;
 }
 
 div.languages img {
   vertical-align: text-top;
 }
-div.languages select{
- background:white;
+
+div.languages select {
+ background: white;
 }
 
 /* Repository menu */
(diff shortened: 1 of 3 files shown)


app/resources/css-default.template
```diff
@@ -177,11 +177,16 @@ div.languages {
   text-decoration: none;
   color: black;
   text-align: {{end}};
+  float:right;
```
gimite: Can you put a whitespace after a colon? Same for below.

app/resources/css-default.template
```diff
@@ -177,11 +177,16 @@ div.languages {
...
+  margin-right:5%;
```
gimite: We usually don't put a margin with "%", I believe. How does it look with no margin?

app/resources/css-default.template
```diff
@@ -177,11 +177,16 @@ div.languages {
...
 }
 
 div.languages img {
   vertical-align: text-top;
 }
+div.languages select{
```
gimite: Put a whitespace before "{".

gimite:
Looks great, thanks for working on it! Left some comments. Can you take a look?

CI is failing, but looks for unrelated reason. So it would be OK to ignore it.

app/resources/css-default.template
RajibHossen: Without margin it looks like this
screencapture-localhost-8000-global-home-html-1491375611176

without margin and with padding-right:30px,it look like this
screencapture-localhost-8000-global-home-html-1491375636756

gimite: Thanks. Can you make it 20px padding, to balance with the padding for .title-bar-button?

Also, change padding-right to padding-{{end}}. It is automatically expanded into padding-right for left-to-right languages and padding-left for right-to-left languages.

RajibHossen: Okay. I am doing this.
One more thing->
I believe there unnecessary width:100% in .header class of sidebar.css file. For this, in my local machine there's unnecessary horizontal scroll bar at the bottom of the page. Without width:100% it works fine. could you please look at it.

app/resources/css-default.template
```diff
@@ -177,15 +177,16 @@ div.languages {
   text-decoration: none;
   color: black;
   text-align: {{end}};
-  float:right;
-  margin-right:5%;
+  float: right;
```
gimite: Oops and this one must be also float: {{end}}.

RajibHossen:
I commit changes according to your comment on my last commit.

app/resources/css-default.template
gimite: Thanks for the finding. It looks related to #351 which I have just filed. Looks like the whole .header rule in sidebar.css is not necessary. It should be fixed together with #351.

app/resources/css-default.template
```diff
@@ -177,12 +177,18 @@ div.languages {
...
+  padding-{{end}}: 20px;
 }
 
 div.languages img {
   vertical-align: text-top;
 }
 
+div.languages select {
+ background: white;
```
gimite: Put one more whitespace at the beginning of this line.

RajibHossen:
Sorry, I am rushing as a result making some silly mistakes. In future commits, I will follow all the coding convention. Please take a look now.

gimite:
No problem, now it looks good, thanks! Merged.

Issue #354, opened 2017-04-05
Fix CI failure


gimite:
CI started failing recently:
https://travis-ci.org/google/personfinder/builds/218793760

Looks like zlib is missing in the Docker image? Not sure why it suddenly started failing, though.

RajibHossen:
I want to work on this issue. Could you please provide me some guidelines or right directions about it?

AwesomeTurtle:
It seems that Pillow was updated and now zlib is required. So we need to update Dockerfile and install zlib.

RajibHossen:
I am working on this.

AwesomeTurtle:
This [issue](https://github.com/python-pillow/Pillow/issues/1412) can help. 

We have two solutions: 

+ Install zlib
+ Use a specific `Pillow` version

RajibHossen:
Hello, My host machine is Ubuntu 16.04. In my host machine, I can successfully build the image from Dockerfile, then I run the image
`~/Sources/pf$ sudo docker build -t personfinder-local .`
`~/Sources/pf$ sudo docker images`
`[sudo] password for rajib: `
`REPOSITORY           TAG                 IMAGE ID            CREATED             SIZE`
`personfinder-local   latest              ef96376d6411        2 minutes ago       822 MB`
`phusion/baseimage    latest              b6b482650b50        2 weeks ago         247 MB`
`~/Sources/pf/docker$ ./run-container.sh personfinder-local`
`root@86fda3c30dac:/opt/personfinder# pwd`
`/opt/personfinder`
`root@86fda3c30dac:/opt/personfinder# ./docker/gae-run-app.sh` 
`Starting Person Finder server`
`Setting datastore_path to /tmp/8000.datastore`
`INFO     2017-04-05 12:37:31,346 devappserver2.py:764] Skipping SDK update check.`
`WARNING  2017-04-05 12:37:31,886 simple_search_stub.py:1146] Could not read search indexes from /tmp/appengine.googlepersonfinder.root/search_indexes`
`INFO     2017-04-05 12:37:31,887 api_server.py:268] Starting API server at: http://172.17.0.2:49532`
`INFO     2017-04-05 12:37:31,930 dispatcher.py:199] Starting module "default" running at: http://172.17.0.2:8000`
`INFO     2017-04-05 12:37:31,931 admin_server.py:116] Starting admin server at: http://172.17.0.2:8001`

I believe pillow error occur in windows or mac. If this is the case then how should I proceed to solve this issue?

[...]

PR #355, opened 2017-04-05
Issue #354: Added specific version to pillow

RajibHossen:
Travis build was failing due to pillow version issue.
Changed pillow version to 4.1.0.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ RUN apt-get update && apt-get install -y \
 	&& apt-get clean \
 	&& rm -rf /var/lib/apt/lists/*
 
-RUN pip install --upgrade pip && pip install pytest lxml cssselect pillow
+RUN pip install --upgrade pip && pip install pytest lxml cssselect pillow==4.1.0
 
 # Install app engine
 WORKDIR   /opt/


gimite:
Thanks for the fix! Merged.

Looks like pillow 4.1.0 is a new version which has been just released. I found that CI has been fixed just by rerunning it (which picks up pillow 4.1.0) even without this PR. But I guess that fixing the library versions is better because it makes our CI more stable. There's pros&cons, though. We should fix the version of other libraries too if we go with this direction.

RajibHossen:
Thanks @gimite 
yeah, I also think that. We can identify which specific version we need for each package rather depending on latest version. When there is need to update the whole project, then we can determine package inconsistencies and dependencies in a more controlled way. That's my thought too.

AwesomeTurtle:
I also think that so we should freeze
 + lxml
 + cssselect
 + phusion/baseimage

And, @gimite I wondering if we need to freeze GAE too ?

gimite:
Looks like AppEngine SDK is already frozen to 1.9.50?
https://github.com/google/personfinder/blob/master/Dockerfile#L29

PR #356, opened 2017-04-10
Move the language selector below the header bars.

gimite:
Follow-up fix for #344.

#350 moved the language selector inside the header bar. But I found that it causes an issue on mobile; It gives too little space for the title. So I decided to move the language selector to just below the header bars.

Before:
huoxpyupign

After:
z4hosyxhlg1

### DIFF
--- a/app/resources/base.html.template
+++ b/app/resources/base.html.template
@@ -71,9 +71,6 @@
     onload="{{onload_function}}">
   {% block body %}
     <div class="header title-bar" role="banner">
-      {% if env.show_language_menu %}
-        {% include "language-menu.html.template" %}
-      {% endif %}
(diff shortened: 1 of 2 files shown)


gimite:
Thanks. Merged.

gimite:
> Separate from this issue, I'm wondering if it makes sense to automatically detect the language of user device/browser to show the page in user preferred language (assumed it is the same as the existing device/browser language).

I believe we originally didn't do that because we believed that the system language often doesn't mach the user's native language in developing countries (maybe the system language is often always English?), or is not properly set on feature phones. I'm not sure if it is still true today. Let's discuss with @nmakida in the meeting.

PR #364, opened 2017-08-07
Only the text part is clickable on [I'm looking for someone] button etc.

myoshimu:
The whole button should be clickable.

e.g., On:
https://google.org/personfinder/japan


gimite:
Looks like you still haven't set your git command email address?

app/resources/start.html.template
```diff
@@ -18,18 +18,18 @@
   <div class="main">
     <div class="row">
       <div class="action-outer start">
-        <div class="action-inner">
-          <a href="{{seek_url}}" tabindex="1">
+        <a href="{{seek_url}}" tabindex="1">
```
gimite: This is actually invalid in HTML standard. A tag is an inline element while DIV tag is a block element, and an inline element cannot contain a block element. See e.g.:
http://www.htmq.com/htmlkihon/005.shtml
It actually works in most browsers, but we want to be HTML standard compliant when possible.

It seems nontrivial to make a whole DIV tag as a link. It is discussed here:
https://stackoverflow.com/questions/796087/make-a-div-into-a-link

You can read and try answers there to see which works best for us.

https://stackoverflow.com/a/3494108 is popular in that entry, but I personally don't like it so much :) It looks too hacky and is not very readable.

https://stackoverflow.com/a/796096 looks like a simple good solution. But not sure if we can apply it together with our action-outer/action-inner technique, which is used to centerize the text.

https://stackoverflow.com/a/796094 is an answer many people dislike, but I feel it's not so bad in our case. You may fall back to this solution.

app/resources/start.html.template
```diff
@@ -19,14 +19,14 @@
     <div class="row">
       <div class="action-outer start">
         <a href="{{seek_url}}" tabindex="1">
```
gimite: Put this A tag back to inside the DIV tag.

app/resources/start.html.template
```diff
@@ -19,14 +19,14 @@
...
-          <div class="action-inner">
+          <div class="action-inner" onclick="location.href={{seek_url}};" tabindex="1" style="cursor:pointer;">
```
gimite: I guess it's better to do this for "action-outer" DIV instead? "action-outer" corresponds to the whole button.

Also, add a comment explaining that this is to make the whole button clickable.

I guess you need single quotes around the URL. Also, put whitespaces around `=` in JS code: `location.href={{seek_url}};` -> `location.href =
 '{{seek_url}}';`

I guess you don't need tabindex attribute here. It' sufficient if the A tag has tabindex.

Put a whitespace after `:`: `cursor:pointer` -> `cursor: pointer`

app/resources/start.html.template
```diff
@@ -19,14 +19,14 @@
...
             {% trans "I'm looking for someone" %}
           </div>
         </a>
       </div>
       <div class="action-outer end">
         <a href="{{provide_url}}" tabindex="2">
-          <div class="action-inner">
+          <div class="action-inner" onclick="location.href={{provide_url}};" tabindex="1" style="cursor:pointer;">
```
gimite: Same as above.

app/resources/start.html.template
```diff
@@ -17,19 +17,21 @@
 
   <div class="main">
     <div class="row">
-      <div class="action-outer start">
-        <a href="{{seek_url}}" tabindex="1">
-          <div class="action-inner" onclick="location.href={{seek_url}};" tabindex="1" style="cursor:pointer;">
+      {# Following onclick is for making the whole button clickable purpose. #}
```
gimite: Should be either:
* Following onclick is to make the whole button clickable.
* Following onclick is for the purpose of making the whole button clickable.

myoshimu: Sorry, I messed something during pushing the altered code.
Would you please see the[ latest code](https://github.com/google/personfinder/pull/364/files)?

gimite: I still see this your original comment:
`{# Following onclick is for making the whole button clickable purpose. #}`
following your link. Do you see something different?

app/resources/start.html.template
```diff
@@ -17,14 +17,16 @@
...
   <div class="main">
     <div class="row">
-      <div class="action-outer start">
+      {# Following onclick is to make the whole button clickable. #}
+      <div class="action-outer start" onclick="location.href = '{{seek_url}}';" style="cursor: pointer;">
         <div class="action-inner">
           <a href="{{seek_url}}" tabindex="1">
             {% trans "I'm looking for someone" %}
-          </a>
+	  </a>
```
gimite: Indent here is weird. Add one more whitespace.

gimite:
Thanks! LGTM. Merged.

PR #368, opened 2017-08-17
Add and fix a comment for translators.

gimite:
The comment for "Frequently asked questions" didn't take effect because it was not on the previous line of {%trans%} tag.

### DIFF
--- a/app/locale/af/LC_MESSAGES/django.po
+++ b/app/locale/af/LC_MESSAGES/django.po
@@ -812,6 +812,12 @@ msgid ""
 "NTT."
 msgstr "Vir fone benewens selfone, gebruik die 171-diens van NTT."
 
+msgid "For responders"
+msgstr ""
+
+msgid "Frequently asked questions"
+msgstr ""
+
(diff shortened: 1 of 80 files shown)


app/locale/af/LC_MESSAGES/django.po
```diff
@@ -812,6 +812,12 @@ msgid ""
 "NTT."
 msgstr "Vir fone benewens selfone, gebruik die 171-diens van NTT."
 
+msgid "For responders"
```
kevinalle: I don't see this message being used in the template?

gimite: It's here:
https://github.com/google/personfinder/blob/d000d626fc2bcb14b54e7cec5790add90e938bde/app/resources/static-base.html.template#L67
though it's been already there, not introduced in this pull request. I believe I forgot to run update_message script when I added the message.

kevinalle: ok!

Issue #363, opened 2017-08-04
The layout of the map is weird on mobile


gimite:
https://goo.gl/photos/ChASwkn3Mfq6mNvq7

It should be wider.

PR #369, opened 2017-08-17
Fix layout of the map section.

gimite:
#363

Remove "field" class from the map-container so that "display: flex" (applied for "field" class, which layouts its children horizontally) is not applied to the map-container.

Also applied minor HTML clean up.

Before: https://goo.gl/photos/ChASwkn3Mfq6mNvq7
After: 
lfr6ml5ubb9


gimite:
Thanks! Merged.

PR #370, opened 2017-08-21
Add temporary Japanese translations.

gimite:
Because we want the site to look good in Japanese in the preparedness week in Japan (from Aug 28), and I'm not confident if we can get translation by translators by then.


gimite:
Thanks! Merged.

Issue #362, opened 2017-07-26
Continuous Integration failed


AwesomeTurtle:
Since last build (2 months ago), CI does not success.

6 tests have failed especially into  `person_note_tests.py` file. 

Code has not changed since 2 latest months, so I suppose we have a requirement who is not fixed. And this requirements has been upgraded beetween may and today.

AwesomeTurtle:
(I am on it btw 😄 )

gimite:
Great, thanks!

AwesomeTurtle:
If you have any ideas, I reverted my fork to this commit [4c0acf6](https://github.com/google/personfinder/commit/04c0acf628b2ada903884e2b3f72bad33d398056) and tests failed always at the same place. 
Also, I fixed Python requirements with "good" version from 3 months ago. And It does not work.

PR #373, opened 2017-09-04
Fix server test failure

gimite:
#362

I believe the recent failure was because reCAPTCHA server slightly changed its behavior. Looks like it returned failure on the dummy API key + empty request before, but now it returns success.

But anyway we shouldn't rely on the real server's response because we want to fake the response.

So I replaced "test_mode" parameter with "faked_captcha_response" parameter, which allows faking a failure response as well as success.

Also, the faking is now done in utils.BaseHandler.get_captcha_response(). So it's no longer necessary for each handler to check env.test_mode.


gimite:
Thanks! Merged.

PR #382, opened 2017-10-03
Fix an issue that "write permission" checkbox state is not preserved.

### LINKED ISSUE #361: "Write permission" checkbox state is not preserved in API key management page
If you give a write permission to an API key and go back to its API key edit page, [Write permission] checkbox is unchecked. So if you save the key as is, the write permission is removed.

gimite:
Fix #361.

Initializes the state of the checkbox based on the value of domain_write_permission.


gimite:
Thanks!

PR #383, opened 2017-10-03
Fix #380: Align share buttons vertically.

### LINKED ISSUE #380: Share buttons not aligned on mobile
3sqmaagwlcw

gimite:


gimite:
Thanks for the review.

I don't really know. The all buttons have the same height, which I guess is 10px? It does work, and I just thought it is relatively better solution than other answers in Stackoverflow.

PR #385, opened 2017-10-11
Forbid direct HTTP access to dynamic resources in a better way.

gimite:
Move all static resources to app/resources/static, and allow direct HTTP access only for files in this directory.


gimite:
Thanks! Merged. (I messed up with my branch, so I merged it manually.)

PR #387, opened 2017-10-26
Fix #379 : Center the custom text in home page

### LINKED ISSUE #379: Custom text should be centered in the repository home page
Horizontal alignment of custom text, should be centered below action buttons.

vuqieerycnr

akshatnitd:


akshatnitd:
@zestyping @gimite @leeschumacher Please review :)

gimite:
Thanks for the pull request! This is how it looks now:
fxcm6kjw9fu

This is how it looks by applying your change:
gvjveyhmarp

It's better but I can still see two issues:
* The message looks too wide now. This can be fixed by adding back `max-width: 500px;`.
* It still looks the message is still not aligned with the blue and green buttons above. This is because the two buttons are not centered either.

Maybe we should center the two buttons too. It's a bit tricky because we use `float: left;` (for .main .action-outer.start) to make it responsible (stack them vertically not horizontally on mobile):
q4q6rgv8vn7
Maybe we should use other techniques like CSS media query instead to make it responsible but also centered. 

It would be great if you explore that too. Otherwise some of us will take a look when we have time.

akshatnitd:
@gimite Okay. Looking to solve it completely :+1:

akshatnitd:
@gimite Unable to build the application. Can you please help me out?
Here is the log

```
akshat (master *) personfinder $ tools/gae run app --port=8000
Setting datastore_path to /tmp/app.datastore
INFO     2017-10-27 08:16:25,906 devappserver2.py:105] Skipping SDK update check.
INFO     2017-10-27 08:16:26,004 api_server.py:300] Starting API server at: http://localhost:43931
WARNING  2017-10-27 08:16:26,010 inotify_file_watcher.py:203] There are too many directories in your application for changes in all of them to be monitored. You may have to restart the development server to see some changes to your files.
INFO     2017-10-27 08:16:26,307 dispatcher.py:251] Starting module "default" running at: http://localhost:8000
INFO     2017-10-27 08:16:26,308 admin_server.py:116] Starting admin server at: http://localhost:8001
ERROR    2017-10-27 08:16:37,063 cgi.py:122] Traceback (most recent call last):
  File "/home/akshat/Documents/git_repos/hackoctoberfest/personfinder/app/main.py", line 19, in <module>
    import django_setup  # always keep this first
  File "/home/akshat/Documents/git_repos/hackoctoberfest/personfinder/app/django_setup.py", line 27, in <module>
    import django
ImportError: No module named django
```

gimite:
This thread may provide some solutions for the error:
https://groups.google.com/forum/#!topic/google-appengine/SC8UZ6eZJw4

Alternatively, you can follow steps to set up Person Finder with Docker, which may be easier:
https://github.com/google/personfinder/tree/master/docker

There's also a prebuilt Person Finder Docker image:
https://hub.docker.com/r/gimite/personfinder/

gimite:
@akshatnitd: Do you have any update for this PR? Have you resolved the error above?

akshatnitd:
I am still unable to build it.

gimite:
Can you check your App Engine version, and update it to the latest if not? Also, have you tried solutions suggested in:
https://groups.google.com/forum/#!topic/google-appengine/SC8UZ6eZJw4
e.g., Trying this command?
`gcloud components install app-engine-python`

akshatnitd:
@gimite I was able to build the app.

I have centered the custom text and also added the ```max-width: 500px;```.

I was unable to center the buttons, maybe that can be handled in a separate PR :smile:

app/resources/start.html.template
```diff
@@ -55,36 +55,39 @@
...
         <div class="end-multi-columns"></div>
       </div>
-      <div class="row stats">
-        {% if num_people %}
-          {% blocktrans %}Currently tracking about {{num_people}} records.{% endblocktrans %}
-        {% endif %}
-      </div>
-      <div class="row custom-message">
-        {{env.start_page_custom_html|default:""|safe}}
+      <div class="row">
```
gimite: Do you have reason to break "row" div and "stats" div?

akshatnitd: Yes, it was creating problems in getting desired results. Probably, this was the way how I could fix it

gimite: I found that "custom-message" class is used in other pages too (you can grep "custom-message" in app/resources/*.html.template), but I guess we don't want to centerize it in other places. Can you leave .custom-message CSS rule as is, and instead define another rule specific to the custom message in this page, like this?:
```css
.main .row.custom-message {
  margin: 24px auto;
}
```

Then you should be able to keep the original HTML structure here (i.e., "row stats" div and "row custom-message" div).

Also, I feel text-align: center is not necessary as long as you use "auto" margin to centerize the box.

akshatnitd:
@gimite Made the changes :)

app/resources/css-default.template
```diff
@@ -597,6 +597,11 @@ td.submit {
   margin: 1em 0;
 }
 
+.main .row.custom-message {
+  margin: 24px auto;
+  text-align: center;
```
gimite: Can you remove this line as I wrote:

> Also, I feel text-align: center is not necessary as long as you use "auto" margin to centerize the box.

akshatnitd: @gimite I had tried removing it, but the centering goes away.

gimite: Oh I see. Looks like you can fix the issue by:
* Add `display: table;` in this rule
* Remove `width: 90%;` from `.custom-message` rule (which I believe is fine)

akshatnitd:
@gimite This worked perfectly :)

gimite:
Thanks! Merged.

PR #389, opened 2017-12-04
Fix #388: Serve the global home page without redirect.

### LINKED ISSUE #388: Serve the global home page without redirect
Serve the global home page at / directly instead of redirecting it to /global/home.html.

gimite:
This makes the home page loading faster and make its URL nicer.


gimite:
Thanks! Merged.

Issue #390, opened 2017-12-05
AMP page contains script tags


githaruo:
Hi team,

I found some script tags in the [AMP start page](https://www.google.org/personfinder/japan/amp_start?#development=1). See console on your chrome browser.

gimite:
Oh I see. I wonder why it's not caught by:
https://search.google.com/test/amp?id=TK4IrNMOAgVCDyq-WHULIA

The two scripts are for Analytics and Google Tag Manager. Looks like they both provide a way to support AMP. Can you take a look? @myoshimu may know some hints.

githaruo:
I chatted with @myoshimu about this part. I'll add amp-analytics tags to gtm.html.template.

gimite:
Thanks for taking a look!

PR #393, opened 2018-06-01
Added AMP GTM tags for AMP pages.

githaruo:
This pull request is for issue #390. Also we need to prepare AMP Google Tag Managers. I just create an empty AMP GTM container for this purpose.

### DIFF
--- a/app/resources/base.html.template
+++ b/app/resources/base.html.template
@@ -23,7 +23,9 @@
   {% block head %}
     {# TODO(haruo) Consider AMP page encoding to utf-8 only. #}
     <meta charset="{{env.charset}}">
-    {% include "gtm.html.template" %}
+    {% if not env.amp %}
+      {% include "gtm.html.template" %}
+    {% endif %}
     <meta name="keywords" content="{{config.keywords}}">
     <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
     {% if env.amp %}
@@ -64,6 +66,9 @@
         {% endwith %}
       {% endif %}
     </style>
+    <!-- AMP Analytics -->
+    <script async custom-element="amp-analytics"
+            src="https://cdn.ampproject.org/v0/amp-analytics-0.1.js"></script>
   {% else %}
     <link
        rel="stylesheet"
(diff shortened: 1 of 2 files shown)


app/resources/base.html.template
```diff
@@ -64,6 +66,9 @@
         {% endwith %}
       {% endif %}
     </style>
+    <!-- AMP Analytics -->
```
gimite: Optional: It may be easier to read if you put both analytics implementations at the same place:

```
{% if env.amp %}
  <script async custom-element="amp-analytics"
              src="https://cdn.ampproject.org/v0/amp-analytics-0.1.js"></script>
{% else %}
  {% include "gtm.html.template" %}
{% endif %}
```

gimite: Use `{# AMP Analytics #}` instead. Then it's removed from the generated HTML.

app/resources/base.html.template
```diff
@@ -102,6 +107,10 @@
     class="{{env.ui}} {% block extra_body_class %}{% endblock %}"
     {% if not env.amp %}onload="{{onload_function}}"{% endif %}>
   {% block body %}
+    {% if env.amp %}
+      <amp-analytics config="https://www.googletagmanager.com/amp.json?id=GTM-MVDGQ3H"
```
gimite: Is this id=GTM-MVDGQ3H specific to google.org domain? If so, put the value in `config` to make it customizable.

app/resources/base.html.template
githaruo: I tried it but AMP Analytics require this position. @myoshimu can we move normal GTM tag to same place?

githaruo: done

app/resources/base.html.template
githaruo: done

app/resources/base.html.template
githaruo: Also I chatted with @myoshimu we need separate account for AMP GTM so this tag is essential for this purpose.

githaruo:
Note for Googlebots CLA part I tried login and I wad covered by agreement due to first commit I forgot to add my email address so that's happened.

PR #394, opened 2018-06-19
Docker file modification for TravisCI

githaruo:
Added --user to pip install part.


githaruo:
There seems still causing error on TravisCI. I'll keep researching on it.

githaruo:
Put some particular version pip packages to avoid test errors for tentative solution.

Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN apt-get update && apt-get install -y \
 	&& apt-get clean \
 	&& rm -rf /var/lib/apt/lists/*
 
-RUN pip install --upgrade pip && pip install pytest lxml cssselect pillow==4.1.0
+RUN pip -vv install --upgrade pip==9.0.1 && pip -vv install pytest==3.6.0 lxml cssselect pillow==4.1.0
```
gimite: What is `-vv`? Is it needed?

gimite: Just curious, what was the problem using pip 10?

app/main.py
```diff
@@ -217,18 +217,31 @@ def select_charset(request):
 
 def select_lang(request, config=None):
     """Selects the best language to use for a given request.  The 'lang' query
-    parameter has priority, then the django_language cookie, then the first
-    language in the language menu, then the default setting."""
-    default_lang = (config and
-                    config.language_menu_options and
-                    config.language_menu_options[0])
+    parameter has priority, then the django_language cookie, then
```
gimite: Looks like your pull request contains changes from other changes. Maybe try git pull origin master and git push again?

Dockerfile
githaruo: This link helps you to understand easily. There is a big change on pip 10 they changed basic file structure of it and it causes error for apt distribution packages. Current temporary solution requires code modification or set pip 9.
https://github.com/pypa/pip/issues/5240

githaruo: I think this part is typo. I just want to show verbose mode to check errors, but actual arg are -v or --verbose. But it seems working and show verbose logs I will change this part to --verbose to help reviewers.

app/main.py
githaruo: I will do that.

Dockerfile
gimite: Thanks, but it looks not recommended to use pip 9:
https://github.com/pypa/pip/issues/5240#issuecomment-383303439
and instead they suggested some workarounds:
https://github.com/pypa/pip/issues/5221#issuecomment-382069604

Did you try those, like `pip install --upgrade --user pip` or `hash -r pip`?

It's optional (so far we don't see any issue using pip 9), but would be nice if we can use the latest version.

gimite: Or you can remove `--verbose` for now, since people usually won't need detailed debug log of pip (when it works). We can add it back later for analysis when needed again. It's up to you though.

githaruo: I tested different way to avoid apt package today and it's succeeded. I'll push that version which uses pip 10.

gimite: Great, thanks!

PR #398, opened 2018-07-09
Fix a bug that an extra whitespace is inserted to the display query.

gimite:
Before: Search results for: "ichikawa "
After: Search results for: "ichikawa"


myoshimu:
LGTM

PR #403, opened 2018-07-22
AMP Variable for Google Analytics ID implementation

myoshimu:
Right now AMP GTM variable is not implemented to capture Google Analytics ID.
We should implement it to allow dynamic GA property ID implementation.

### DIFF
--- a/app/resources/base.html.template
+++ b/app/resources/base.html.template
@@ -110,6 +110,7 @@
     {% if env.amp and env.enable_analytics %}
       <amp-analytics config="https://www.googletagmanager.com/amp.json?id={{config.amp_analytics_id}}"
                      data-credentials="include">
+        //AMP variable for dynamic GA property setting
         <script type="application/json">
         {
           "vars": {
--- a/app/resources/base.html.template
+++ b/app/resources/base.html.template
@@ -109,7 +109,15 @@
   {% block body %}
     {% if env.amp and env.enable_analytics %}
       <amp-analytics config="https://www.googletagmanager.com/amp.json?id={{config.amp_analytics_id}}"
-                     data-credentials="include"></amp-analytics>
+                     data-credentials="include">
+        <script type="application/json">
+        {
+          "vars": {
+          "gaProperty": "{{env.analytics_id}}"
+          }
+        }
+  </script> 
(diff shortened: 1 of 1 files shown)


app/resources/base.html.template
```diff
@@ -109,7 +109,16 @@
   {% block body %}
     {% if env.amp and env.enable_analytics %}
       <amp-analytics config="https://www.googletagmanager.com/amp.json?id={{config.amp_analytics_id}}"
-                     data-credentials="include"></amp-analytics>
+                     data-credentials="include">
+        //AMP variable for dynamic GA property setting
+        <script type="application/json">
+        {
+          "vars": {
+          "gaProperty": "{{env.analytics_id}}"
```
gimite: Indent this line by 2 more whitespaces.

app/resources/base.html.template
```diff
@@ -109,7 +109,16 @@
...
+          }
+        }
+  </script>
```
gimite: Fix indent here (align with `<script>` above).

app/resources/base.html.template
githaruo: I think your suggested key name amp_gtm_id is better than config.amp_analytics_id. If possible can you rename this part too?

myoshimu: Thanks gimite and githaruo!

@githaruo 
Can I just fix the line 111 or should I modify anywhere else to update the variable name?

app/resources/base.html.template
myoshimu: done, thanks!

app/resources/base.html.template
gimite: There are a few other references too. Try searching in the code repository:
https://github.com/google/personfinder/search?q=amp_analytics_id&unscoped_q=amp_analytics_id

gimite: Looks the indent on **this line** (line 117) hasn't been fixed.

myoshimu: done, thanks!

myoshimu: Regarding the amp_analytics_id variable, let me raise another issue to fix them.

PR #412, opened 2018-08-20
Fix #411: Add "(First name)" / "(Surname)" to name labels.

### LINKED ISSUE #411: First Name, Last Name labels along with or instead of Given Name, Family Name labels currently in place
Why adding information about someone the page displays Given Name and Family Name fields in the page. 
Which could be confusing to some.
Can we have first name , last name labels as well on the page.

gimite:
Manually inherited the existing translation only for Malayalam (ml) because:
* It's a second language of the active repository, so I don't want to lose translation of the language now
* Confirmed with Malayalam speaker that the existing translation is clear enough

See #411 for context.

### DIFF
--- a/app/locale/af/LC_MESSAGES/django.po
+++ b/app/locale/af/LC_MESSAGES/django.po
@@ -795,11 +795,11 @@ msgstr ""
 msgid "Failed to obtain search result for the phone number."
 msgstr "Kon nie soekresultaat vir die foonnommer kry nie."
 
-msgid "Family name"
-msgstr "Van"
+msgid "Family name (Surname)"
+msgstr ""
 
-msgid "Family name (required)"
-msgstr "Van (vereis)"
+msgid "Family name (Surname) (required)"
+msgstr ""
 
 msgid "Feed of updates about this person"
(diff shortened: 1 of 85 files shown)


app/locale/af/LC_MESSAGES/django.po
```diff
@@ -795,11 +795,11 @@ msgstr ""
 msgid "Failed to obtain search result for the phone number."
 msgstr "Kon nie soekresultaat vir die foonnommer kry nie."
 
-msgid "Family name"
-msgstr "Van"
+msgid "Family name (Surname)"
+msgstr ""
```
nworden: It looks like this change deletes the msgstrs for name label translations for every language besides en and ml. Is that expected?

gimite: Yes, that's the default behavior by rerunning tools/update_messages because messages are keyed by the English text. They will be retranslated in the next translation cycle.

I manually updated Malayalam translation as I wrote in the pull request description.

nworden: I see, ok.

Issue #416, opened 2018-08-24
tools/all_tests depends on old version of pytest


nworden:
tools/all_tests uses _pytest.config.Parser, but Parser appears to have moved to the argparsing module in version 3.6.2.
I'll try to find a way to avoid depending directly on Parser; since they moved it without a mention in the release notes, I'm not sure they meant for it to be used like that. Otherwise I'll just update the reference to it to use its new module.

nicoddemus:
> I'll try to find a way to avoid depending directly on Parser; since they moved it without a mention in the release notes, I'm not sure they meant for it to be used like that. 

Although not part of the public API, it definitely should be at least mentioned in the CHANGELOG, sorry about that. 👍

nworden:
No apology needed!

gimite:
Actually it might be safer to reimplement the parser in our code base than depending on a private API, though it's redundant. We are probably interested in a small subset of the flags anyway (I personally have only used `-k`).

PR #417, opened 2018-08-24
Update unit_tests.py to not change pytest option parsing.

nworden:
unit_tests.py previously changed an internal pytest option parsing
function, as a way of using the TESTS_DIR directory by default. pytest
now has a way to be called programmatically with arguments:
https://docs.pytest.org/en/latest/usage.html#calling-pytest-from-python-code
So the unit tests script can just check if any of the command-line
arguments are a test directory or something else to test (vs. just
options like --tb=short) and, if none of them are, add TESTS_DIR as a
default. Then that list of arguments can be passed to pytest.main.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -37,11 +37,11 @@ os.environ['APPLICATION_ID'] = 'personfinder-unittest'
 os.chdir(os.environ['APP_DIR'])
 
 # ...but we want pytest to default to finding tests in TESTS_DIR, not the cwd.
-# So, when no arguments are given, we make the option parser return TESTS_DIR.
-import _pytest.config
-original_parse_setoption = _pytest.config.Parser.parse_setoption
-_pytest.config.Parser.parse_setoption = (lambda *args, **kwargs: 
-    original_parse_setoption(*args, **kwargs) or [os.environ['TESTS_DIR']])
+# So, when no arguments besides options are given, we add TESTS_DIR to the list
+# of options passed to pytest.
+args = sys.argv[1:]
+if all([arg.startswith('-') for arg in args]):
+  args += [os.environ['TESTS_DIR']]
 
 # Run the tests, using sys.exit to set exit status (nonzero for failure).
-sys.exit(pytest.main())
+sys.exit(pytest.main(args))


nworden:
This is to fix issue #416.

tests/unit_tests.py
```diff
@@ -37,11 +37,11 @@
...
 # ...but we want pytest to default to finding tests in TESTS_DIR, not the cwd.
-# So, when no arguments are given, we make the option parser return TESTS_DIR.
-import _pytest.config
-original_parse_setoption = _pytest.config.Parser.parse_setoption
-_pytest.config.Parser.parse_setoption = (lambda *args, **kwargs: 
-    original_parse_setoption(*args, **kwargs) or [os.environ['TESTS_DIR']])
+# So, when no arguments besides options are given, we add TESTS_DIR to the list
+# of options passed to pytest.
+args = sys.argv[1:]
+if all([arg.startswith('-') for arg in args]):
```
gimite: One typical usage of tools/unit_test is e.g.:

    $ tools/unit_test -k foo

to filter tests. This doesn't pass this check I believe because `foo` doesn't start with `-`. So we do need to call the parser unfortunately?

nworden: Is there something special about the letter k?

With this commit, that wouldn't work, but you can run a specific test by just passing it as a regular argument, e.g.:
`$ tools/unit_tests test_pfif`
or you can run multiple like this:
`$ tools/unit_tests test_pfif test_photo`

nworden:
I have a new pull request (#423) to fix the Travis CI (long story short, it looks like the old version of pytest doesn't work with the new version of something else, so I set the Dockerfile to use the new version of pytest). I included this commit in that branch since it wouldn't work otherwise, so this pull request is kinda obsolete now. I'm leaving it open in case it's more convenient to continue the discussion about tools/unit_tests arguments here though.

tests/unit_tests.py
gimite: Oh I didn't know that I can pass test names as regular arguments. But when you write:
```$ tools/unit_tests test_pfif```
doesn't it also prevent it from adding TESTS_DIR to `args`?

Here is document of -k option (run `pytest -h` to get this). I guess it's more flexible than regular arguments?:
```
  -k EXPRESSION         only run tests which match the given substring
                        expression. An expression is a python evaluatable
                        expression where all names are substring-matched
                        against test names and their parent classes. Example:
                        -k 'test_method or test_other' matches all test
                        functions and classes whose name contains
                        'test_method' or 'test_other', while -k 'not
                        test_method' matches those that don't contain
                        'test_method' in their names. Additionally keywords
                        are matched to classes and functions containing extra
                        names in their 'extra_keyword_matches' set, as well as
                        functions which have names assigned directly to them.
```

nworden: Ah, I see what you're saying. Yeah, this isn't going to handle it correctly. I've just gone back to what it was doing before with the parser and switched it to using the new package. I did it on my other branch (pull request #423). I'll close this pull request now.

PR #420, opened 2018-08-25
Center action buttons on repo homepage.

nworden:
I couldn't find a way to get it right on both mobile and desktop without
using @media to add desktop-specific rules, but that's been around for a
while (IE9 supports them) so it seems OK. What do you think?

### DIFF
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -612,12 +612,15 @@ td.submit {
   line-height: 10px;
 }
 
+/* We want to display the "I have info"/"I'm seeking info" buttons side-by-side
+   on desktop, but stacked on top of each other on mobile, so we have an extra
+   couple CSS rules for larger screens. */
 @media (min-width: 700px) {
   .main .action-wrapper {
     display: inline-block;
   }
   .main .action-outer {
-    float: left;
+    display: inline-block;
     margin: 10px;
   }
 }
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -552,10 +552,6 @@ td.submit {
   text-align: center;
 }
 
-.action-wrapper {
-  text-align: center;
-}
-
 .main .action-outer {
   border: none;
   font-size: 20px;
@@ -616,9 +612,6 @@ td.submit {
    on desktop, but stacked on top of each other on mobile, so we have an extra
    couple CSS rules for larger screens. */
 @media (min-width: 700px) {
-  .main .action-wrapper {
-    display: inline-block;
-  }
   .main .action-outer {
     display: inline-block;
     margin: 10px;
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -544,6 +544,7 @@ td.submit {
 
 .main .row {
   margin: 24px;
+  text-align: center;
 }
 
 .main .stats {
@@ -551,11 +552,14 @@ td.submit {
   text-align: center;
 }
 
+.action-wrapper {
+  text-align: center;
+}
+
 .main .action-outer {
   border: none;
   font-size: 20px;
   margin-bottom: 1em;
-  margin-{{start}}: 40px;
   max-width: 240px;
   min-height: 70px;
   padding: 16px;
@@ -569,15 +573,16 @@ td.submit {
   text-decoration: none;
 }
 
+.main .action-outer {
+  margin: 0 auto 10px;
+}
+
 .main .action-outer.start {
   background: #4285f4;
-  float: {{start}};
-  margin-{{end}}: 18px;
 }
 
 .main .action-outer.end {
   background: #15a04a;
-  float: {{start}};
 }
 
 .main .action-inner {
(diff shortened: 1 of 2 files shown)


app/resources/css-default.template
```diff
@@ -607,6 +612,16 @@ td.submit {
   line-height: 10px;
 }
 
+@media (min-width: 700px) {
+  .main .action-wrapper {
+    display: inline-block;
+  }
+  .main .action-outer {
+    float: left;
```
gimite: Use `float: {{start}}` instead, which flips left/right on RTL.

gimite: The goal here is to layout horizontally instead of virtually, right? Can't you use inline-block here too?

app/resources/css-default.template
```diff
@@ -607,6 +612,16 @@ td.submit {
   line-height: 10px;
 }
 
+@media (min-width: 700px) {
```
gimite: Add a comment why it has separated style for large screen.

app/resources/start.html.template
```diff
@@ -29,31 +29,33 @@
...
+            <div class="action-inner">
+              <a href="{{provide_url}}" tabindex="2">
+                {% trans "I have information about someone" %}
+              </a>
+            </div>
           </div>
-        </div>
-        <div class="end-multi-columns"></div>
+          <div class="end-multi-columns"></div>
+	</div>
```
gimite: Indent by 2 more whitespaces.

app/resources/start.html.template
```diff
@@ -29,31 +29,33 @@
...
-        <div class="action-outer start"
-          {% if not env.amp %}
-            onclick="location.href = '{{seek_url}}';" style="cursor: pointer;"
-          {% endif %}
-        >
-          <div class="action-inner">
-            <a href="{{seek_url}}" tabindex="1">
-              {% trans "I'm looking for someone" %}
-            </a>
+        <div class="action-wrapper">
```
gimite: Why did you introduce a new div? Can't you apply the styles for `row` instead?

gimite:
Use of @media query should be OK. If possible, can you test with major browsers including IE to see if it looks OK?

app/resources/css-default.template
nworden: Thanks! It turns out you were right about inline-block (below) so I don't need to use float after all.

nworden: Yep, done.

app/resources/start.html.template
nworden: I couldn't find a way to do it without a wrapper div around the two buttons to center them together. If I just apply all the rules to the row class, it ends up left-aligned instead of centered.

nworden:
Screenshots for:
- [Chrome, desktop](https://user-images.githubusercontent.com/9059304/45258360-0c680a00-b384-11e8-8223-f101cd3a7080.png)
- [Chrome, mobile emulation](https://user-images.githubusercontent.com/9059304/45258363-125deb00-b384-11e8-9ab9-66d78a94c8d4.png)
- [Firefox](https://user-images.githubusercontent.com/9059304/45258365-17229f00-b384-11e8-8aac-53df38d007f2.png)
- [IE Edge](https://user-images.githubusercontent.com/9059304/45258368-1b4ebc80-b384-11e8-9942-dd1e99c2ccfc.png)

app/resources/start.html.template
gimite: Looks like you have a tab character ("\t") on this line? Use whitespaces instead of tabs. Maybe we should have presubmit check for it...

app/resources/start.html.template
gimite: OK, then it's fine, but just curious, did you find the reason why it ended up left-aligned? Maybe due to some CSS rule conflict?

nworden:
Sorry, I pushed the tab-deletion commit before I replied, and now GitHub's automatically "resolved" the comment threads and I can't find a way to reply directly to them, so I'll put replies here:
Regarding the tab: wow, nice catch, you're right, there was a tab.
Regarding the left-alignment: I think you're right, I think it was the display-inline. I have something without the wrapper that seems to work on Chrome, so I'll re-test it on IE and Firefox and then push the new version.

nworden:
Screenshots for the latest code:
- [Chrome, desktop](https://user-images.githubusercontent.com/9059304/45323997-dd37d100-b51a-11e8-80d5-983f582ccbe1.png)
- [Chrome, mobile emulation](https://user-images.githubusercontent.com/9059304/45324002-df019480-b51a-11e8-99ad-f404afbef595.png)
- [Firefox](https://user-images.githubusercontent.com/9059304/45324025-f6408200-b51a-11e8-9499-f3c077c8dbea.png)
- [IE Edge](https://user-images.githubusercontent.com/9059304/45324034-0193ad80-b51b-11e8-90f8-a0edb44d843b.png)
- [IE 11](https://user-images.githubusercontent.com/9059304/45324051-0bb5ac00-b51b-11e8-9a18-28829c34380b.png)

PR #421, opened 2018-09-01
Avoid horizontal stretch of images on results page.

nworden:
height was set to auto but not width, which let tall images get
stretched out.

### DIFF
--- a/app/resources/css-default.template
+++ b/app/resources/css-default.template
@@ -900,7 +900,7 @@ label.option {
 }
 
 img.resultImage {
-  width: 80px;
+  width: auto;
   height: auto;
   max-width: 80px;
   max-height: 80px;


nworden:
This looks different from the pytest version issue. 25 tests fail with the same message (from the same place):
AttributeError: '_io.BufferedWriter' object has no attribute 'file'
the stack trace that gets printed out starts and ends in /usr/local/lib/python2.7/, so it's not clear what the root of the problem is.

These tests all pass locally, so I assume the issue has something to do with Docker. Could you restart it (I don't have permission)? Knowing if it's transient seems like a good place to start debugging.

nworden:
Never mind, I found a fix for Travis, it's in pull request #423.

gimite:
Thanks for fixing the test!

PR #422, opened 2018-09-03
Don't approve this pull request.

nworden:
This is just a test to see if the Travis issues are actually because of code changes. They're failing at head now, even though the only change was to add scrollbars to code snippets (and everything passed when I originally opened that PR). This PR would undo that change. If the tests fail for this PR too, it means that something besides code changes are for sure causing the failures.


nworden:
It did fail, with the same message as at head:
https://travis-ci.org/google/personfinder/builds/424009896

PR #424, opened 2018-09-03
Add cssselect and lxml to requirements.txt.

nworden:


requirements.txt
```diff
@@ -1,4 +1,6 @@
 Babel==2.3.4
+cssselect==1.0.3
+lxml==4.2.4
```
gimite: It specifies use of lxml from AppEngine:
https://github.com/google/personfinder/blob/6ee9de5829cdb7e81cf9ab34125a95b0570c4bb8/app/app.yaml#L46
So I thought you don't need to add it here. Did you encounter any issue without adding it here?

By adding it here, it adds lxml to app/vendors once you follow instruction here:
https://github.com/google/personfinder/blob/master/app/vendors/README.md
which causes unnecessary increase of uploaded files, which I want to avoid unless necessary.

Not sure about cssselect, but maybe it's part of lxml in AppEngine?

nworden: I did need to install cssselect to get things to run. I'm not sure about lxml though (I just added it because I saw it listed as a requirement in a Docker file), so I've removed that.

gimite: I see. Can you:

1. Follow https://github.com/google/personfinder/blob/master/app/vendors/README.md and add changes to app/vendors as well
2. Run PF on AppEngine (not locally) to make sure it works without problems after making the change

gimite:
It may make sense to separate requirements.txt for local development and AppEngine? It would still be nice to make local environment setup easier using requirements.txt than installing one by one.

PR #428, opened 2018-09-08
Clean up some pytest warnings.

nworden:
pytest tries to collect tests from TestCase and TestPerson because of their names, and then raises warnings because that doesn't work out. This PR prefixes their names with underscores to clean up the warnings.

### DIFF
--- a/tests/test_pfif.py
+++ b/tests/test_pfif.py
@@ -56,7 +56,7 @@ def dict_to_entity(dict):
     return entity
 
 
-class _TestCase:
+class PFIFTestCase:
     """A container for test data and expected results."""
     def __init__(self, pfif_version, xml, person_records=[], note_records=[],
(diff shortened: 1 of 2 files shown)


tests/test_pfif.py
```diff
@@ -56,7 +56,7 @@ class Entity:
     return entity
 
 
-class TestCase:
+class _TestCase:
```
gimite: Can you comment the reason of "_" since it's a bit unusual naming?

Can be workarounded in other namings like `PFIFTestCase` or `PersonForTest`.

nworden: The underscore was just meant as Python's "private" signifier; it made sense to me since the classes are private to the test. I've switched them to the more specific names you suggested though.

gimite: I see. I haven't seen the underscore used for class names, but that sounds reasonable. You can leave the underscore too if you prefer.

PR #429, opened 2018-09-09
Thumbnails for results page.

nworden:
The feature request for this was filed internally (b/2397238).


app/photo.py
```diff
@@ -71,6 +72,29 @@ def create_photo(image, handler):
     photo_url = get_photo_url(photo, handler)
     return (photo, photo_url)
 
+
+def set_thumbnail(photo):
```
gimite: Add doc comment for this method.

app/tasks.py
```diff
@@ -437,3 +438,19 @@ def _should_notify(self, count_of_unreviewed_notes):
...
 
+
+class ThumbnailPreparer(utils.BaseHandler):
+
+    repo_required = False
+    ACTION = 'tasks/thumbnail_preparer'
+
+    def get(self):
+        if self.repo:
+            for p in (model.Photo.all()
```
gimite: It may be rare, but this may not finish within task queue timeout (10 min?). Can you handle timeout and spawn a new task to continue the task? See https://github.com/google/personfinder/blob/23d7030eec886204c06598e5403700aa11dad16d/app/tasks.py#L86 for reference. It may be nice to define a common subclass to deal with the retry-on-timeout logic.

app/photo.py
```diff
@@ -71,6 +72,29 @@ def create_photo(image, handler):
...
+        image.resize(MAX_THUMBNAIL_DIMENSION,
+                     image.height * MAX_THUMBNAIL_DIMENSION / image.width)
+    else:
+        image.resize(image.width * MAX_THUMBNAIL_DIMENSION / image.height,
+                     MAX_THUMBNAIL_DIMENSION)
+    try:
+        thumbnail_data = image.execute_transforms(output_encoding=images.PNG)
+    except RequestTooLargeError:
+        raise SizeTooLargeError()
+    except Exception:
```
gimite: Any specific exceptions you want to catch here? Would be better to catch more specific exception types here rather than catching all Exception.

app/utils.py
```diff
@@ -954,6 +955,15 @@ def should_show_inline_photo(self, photo_url):
         else:
             return True
 
+    def get_thumbnail_url(self, photo_url):
+        if not photo_url:
+            return None
+        parsed_url = list(urlparse.urlparse(photo_url))
```
gimite: Users can either specify an external photo URL or upload their local photo to PF when creating a record. So `photo_url` can be an arbitrary external URL. Can you add a check for it?

app/utils.py
```diff
@@ -954,6 +955,15 @@ def should_show_inline_photo(self, photo_url):
...
+        params_dict = dict(urlparse.parse_qsl(parsed_url[4]))
```
gimite: Can you define a constant (e.g., `QUERY_INDEX`) for a magic number 4?

app/utils.py
```diff
@@ -954,6 +955,15 @@ def should_show_inline_photo(self, photo_url):
         else:
             return True
 
+    def get_thumbnail_url(self, photo_url):
```
gimite: Add doc comment for this method.

tests/server_test_cases/photo_tests.py
```diff
@@ -100,3 +102,32 @@ def test_upload_broken_photo(self):
         assert not doc.cssselect('img.photo')
         assert 'unrecognized format' in doc.text
 
+    def test_set_thumbnail(self):
+        """Tests that a thumbnail is generated."""
+        photo = model.Photo.create(
+            'haiti', image_data=file('tests/testdata/small_image.png').read())
```
gimite: Can you close the file, maybe using `with` syntax? I guess it won't close automatically? We should fix other tests too, I guess...

app/utils.py
nworden: Thanks, added a check in app/results.py (so this doesn't get called for non-uploaded photos).

app/tasks.py
nworden: I think it's actually ok if it doesn't finish, since the cron job's next run will start soon anyway. And automatic restarts would risk having two tasks running at once, which is kinda messy since they'd both be trying to do the same thing.

app/photo.py
nworden: This is based on the general resizing function above, which catches Exception. It looks like there's a variety of images package and other errors that can come up so they decided to catch everything to be safe. I've copied the comment there down to my function too.

tests/server_test_cases/photo_tests.py
nworden: Done for all the tests in the file.

app/tasks.py
gimite: Oh I see, so the key here is that you set `filter('thumbnail_data =', None)` so records which it has generated a thumbnail for will be skipped in the next task.

Can you add a comment explaining why it has no (explicit) retry logic?

app/tasks.py
```diff
@@ -437,3 +438,19 @@ def _should_notify(self, count_of_unreviewed_notes):
         return  count_of_unreviewed_notes > self.config.get(
                 'unreviewed_notes_threshold')
 
+
+class ThumbnailPreparer(utils.BaseHandler):
```
gimite: Add document comment for this class.

app/utils.py
```diff
@@ -954,6 +955,24 @@ def should_show_inline_photo(self, photo_url):
         else:
             return True
 
+    URL_PARSE_QUERY_INDEX = 4;
```
gimite: Remove the semicolon.

You can optionally move it to inside get_thumbnail_url() unless you plan to use it in other methods.

PR #432, opened 2018-09-13
Make feedback link configurable

nworden:
It's currently hard-coded to use a support.google.com feedback URL, even when the instance is set to "brand as other organization."


nworden:
I've set it in production.

PR #434, opened 2018-09-13
Fix #400: use query param as query_name when it's used

### LINKED ISSUE #400: Bring back support of "query" parameter for /results
We renamed "query" to "query_name" a while ago, but some third parties directly link to a URL with "query". So we should support both.

Example:
http://google.org/personfinder/japan/results?role=seek&lang=ja&query=mike

nworden:


app/results.py
```diff
@@ -206,10 +206,14 @@ def get(self):
                     return self.redirect(create_url)
 
         if self.params.role == 'seek':
+            # The query_name parameter was previously called "query", and we
+            # continue to support it for third-parties that link directly to
+            # search results pages.
+            if self.params.query:
+                self.params.query_name = self.params.query
```
gimite: Would be better to avoid rewriting self.params directly, which is usually considered immutable, though the existing code was written in that way.

Also, nitpick, but `query_name` should probably win (because it's the canonical one) if both `query_name` and `query` are specified.

So it should be:
`query_name = self.params.query_name || self.params.query`
and use `query_name` instead of `self.params.query_name` below.

nworden: Done.

app/results.py
```diff
@@ -96,11 +96,12 @@ def search(self, query_dict):
                 results = indexing.search(self.repo,
                                           TextQuery(query_txt), MAX_RESULTS)
 
+        query_name = self.params.query_name or self.params.query
```
gimite: Can you document the reason to accept both parameter names here too? Same for below.

Or, may be better to create a helper function which returns:
`self.params.query_name or self.params.query`
instead of writing the same logic 3 times? Then you can also write the document only in the function.

Issue #212, opened 2015-11-11
Move app/*.py into subdirectories


gimite:
It looks there are too many files in app/ directory. How about creating subdirectories to categorize them into e.g., HTTP request handlers, datastore models and utility modules?

ee7klt:
@gimite I'm new to open source and would like to start contributing. After I setup as per "Getting Started", how do I get assigned this issue?

gimite:
Thanks for the interest to the contribution! You can just claim that you will work on it here, and send a pull request.

ee7klt:
I've had a look at this and I think I'm still not familiar enough with the code base to be able to do this in an expedient manner. I will release this issue and try one of the other ones.

ashutoshnarayan:
@gimite What are the names of the directories that needs to be created ? Where all in code should the change go apart from app/ directory ?

gimite:
Thanks for the interest to the contribution! 

Note that all files under app/ should still be inside app/ because only files under app/ are uploaded to App Engine. We should create subdirectories under app/.

Here is my idea:

* Keep directly under app/ - main.py, model.py, app.yaml, cron.yaml, index.yaml
* app/handlers/ - HTTP handlers (subclasses of utils.BaseHandler) e.g., view.py, start.py
* app/utils/ - Everything else e.g., utils.py, atom.py
* app/data/ - Data files (non-code) e.g., japanese_name_location_dict.txt

We may want to break them further but I don't have good idea immediately. @nworden What do you think?

One more idea is to break classes in model.py into each file and put it under app/model/. It can be done separately though.

ashutoshnarayan:
@gimite I have developed a python code to create directories as you have mentioned and it's placed under tools/create_subdirectory.py Please review it and let me know if it's fine or needs some changes. I have raised a PR #448 for the same.

gimite:
This task is currently blocked by #478 . See https://github.com/google/personfinder/pull/450#issuecomment-455962948 for details.

kshithijiyer:
I see that the code is already moved. Should this issue be closed now?

nworden:
Some of it's been moved (and I'm moving more as I go, with the migration to Django), but it's not done yet.

PR #450, opened 2018-11-07
Code for moving files from app/ into sub-directories

ashutoshnarayan:
Code for moving files from app/ into sub-directories leaving few. #212


gimite:
The conflict looks resolved, thanks! Can you commit the result of execution of the script as you did before?

app/handlers/README.md
```diff
@@ -0,0 +1,21 @@
+This directory contains resource files e.g., HTML templates and images.
```
gimite: This looks like README of somewhere else. Shouldn't be copied here.

ashutoshnarayan: Have moved the README file to correct path under resources

ashutoshnarayan:
@gimite Can you please let me know why tests are failing? Looks like some modules are not getting imported. It was working fine earlier though.

gimite:
> @gimite Can you please let me know why tests are failing? Looks like some modules are not getting imported. It was working fine earlier though.

It is because you moved files :) Can you update `import` lines as well as I mentioned earlier?

Other comments:

* Do not move files under app/vendors/:
    * app/vendors/six.py → app/handlers/six.py
* These files should be in app/handlers instead of app/utils:
    * amp_start.py, send_mail.py

ashutoshnarayan:
> > @gimite Can you please let me know why tests are failing? Looks like some modules are not getting imported. It was working fine earlier though.
> 
> It is because you moved files :) Can you update `import` lines as well as I mentioned earlier?

I have updated the import lines to include the path for modules. 3 tests have failed due to special character. Can you please take a look again at Build No. 326 ?
> 
> Other comments:
> 
> * Do not move files under app/vendors/:
>   
>   * app/vendors/six.py → app/handlers/six.py

Have fixed this in the code itself.

> * These files should be in app/handlers instead of app/utils:
>   
>   * amp_start.py, send_mail.py

Yes, they were present initially due to order of execution, after re-execution of code they are in place.

tests/load_test/load_test.py
```diff
@@ -13,6 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# added by Ashutosh Narayan
+
+import sys
+sys.path.append('../app/handlers/')
```
gimite: If you add it here, it would only fix tests, but not the actual app. Can you follow:
https://github.com/google/personfinder/wiki/GettingStarted#step-3-run-the-app-locally
to run it locally and access it? I believe it currently fails with your change. There are two ways to fix it:

* Add something similar to app/main.py as well.
* Update `import` lines in all source files to include the subdirectory name e.g.,
    `import const` → `from utils import const`

I would prefer the second option which is more explicit. But that would involve more work. Would be appreciated if you follow the second option, but I'm OK if you follow the first option fo now, then someone will later switch to the second option.

ashutoshnarayan: I have removed README.md file from handlers/ and also updated dictionary files.
Ran the tests locally after adding const module in all sources file but still tests failed. Can you review the new files with their changes which got committed and needs any further modification ? The tests are failing in import module itself both locally and during CI.

ashutoshnarayan: @gimite Have fixed the unit tests and they are passing now, server_tests are failing though.

app/handlers/config.py
```diff
@@ -19,6 +19,9 @@
 See here for usage examples:
 https://github.com/google/personfinder/wiki/DeveloperFaq
 """
+import sys
+sys.path.append('../app/handlers/')
```
gimite: Can you avoid appending to sys.path like here, and instead update `import` lines by adding subdirectories?

gimite:
Talked to @ashutoshnarayan directly too, but I asked him to put this on hold for now because it will likely cause a lot of conflict with ongoing #478. This pull request is a snapshot of what he has done. He or someone can continue from here once #478 is settled.

nworden:
This is a big pull request that affects just about everything, so I'm not sure there'll ever be a good time to get something like this in. Have you considered splitting it up into separate PRs? The data files could be moved to their own directory in one PR, and then the utils files in another, etc. That would reduce the conflicts that you'd have to deal with for each PR, and probably reduce conflicts overall too since each small PR could be done much faster than one really big PR.

nworden:
I think the data files and the handlers for admin pages could be moved without any conflicts. They might be a good thing to start with (if you have time for it, Ashutosh), and then we could find a way to coordinate the rest after that. I agree with Hiroshi that some of it is blocked on the UI work, since there's some refactoring of utility functions and handlers for public-facing pages going on.

gimite:
> I think the data files and the handlers for admin pages could be moved without any conflicts.

Sounds good for data files. It is slightly tricky to move just admin page handlers without moving other handlers because you need to introduce exception to this logic:
https://github.com/google/personfinder/blob/master/app/main.py#L81

@ashutoshnarayan Maybe can you start by just moving data files if you have time?:
* app/chinese_family_name_dict.txt → app/data/chinese_family_name_dict.txt
* app/japanese_name_location_dict.txt → app/data/japanese_name_location_dict.txt

ashutoshnarayan:
I agree with @nworden 
@gimite I already have * app/chinese_family_name_dict.txt and * app/japanese_name_location_dict.txt moved to app/data directory when the code is executed. What other changes are you requesting for ?

gimite:
> I agree with @nworden
> @gimite I already have app/chinese_family_name_dict.txt and app/japanese_name_location_dict.txt moved to app/data directory when the code is executed. What other changes are you requesting for ?

The request here is to create a pull request (maybe a new one) which **only** move app/chinese_family_name_dict.txt and app/japanese_name_location_dict.txt to app/data but not other files. Since we want to move other files later, not now.

gimite:
Closing this one for now. Can you create a new pull request with the changes suggested above once ready?

ashutoshnarayan:
@gimite sure, will create a new PR once the requested changes are ready.

PR #452, opened 2018-11-08
Fix #425: Add web UI to download records in CSV format

### LINKED ISSUE #425: Allow API users to download records directly from the web site
We allow **bulk_read** API users to download records. But it currently requires users to install Python and Git, clone Person Finder Github repository and use a command line tool tools/download_feed.py. This can be hard, considering API users are not always tech savvy.

It would be nice if we allow users to download records as CSV directly from the web site, though still requiring the **bulk_read** API key.

Implementation is a bit tricky because fetching all records in the repository can take longer than AppEngine request timeout. download_feed.py issues multiple HTTP requests. Maybe first create a CSV file on Blobstore using a task queue, and then let the user download the file from Blobstore?

gimite:


nworden:
It can be in a separate PR, but it seems like it'd be worth adding tests for some of the new functionality.

gimite:
Sure. It is a bit tricky to test (I probably need to stub out CloudStorage class), so I'm planning to do it later separately.

PR #453, opened 2018-11-08
Avoid translating messages in app/vendors

gimite:


tools/update_messages.py
```diff
@@ -156,7 +156,8 @@ def django_admin(*args):
     old_msgids = set(message.id for message in old_template)
 
     # Run makemessages to update the English .po file.
```
nworden: Could you add to the comment here to say that 'vendors' is what's being excluded (and/or pass the string as a keyword/named argument)? If I were just reading this line without context, I'd probably think it was only including the vendors directory.

gimite: Good point, done.

PR #459, opened 2018-12-04
Use print() function in both Python 2 and Python 3

cclauss:
These changes are compatible with both Python 2 and Python 3.

__print()__ is a function in Python 3 so these changes should ease migration to Python 3 without breaking compatibility with Python 2.  This PR was created by running the command __futurize -f libfuturize.fixes.fix_print_with_import -w .__ as discussed at: https://python-future.org

### DIFF
--- a/app/vendors/rsa/key.py
+++ b/app/vendors/rsa/key.py
@@ -32,6 +32,7 @@ of pyasn1.
     or unauthenticated source.
 
 """
+from __future__ import print_function
 
 import logging
 import warnings
--- a/app/vendors/rsa/key.py
+++ b/app/vendors/rsa/key.py
@@ -32,7 +32,6 @@ of pyasn1.
     or unauthenticated source.
 
 """
-from __future__ import print_function
 
 import logging
 import warnings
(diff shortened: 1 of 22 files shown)


app/vendors/rsa/key.py
```diff
@@ -32,6 +32,7 @@
     or unauthenticated source.
 
 """
+from __future__ import print_function
```
gimite: Can you revert your changes inside app/vendors? Files in this directory are snapshot of third party libraries installed via PIP, so we don't own the files.

tools/babel/messages/extract.py
```diff
@@ -20,6 +20,7 @@
 The main entry points into the extraction functionality are the functions
 `extract_from_dir` and `extract_from_file`.
 """
+from __future__ import print_function
```
gimite: Can you also revert your changes inside tools/babel? Looks it's also third party source code.

cclauss:
Done via __git checkout master app/vendors/rsa/key.py app/vendors/rsa/pkcs1.py app/vendors/rsa/pkcs1_v2.py  app/vendors/rsa/prime.py tools/babel/messages/extract.py tools/babel/messages/frontend.py__

gimite:
Thanks! Merged.

PR #464, opened 2018-12-06
Fix #451: Add support for CSV download with sensitive fields in the web UI

### LINKED ISSUE #451: Add support for CSV download with sensitive fields in the web UI
Currently the CSV download from the web UI always removes sensitive fields. Should allow downloading records with sensitive fields when the API key has full_bulk_read access.

To implement it, DumpCSV cron job should generate two versions of CSV files, one with sensitive fields and one without them.

gimite:
Generates both "full" version (with sensitive fields) and "basic" version (without sensitive fields) in the cron job, and serves one of them based on the API key permission.

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -337,7 +337,7 @@ class Import(utils.BaseHandler):
         if self.auth.full_read_permission:
             object_name = self.config.latest_full_csv_object_name
         else:
-            object_name = self.config.latest_basic_csv_object_name
+            object_name = self.config.latest_filtered_csv_object_name
         if object_name:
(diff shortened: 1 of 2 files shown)


gimite:
This word is user visible as exported CSV file name. So I'm slightly concerned that "nonsensitive" gives a wrong impression because the data is still sensitive (containing a list of person names for example). How about "filtered", or maybe simply "default"?

nworden:
"filtered" SGTM.

gimite:
> "filtered" SGTM.

Done.

PR #465, opened 2019-01-11
Start of a README file for the React UI code.

nworden:


ui/README.md
```diff
@@ -0,0 +1,64 @@
+# UI
```
gimite: Is it safe to merge all your changes to the master branch? Doesn't it break the existing implementation?

ui/README.md
```diff
@@ -0,0 +1,64 @@
...
+The easiest way to develop the UI is to run two local servers:
+
+1. Run the Django (backend) server as usual, at port 8000.
+2. Run the UI with `tools/ui run`. It will talk to the backend at port 8000.
+
+Changes you make to the JS and CSS will take effect when you refresh the page.
+If you make other changes (e.g., to install a new module or modify the Webpack
+config) you'll likely need to restart the server.
+
+Currently, it is necessary to change `USE_REACT_UI` in `app/main.py` to `True`.
```
gimite: Can you use config to configure this?

ui/README.md
```diff
@@ -0,0 +1,64 @@
...
+Please be careful not to commit this change.
+
+## Deployment
+
+You can run `tools/ui buildtoae` to compile the React app into language-specific
```
gimite: What do you mean by language-specific? Is it natural language or programming language?

nworden: Natural language. I changed the doc to clarify it.

ui/README.md
nworden: For the ui directory code: it's certainly safe to merge the ui directory in, because there's no direct connection between that directory and the Django app's directory. You have to compile the React app and move it (either manually or with `tools/ui buildtoae`) into /app for the /ui code to even end up on AppEngine.
For the Django code: so far I haven't made any changes to the Django code that should affect the default behavior (and none of that's out for review yet).

ui/README.md
nworden: I could, but I think there's a small advantage to doing it with a constant like this: when we want to make the switch for real in prod, we can upload a new version to AE with the boolean flipped and then use AppEngine versioning to put a small percentage of traffic on the new version at first, as a canary.

ui/README.md
gimite: Thanks, sounds good. I guess your plan for Django code changes is to still submit them into the master branch and guard them with USE_REACT_UI so that it doesn't affect production until it's ready?

ui/README.md
gimite: Understood. It's up to you, but I found it's often annoying when I always need to flip a constant value locally and make sure to revert it before submitting it.

Maybe define useReactUI() function which is backed by a config for now, and switch it to a constant once it's ready to ramp up? We can discuss later though.

ui/README.md
nworden: Yep. I'm going to do some refactoring along the way (to avoid code duplication), but all the code that changes behavior will be flag-guarded.

PR #467, opened 2019-01-11
Translation files for development.

nworden:
I don't have a pipeline for generating these files for real yet, but
it's useful to have a couple language files just to be sure the
language/translation code is working. The Spanish file just has two
messages as it's just proof-of-concept (for anything else it'll use the
default message). For English, it doesn't really need anything in the
file since it's the default language anyway (at some point I'd like to
explore whether it's possible to avoid including the default messages in
the minified JS, but that doesn't seem like an immediate requirement).

### DIFF
--- a/ui/src/translations/en.js
+++ b/ui/src/translations/en.js
@@ -1,0 +1,2 @@
+module.exports = {
+};
--- a/ui/src/translations/en.js
+++ b/ui/src/translations/en.js
@@ -1,2 +1,7 @@
+/**
+ * The English translation file has to exist to avoid complicating things for
+ * index.js, but it doesn't need to actually have any translations, because that
+ * would just be redundant with the default messages (which are all in English).
+ */
 module.exports = {
 };
--- a/ui/src/translations/es.js
+++ b/ui/src/translations/es.js
@@ -1,3 +1,6 @@
+/**
+ * This file is just for temporary testing use.
+ */
 module.exports = {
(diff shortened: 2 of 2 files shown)


ui/src/translations/en.js
```diff
@@ -0,0 +1,2 @@
+module.exports = {
```
gimite: Can you add a comment explaining why it's empty?

Also, it would be helpful if there's ui/src/translations/README.md which explains the format of files in the directory. It can be done later though.

ui/src/translations/es.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  'GlobalHome.howDoesItWork': '¿Como funciona?',
```
gimite: Add a comment saying these are temporary translations for testing.

ui/src/translations/en.js
gimite: Also, I believe you need copyright & license header for every file.

ui/src/translations/es.js
nworden: Done.

ui/src/translations/en.js
nworden: Added a comment about why it's empty.
About adding docs: noted. I'll do that when I write a script to get messages to TC.
About license/copyright: see thread on pull request #466.

PR #468, opened 2019-01-11
Index HTML file for local development.

nworden:
In prod, the Django server will serve the base HTML to clients.
However, for local development, it's much more convenient to use npm to
run the React app (since it supports edit-refresh). This HTML file is
what's served by npm for local development.


ui/static/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
```
gimite: I believe you need copyright & license header for every file.

Also, can you add a comment saying this file is used only for development? May be even clearer if it's named e.g., dev_index.html.

nworden: Renamed to dev_index.html.
See PR #466 about licenses.

nworden: Hm, actually webpack seems to expect the file to be called index.html (they have an "index" setting that sounded promising, but it doesn't do what we need). Added a comment at the top instead.

PR #469, opened 2019-01-11
Router (and index.js) for React app.

nworden:


ui/src/index.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
```
gimite: Be consistent whether to use single quotes or double quotes.

ui/src/index.js
```diff
@@ -0,0 +1,15 @@
...
+import {addLocaleData, IntlProvider} from 'react-intl';
+
+import App from "./App.js";
+
+addLocaleData(require('react-intl/locale-data/' + BUNDLE_LANGUAGE));
+
+ReactDOM.render(
+    <IntlProvider
+        locale={BUNDLE_LANGUAGE}
+        messages={require('./translations/' + BUNDLE_LANGUAGE + '.js')}>
```
gimite: Optional: This may look better:

```js
`./translations/${BUNDLE_LANGUAGE}.js`
```

ui/src/App.js
```diff
@@ -0,0 +1,25 @@
...
+import RepoHome from "./pages/RepoHome.js";
+import Results from "./pages/Results.js";
+import View from "./pages/View.js";
+
+import './css/all.scss';
+
+const App = () => (
+  <BrowserRouter>
+    <div>
+      <Route exact path="/" component={GlobalHome} />
```
gimite: You don't need to support those initially, but FYI. May be worth leaving TODO. We should support these URLs for backward compatibility:

* /global/home.html (had been the canonical URL of the global home page for long)
* /:repoId/ (with a trailing slash)

ui/src/index.js
nworden: The Google JS guide prefers single so I went with that.

ui/src/index.js
nworden: Done.

ui/src/App.js
nworden: Added a todo for global/home.html. Trailing slashes already work; I guess React Router just handles that for you.

ui/src/App.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
```
gimite: Change double quotes in this file into single quotes as well.

ui/src/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import {addLocaleData, IntlProvider} from 'react-intl';
+
+import App from './App.js';
+
+addLocaleData(require('react-intl/locale-data/' + BUNDLE_LANGUAGE));
```
gimite: Optional: You can write
```js 
    `react-intl/locale-data/${BUNDLE_LANGUAGE}`
```
here too which may look more consistent with one below.

ui/src/App.js
```diff
@@ -0,0 +1,28 @@
...
+import RepoHome from "./pages/RepoHome.js";
+import Results from "./pages/Results.js";
+import View from "./pages/View.js";
+
+import './css/all.scss';
+
+const App = () => (
+  <BrowserRouter>
+    <div>
+      /**
```
gimite: Use a normal comment // ... instead of JSDoc /** ... */ for this.

ui/src/App.js
nworden: Done.

ui/src/App.js
nworden: Done.

ui/src/index.js
nworden: Done.

PR #470, opened 2019-01-15
Master CSS file and Material value definitions.

nworden:
Master CSS file and Material value definitions.


ui/src/css/_material-values.scss
```diff
@@ -0,0 +1,6 @@
+/**
```
gimite: Any reason why this file name starts with an underscore? Is it a convention of SCSS?

nworden: Oh, sorry, I'll add that to the README. Yes, it's a Sass convention (a little more than a convention actually -- it's a signal to the compiler). Files with underscores are files that are just meant for inclusion in another file and shouldn't produce their own CSS file. I don't think it'd really be a problem if we left off the underscore, but we'd end up with a bunch of extra CSS files that we don't need for anything, so I gave everything but all.scss an underscore.

gimite: Thanks for explanation!

PR #471, opened 2019-01-15
Footer React component and CSS.

nworden:


ui/src/components/Footer.js
```diff
@@ -0,0 +1,27 @@
+import React, {Component} from 'react';
+import {FormattedMessage, defineMessages, injectIntl} from 'react-intl';
+
+const messages = defineMessages({
+  disclaimerText: {
+    id: 'Footer.disclaimerText',
+    defaultMessage: ('PLEASE NOTE: all data entered is available to the public '
```
gimite: I believe "all" should be capitalized (as in the current production)?

ui/src/components/Footer.js
```diff
@@ -0,0 +1,27 @@
...
+        + 'and usable by anyone. Google does not review or verify the accuracy '
+        + 'of this data. Google may share the data with public and private '
+        + 'organizations participating in disaster response efforts.'),
+    description: 'A disclaimer shown at the footer of the page.',
+  },
+});
+
+class Footer extends Component {
```
gimite: Optional: This can be a simple function instead of a class as you did for App component:

```jsx
const Footer = () => (
  <div>...</div>
);
```

ui/src/components/Footer.js
nworden: Ok, done. English in general doesn't have a grammar rule about that afaik, but most style guides seem to prefer capitalizing it in this case.

ui/src/components/Footer.js
nworden: Done.

ui/src/components/Footer.js
```diff
@@ -0,0 +1,23 @@
...
+  },
+});
+
+const Footer = () => (
+  <div className="footer">
+    <p className="mdc-typography--body1">
+      <FormattedMessage {...messages.disclaimerText} />
+    </p>
+  </div>
+}
```
gimite: This must be `);` instead.

nworden: Done.

nworden:
Just FYI, I moved _Footer.scss into a components subdirectory of the css directory. I'd done a little restructuring and overlooked this pull request when I was updating things (the master CSS file PR #470 already expects that it'll be in /css/components).

PR #473, opened 2019-01-15
Global homepage React component and CSS.

nworden:
Global homepage React component and CSS.


ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+    };
+  }
+
+  componentDidMount() {
+    fetch("/global/d/repoinfo")
+      .then(res => res.json())
+      .then(
+        (repos) => {
+          for (var i = 0; i < repos.length; i++) {
+            repos[i]['displayCat'] = i % 5;
```
gimite: Optional: May be better to spell out "Cat" (category?)

You can also alternatively write this as:
```js
repos[i].displayCat = i % 5;
```

gimite: Add comment or make it a constant to explain why it's 5.

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+          <img src="/static/icons/maticon_person_pin.svg" />
+          <FormattedHTMLMessage {...messages.productNameWithEmphasis} />
+        </h1>
+        <p className="mdc-typography--body1 globalhome-headerdesc">
+          <FormattedMessage {...messages.tagline} />
+        </p>
+        <Button
+          className="pf-button-secondary globalhome-howsitworkbutton"
+          raised
+          onClick={() => console.log('clicked!')}
```
gimite: Add TODO to implement it. Same for below.

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+          </CardActionButtons>
+        </CardActions>
+      </Card>
+    );
+  }
+
+  renderThirdPartyHelpCards() {
+    return (
+      <div className="globalhome-thirdpartywrapper">
+        <div className="endbars-headerline-wrapper" dir="ltr">
```
gimite: Why is it hard-coded to "ltr"?

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+          </Row>
+        </Grid>
+      </div>
+    );
+  }
+
+  renderRepoList() {
+    var cells = this.state.repos.map(repo => (
+      <Cell key={repo.repoId}>
+        <RepoCard repoInfo={repo} />
```
gimite: `repoInfo` sounds redundant. Why not just `repo`?

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+      </Cell>
+    ));
+    return <Grid><Row>{cells}</Row></Grid>;
+  }
+
+  render() {
+    if (this.state.error) {
+      return <div>An error occurred</div>
```
gimite: Maybe add a TODO to give more useful (and i18ned) error message?

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+    }
+    if (!this.state.isLoaded) {
+      return <LoadingIndicator />;
+    }
+    return (
+      <div className="globalhome-wrapper">
+        {this.renderHeader()}
+        {this.renderRepoList()}
+        {this.renderThirdPartyHelpCards()}
+        <Footer />
```
gimite: Maybe Footer should be added to App component instead because I guess it should be added to all pages?

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+      </div>
+    );
+  }
+}
+
+/**
+ * RepoCard is a component for a card in the repo list.
+ */
+
+const repoCardMessages = defineMessages({
```
gimite: Any reason to define this separately from `messages`?

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+  repoRecordCount: {
+    id: 'RepoCard.repoRecordCount',
+    defaultMessage: 'Tracking {recordCount} records.',
```
gimite: Nit: Looks it shouldn't end with a period per the mock.

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+});
+
+class RepoCardImpl extends Component {
+  constructor(props) {
+    super(props);
+    this.goTo = this.goTo.bind(this);
+  }
+
+  goTo() {
+    this.props.history.push("/" + this.props.repoInfo.repoId);
```
gimite: Can be done later, but don't we need special handling for a case where the app root path is not "/", as in our production (https://google.org/personfinder/)?

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+            {...repoCardMessages.repoRecordCount }
+            values={{
+              "recordCount": this.props.repoInfo.recordCount
+            }} />
+        </p>
+      );
+    }
+    return (
+      <Card className="repocard">
+        <CardPrimaryContent className="repocard-content" onClick={this.goTo}>
```
gimite: Just curious, do you know if there's a way to generate A tag with the target path in addition to the JS click handler? A tag has a nice property like the user can right-click it and open the target page in a new tab. It should probably have the click handler as well to allow page transition via pushState though. We can explore it later though.

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+          <div className={repocardImageClassName}>
+            <p className="mdc-typography--body1">
+              {this.props.repoInfo.title[0]}
```
gimite: `title[0]` doesn't handle UTF-16 surrogate pairs correctly. Seems like `Array.from(title)[0]` should work:
https://stackoverflow.com/a/42596897/4339963

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+          });
+        }
+      );
+  }
+
+  renderHeader() {
+    return (
+      <div>
+        <h1 className="mdc-typography--headline1">
+          <img src="/static/icons/maticon_person_pin.svg" />
```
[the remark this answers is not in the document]
nworden: I think it should be fine, since afaik we just need really basic support for SVG:
https://caniuse.com/#search=svg

ui/src/pages/GlobalHome.js
nworden: Yeah, I am going to need to do something about that. I haven't dug too far into what's involved, but it doesn't look like it'll be too bad (React/react-router do have support for that kind of thing).

ui/src/pages/GlobalHome.js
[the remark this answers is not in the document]
gimite: Have you decided which browsers we should support? Looks like IE8 and Android Browser 4.3 and below doesn't support SVG (at all or partially). I wonder if we want to support them.

ui/src/css/pages/_GlobalHome.scss
```diff
@@ -0,0 +1,124 @@
...
+
+.globalhome-howsitworkbutton {
+  margin: 40px 0;
+}
+
+.globalhome-thirdpartywrapper .endbars-headerline-wrapper {
+  margin-bottom: 8px;
+}
+
+.globalhome-thirdpartywrapper {
```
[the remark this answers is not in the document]
nworden: Done.

ui/src/pages/GlobalHome.js
nworden: No, I guess not really.

ui/src/pages/GlobalHome.js
nworden: More discussion on PR #472, but in short it's to make the CSS work without using CSS properties that aren't widely supported.

ui/src/pages/GlobalHome.js
nworden: I don't, I'll look into it. It might be pretty tricky though, since we're using Material buttons and don't have control over their internals.

ui/src/pages/GlobalHome.js
nworden: I'm not sure yet; I don't know what the static-y pages (like the how-does-it-work page and ToS) are going to look. If we can't put Footer in App though, we could alternatively have a base RepoPage component (that the others subclass) that includes both the Footer and the RepoHeader.

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,293 @@
...
+    super(props);
+    this.state = {
+      isLoaded: false,
+      error: null,
+      repos: []
+    };
+  }
+
+  componentDidMount() {
+    fetch("/global/d/repoinfo")
```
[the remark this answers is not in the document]
nworden: It looks like it'd be possible but probably more trouble than it's worth, considering that we don't need to make too many API calls (just once per page load and form submission). You can't do it without babel-polyfill ([SO question about it here](https://stackoverflow.com/questions/33527653/babel-6-regeneratorruntime-is-not-defined)), and I'd rather not add another dependency just to make the calls somewhat cleaner.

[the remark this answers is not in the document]
gimite: Sounds good. It may be nice if we document this kind of general decisions (e.g., which ES6 features to use) somewhere.

ui/src/pages/GlobalHome.js
```diff
@@ -0,0 +1,314 @@
...
+    description: 'A one-line summary of what the product is for.',
+  },
+  youCanHelp: {
+    id: 'GlobalHomePage.youCanHelp',
+    defaultMessage: 'You can help',
+    description: ('Header on a section describing how people can help.'),
+  }
+});
+
+const NUMBER_OF_DISPLAY_CATEGORY_COLORS = 5;
```
gimite: Can you also document that this number should match one in the CSS file?

PR #474, opened 2019-01-15
Simple loading indicator component.

nworden:
We might want to improve on this later, but for now I just want to have a simple component that we can use in all the pages, so it'll be easy to replace them all with a spinner or whatever we want later.

### DIFF
--- a/ui/src/components/LoadingIndicator.js
+++ b/ui/src/components/LoadingIndicator.js
@@ -1,0 +1,18 @@
+import React, {Component} from 'react';
+import {FormattedMessage, defineMessages, injectIntl} from 'react-intl';
+
+const messages = defineMessages({
+  loading: {
+    id: 'LoadingIndicator.loading',
+    defaultMessage: 'Loading...',
+    description: 'A message indicating to the user that the page is loading.',
+  },
+});
+
+class LoadingIndicator extends Component {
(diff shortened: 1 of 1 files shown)


ui/src/components/LoadingIndicator.js
```diff
@@ -0,0 +1,18 @@
+import React, {Component} from 'react';
```
gimite: Add copyright & license header.

ui/src/components/LoadingIndicator.js
```diff
@@ -0,0 +1,18 @@
...
+  loading: {
+    id: 'LoadingIndicator.loading',
+    defaultMessage: 'Loading...',
+    description: 'A message indicating to the user that the page is loading.',
+  },
+});
+
+class LoadingIndicator extends Component {
+  render() {
+    return <FormattedMessage {...messages.loading} />
```
gimite: Nit: Append a semicolon, I guess?

nworden: I changed it to the just-a-function version (like the Footer and App components).

ui/src/components/LoadingIndicator.js
nworden: Done.

PR #480, opened 2019-01-21
.gitignore file for ui directory.

nworden:


ui/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+
+# testing
+/coverage
+
+# build
+/build
+/dist
+
+# local logs
+npm-debug.log*
```
gimite: Just curious, why is it `.log*`? Can something follow `.log`?

nworden: I guess so. These filenames were in a few different recommended gitignore files I found, and they consistently included the asterisk at the end.

PR #481, opened 2019-01-21
Component for repo header.

nworden:


ui/src/css/components/_RepoHeader.scss
```diff
@@ -0,0 +1,27 @@
...
+  float: left;
+}
+
+#repoheader p {
+  margin: 0;
+}
+
+#repoheader p.mdc-typography--subtitle1 {
+  color: #212121;
+  margin: 3px 0 8px;
```
nworden: This might change -- I'll ask Travis about the sizes and spacing along with the other stuff (using the subtitle class for the header feels a little odd, but it's the only thing that seems to be the right text size).

ui/src/components/RepoHeader.js
```diff
@@ -0,0 +1,52 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, {Component} from 'react';
+import {FormattedHTMLMessage, defineMessages, injectIntl} from 'react-intl';
+import {Link} from "react-router-dom";
```
gimite: Use single quotes.

ui/src/components/RepoHeader.js
```diff
@@ -0,0 +1,52 @@
...
+
+const messages = defineMessages({
+  productName: {
+    id: 'RepoHeader.productName',
+    defaultMessage: 'Google Person Finder',
```
gimite: Can be considered later but this name should be configurable. Should be in a different name if a third party runs Person Finder.

ui/src/components/RepoHeader.js
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      <div id='repoheader'>
+        <div id='repoheader-backbutton'>
+          <Link to={'/' + this.props.repoInfo.repoId}>
+            <div><img src='/static/icons/maticon_arrow_back.svg' /></div>
+          </Link>
+        </div>
+        <div id='repoheader-info'>
+          <p className='mdc-typography--subtitle1'>
+            <FormattedHTMLMessage {...messages.productName} />
```
gimite: Why isn't it just FormattedMessage?

ui/src/components/RepoHeader.js
```diff
@@ -0,0 +1,52 @@
...
+        + 'people reconnect with friends/family after a disaster.'),
+  },
+});
+
+class RepoHeader extends Component {
+  render() {
+    return (
+      <div id='repoheader'>
+        <div id='repoheader-backbutton'>
+          <Link to={'/' + this.props.repoInfo.repoId}>
```
gimite: Can be considered later, but does it always go back to the repo home page? In the current UI, it can e.g., go back from the person record page to the search result.

ui/src/css/components/_RepoHeader.scss
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+
+#repoheader {
+  border-bottom: 1px solid #dadce0;
+  overflow: auto;
+}
+
+#repoheader-backbutton {
+  float: left;
```
gimite: Optional: Flexbox may be a better alternative (i.e., with less tricky behavior) to float for this kind of thing.

ui/src/components/RepoHeader.js
nworden: Drats, I thought I got 'em all.

ui/src/components/RepoHeader.js
nworden: Nice catch. I think it was just left over from an earlier version of mocks that had part of it in bold font.

ui/src/components/RepoHeader.js
nworden: That's a good point. I'll slap a TODO on this for now and do some thinking about how to handle white-labeling more generally, including in the React app.

ui/src/css/components/_RepoHeader.scss
nworden: Thanks. I'll consider it when we get the details of the spacing and stuff here nailed down.

ui/src/components/RepoHeader.js
```diff
@@ -0,0 +1,54 @@
...
+  // than Google
+  productName: {
+    id: 'RepoHeader.productName',
+    defaultMessage: 'Google Person Finder',
+    description: ('Name of the product. Person Finder is a tool that helps '
+        + 'people reconnect with friends/family after a disaster.'),
+  },
+});
+
+class RepoHeader extends Component {
```
gimite: Optional: This can be a simple function instead.

PR #484, opened 2019-01-21
Component for general-purpose notched outline.

nworden:
The mocks call for a notched outline around sections of the create form.
However, the Material libraries only seem to have support for notched
outlines on input components, so, at least for now, I think we need our
own component for the section outlines.

### DIFF
--- a/ui/src/components/PFNotchedOutline.js
+++ b/ui/src/components/PFNotchedOutline.js
@@ -1,0 +1,24 @@
+import React, {Component} from 'react';
+
+/**
+ * A component for wrapping something in a notched outline.
+ *
+ * We may be able to get rid of this at some point and use a component built by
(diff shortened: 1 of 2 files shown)


ui/src/components/PFNotchedOutline.js
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * A component for wrapping something in a notched outline.
+ *
+ * We may be able to get rid of this at some point and use a component built by
+ * the Material team, but a) the documentation for the React NotchedOutline is
+ * pretty sparse and b) I'm not sure Material supports using their
+ * NotchedOutline with anything other than TextField and Select components.
+ */
+class PFNotchedOutline extends Component {
```
gimite: Optional: This can be a simple function.

nworden: Done.

PR #486, opened 2019-01-22
Skeleton component for view page.

nworden:
I haven't actually implemented the rendering of this page. This PR is just so that the component exists in the master branch, to get that code closer to being able to compile.

### DIFF
--- a/ui/src/Utils.js
+++ b/ui/src/Utils.js
@@ -19,6 +19,9 @@
  */
 class Utils {
 
+  /*
+   * Get the value of the given URL param from the window.
+   */
   static getURLParam(paramName) {
     return new URL(window.location.href).searchParams.get(paramName);
   }
--- a/ui/src/Utils.js
+++ b/ui/src/Utils.js
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * A class for utility functions.
+ */
+class Utils {
+
+  static getURLParam(paramName) {
+    return new URL(window.location.href).searchParams.get(paramName);
+  }
+}
+
+export default Utils;
--- a/ui/src/pages/View.js
+++ b/ui/src/pages/View.js
@@ -18,6 +18,7 @@ import React, {Component} from 'react';
 import {FormattedMessage, defineMessages, injectIntl} from 'react-intl';
 
 import LoadingIndicator from './../components/LoadingIndicator.js';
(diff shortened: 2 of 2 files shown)


ui/src/pages/View.js
```diff
@@ -0,0 +1,70 @@
...
+    this.state = {
+      isLoaded: false,
+      error: null,
+      repo: null,
+      person: null
+    };
+  }
+
+  componentDidMount() {
+    const personId = new URL(window.location.href).searchParams.get('id');
```
gimite: Optional: Maybe define a utility function to do this (extract a URL parameter of the current page) because you would need this in other places too?

ui/src/pages/View.js
```diff
@@ -0,0 +1,70 @@
...
+    // TODO(nworden): consider if we could have a global cache of repo info to
+    // avoid calling for it on each page load
+    var apiCalls = [
+        '/' + this.props.match.params.repoId + '/d/repoinfo',
+        '/' + this.props.match.params.repoId + '/d/personinfo?id=' + personId,
```
gimite: Probably safer to apply URL encode to this personId to avoid generating any unexpected URL here?

ui/src/pages/View.js
```diff
@@ -0,0 +1,70 @@
...
+      repo: null,
+      person: null
+    };
+  }
+
+  componentDidMount() {
+    const personId = new URL(window.location.href).searchParams.get('id');
+    // TODO(nworden): consider if we could have a global cache of repo info to
+    // avoid calling for it on each page load
+    var apiCalls = [
```
gimite: Should be const.

gimite: Optional: Better to call it apiPaths or apiURLs? "Call" sounds vague.

ui/src/pages/View.js
```diff
@@ -0,0 +1,70 @@
...
+        '/' + this.props.match.params.repoId + '/d/repoinfo',
+        '/' + this.props.match.params.repoId + '/d/personinfo?id=' + personId,
+        ];
+    Promise.all(apiCalls.map(url => fetch(url)))
+        .then(res => Promise.all(res.map(r => r.json())))
+        .then(
+          (res) => {
```
gimite: Optional: You can write:

```js
([repo, person]) => {
  this.setState({isLoaded: true, repo, person});
},
```

ui/src/Utils.js
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * A class for utility functions.
+ */
+class Utils {
+
+  static getURLParam(paramName) {
```
gimite: Add document comment for this method.

ui/src/pages/View.js
```diff
@@ -0,0 +1,72 @@
...
+        '/' + this.props.match.params.repoId + '/d/person?id='
+            + encodeURIComponent(personId),
+        ];
+    Promise.all(apiURLs.map(url => fetch(url)))
+        .then(res => Promise.all(res.map(r => r.json())))
+        .then(
+          ([repo, person]) => {
+            this.setState({
+              isLoaded: true,
+              repo: repo,
```
gimite: Optional: You can simplify `repo: repo` to `repo`.

nworden: Oh, interesting. I think it's more readable this way though, so I'm going to leave it as it is.

gimite: Yeah it is up to you, but I believe it's a commonly used feature of ES6, especially when all keys matches variable names like `{error: error}` to `{error}`.

PR #487, opened 2019-01-22
Component for search results page.

nworden:
There's a little left to do, namely the last-update timestamp line, but
most of the work for that is going to be on the backend.

### DIFF
--- a/ui/src/pages/Results.js
+++ b/ui/src/pages/Results.js
@@ -49,6 +49,7 @@ class Results extends Component {
       repo: null,
       results: null
     };
+    this.repoId = this.props.match.params.repoId;
     this.goToAdd = this.goToAdd.bind(this);
     this.loadResults = this.loadResults.bind(this);
     this.handleSearch = this.handleSearch.bind(this);
@@ -56,24 +57,24 @@ class Results extends Component {
 
   goToAdd() {
     this.props.history.push({
-          pathname: '/' + this.props.match.params.repoId + '/create',
+          pathname: '/' + this.repoId + '/create',
       });
   }
 
   loadResults() {
     const query = new URL(window.location.href).searchParams.get('query_name');
-    var apiCalls = [
-        '/' + this.props.match.params.repoId + '/d/repoinfo',
-        '/' + this.props.match.params.repoId + '/d/results?query=' + query,
+    const apiURLs = [
+        '/' + this.repoId + '/d/repoinfo',
+        '/' + this.repoId + '/d/results?query=' + encodeURIComponent(query),
         ];
-    Promise.all(apiCalls.map(url => fetch(url)))
+    Promise.all(apiURLs.map(url => fetch(url)))
         .then(res => Promise.all(res.map(r => r.json())))
         .then(
-          (res) => {
+          ([repo, results]) => {
             this.setState({
               isLoaded: true,
-              repo: res[0],
-              results: res[1]
+              repo: repo,
+              results: results,
             });
           },
           (error) => {
@@ -89,8 +90,8 @@ class Results extends Component {
   handleSearch(query) {
     this.setState({isLoaded: false});
     this.props.history.push({
-        pathname: '/' + this.props.match.params.repoId + '/results',
-        search: '?query_name=' + query,
+        pathname: '/' + this.repoId + '/results',
+        search: '?query_name=' + encodeURIComponent(query),
       });
     this.loadResults();
   }
@@ -103,7 +104,7 @@ class Results extends Component {
         </p>
       );
     }
-    var results = this.state.results.map(result => (
+    const results = this.state.results.map(result => (
       <SearchResult
         key={result.personId}
         repo={this.state.repo}
@@ -120,7 +121,8 @@ class Results extends Component {
           onClick={this.goToAdd}
           icon={<img src='/static/icons/maticon_add.svg' />}
           textLabel={this.props.intl.formatMessage(
-              messages.createNewRecord)} />
+              messages.createNewRecord)}
+      />
     );
   }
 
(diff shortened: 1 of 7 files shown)


ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+      results: null
+    };
+    this.goToAdd = this.goToAdd.bind(this);
+    this.loadResults = this.loadResults.bind(this);
+    this.handleSearch = this.handleSearch.bind(this);
+  }
+
+  goToAdd() {
+    this.props.history.push({
+          pathname: '/' + this.props.match.params.repoId + '/create',
```
gimite: How about factoring out `this.props.match.params.repoId` as a method or a field because it's used repeatedly?

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+      });
+  }
+
+  loadResults() {
+    const query = new URL(window.location.href).searchParams.get('query_name');
+    var apiCalls = [
+        '/' + this.props.match.params.repoId + '/d/repoinfo',
+        '/' + this.props.match.params.repoId + '/d/results?query=' + query,
```
gimite: Apply URL encoding to `query`.

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+  goToAdd() {
+    this.props.history.push({
+          pathname: '/' + this.props.match.params.repoId + '/create',
+      });
+  }
+
+  loadResults() {
+    const query = new URL(window.location.href).searchParams.get('query_name');
+    var apiCalls = [
+        '/' + this.props.match.params.repoId + '/d/repoinfo',
```
gimite: Optional: Maybe rename `repoinfo` to `repo` in the URL to be consistent with variable names we use to hold its result?

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+
+  goToAdd() {
+    this.props.history.push({
+          pathname: '/' + this.props.match.params.repoId + '/create',
+      });
+  }
+
+  loadResults() {
+    const query = new URL(window.location.href).searchParams.get('query_name');
+    var apiCalls = [
```
gimite: var -> const

gimite: apiCalls -> apiURLs or apiPaths?

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+        '/' + this.props.match.params.repoId + '/d/repoinfo',
+        '/' + this.props.match.params.repoId + '/d/results?query=' + query,
+        ];
+    Promise.all(apiCalls.map(url => fetch(url)))
+        .then(res => Promise.all(res.map(r => r.json())))
+        .then(
+          (res) => {
```
gimite: Optionally:

```js
([repo, results]) => {
  this.setState({isLoaded: true, repo, results});
},
```

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+
+  componentDidMount() {
+    this.loadResults();
+  }
+
+  handleSearch(query) {
+    this.setState({isLoaded: false});
+    this.props.history.push({
+        pathname: '/' + this.props.match.params.repoId + '/results',
+        search: '?query_name=' + query,
```
gimite: Apply URL encoding to `query`.

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+      });
+    this.loadResults();
```
gimite: I just imagined react-router handles this kind of thing automatically i.e., it automatically re-renders the page based on the new URL pushed to the history. Isn't it not the case?

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+
+  renderResults() {
+    if (this.state.results.length == 0) {
+      return (
+        <p className='mdc-typography--body1'>
+          <FormattedMessage {...messages.noResultsFound} />
+        </p>
+      );
+    }
+    var results = this.state.results.map(result => (
```
gimite: var -> const

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  renderAddPersonFab() {
+    return (
+      <Fab
+          className='results-addpersonfab'
+          onClick={this.goToAdd}
+          icon={<img src='/static/icons/maticon_add.svg' />}
+          textLabel={this.props.intl.formatMessage(
+              messages.createNewRecord)} />
```
gimite: Move `/>` to the next line to be consistent with others?

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+          <SearchBar
+              repoId={this.props.match.params.repoId}
+              initialValue={
+                  new URL(window.location.href).searchParams.get('query_name')}
+              onSearch={this.handleSearch} />
+          {this.renderResults()}
+          <Footer />
+        </div>
+        {this.renderAddPersonFab()}
+      </div>);
```
gimite: Move `);` to the next line to be consistent with others?

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+        <RepoHeader
+          repo={this.state.repo}
+          backButtonTarget={'/' + this.state.repo.repoId}
+        />
+        <div className='results-body'>
+          <SearchBar
+              repoId={this.props.match.params.repoId}
+              initialValue={
+                  new URL(window.location.href).searchParams.get('query_name')}
+              onSearch={this.handleSearch} />
```
gimite: Move `/>` to the next line to be consistent with others?

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * A component for an individual search result.
+ */
+class SearchResultImpl extends Component {
+  constructor(props) {
+    super(props);
+    this.goTo = this.goTo.bind(this);
+  }
+
+  goTo() {
```
gimite: Looks a bit weird that a method named `goTo` takes no parameters. Rename to e.g., `goToView`?

ui/src/pages/Results.js
```diff
@@ -0,0 +1,191 @@
...
+    this.props.history.push(
+        '/' + this.props.repo.repoId + '/view?id='
+        + this.props.result.personId);
```
gimite: Would be safer to apply URL encoding to `personId`.

ui/src/pages/Results.js
nworden: For the most part it does, but not in this case. From what I can tell, the reason is because react-router considers its responsibility to be strictly just the routing, and, since this is staying within the results page (the same route), it's not react-router's job.

nworden: So it looks like componentWillReceiveProps was recently deprecated, and I think getDerivedStateFromProps is more complex than we need, but you're right that the prior code is pretty ugly. It turns out that if we get the query_name value from props.location.search (instead of from window) we can use componentDidUpdate, so I did that.

PR #489, opened 2019-01-23
Component for end bar header.

nworden:


ui/src/components/EndBarHeader.js
```diff
@@ -0,0 +1,37 @@
...
+
+import React, {Component} from 'react';
+
+/*
+ * A component for headers with bars on the ends.
+ *
+ * They look something like this:
+ *      ----- Title or whatever -----
+ */
+class EndBarHeader extends Component {
```
gimite: Optional: It can be a function instead.

Issue #490, opened 2019-01-23
Migrate from mox to mock library


nworden:
mox is a mocking library that hasn't been updated in a long time:
https://pypi.org/project/mox/
we use it for test_tasks, test_external_search, and test_send_mail.

Python has a new library simply called mock, which is built-in for Python 3 and available for Python 2 as a library.

nworden:
It turns out everything else we use mox for is related to AppEngine APIs, which we have to get rid of anyway. It doesn't make a lot of sense to migrate tests that are going to need to be rewritten anyway, so I'm marking this blocked (once we rewrite those tests with Cloud APIs or whatever, we can remove mox.py and close this).

PR #491, opened 2019-01-23
Factor some search logic into Searcher class

nworden:
There's some code duplicated between results.py and api.py, and we're about to need the same code for the React API, so I'm factoring it out into its own class. I put it at search/searcher.py; note that search is a new directory: I figure we have enough search-related code to justify its own directory, so my plan is to later move other search-related stuff (indexing.py, text_query.py, etc.) there too.

Most of the files in this PR are to add a new mocking library to requirements.txt and app/vendors (related: issue #490). The files with interesting stuff going on are search/searcher.py, api.py, results.py, and searcher_tests.py.


app/results.py
```diff
@@ -72,29 +73,24 @@ def max_word_length(query_words):
...
-        # External search backends are not always complete. Fall back to the
-        # original search when they fail or return no results.
-        if not results:
-            if config.get('enable_fulltext_search'):
-                results = full_text_search.search(self.repo,
-                                                  query_dict, MAX_RESULTS)
-            else:
-                results = indexing.search(self.repo,
-                                          TextQuery(query_txt), MAX_RESULTS)
+        searcher = Searcher(
```
gimite: Looks like you create Searcher twice (self._searcher and searcher)?

app/search/searcher.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(
+        self, repo, external_search_backends, enable_fulltext_search,
+        max_results):
+        self._repo = repo
+        self._external_search_backends = external_search_backends
+        self._enable_fulltext_search = enable_fulltext_search
+        self._max_results = max_results
+
+    def search(self, query_dict):
```
gimite: Optional: It can take two params query_name and query_location instead of a dictionary. Up to you though.

app/search/searcher.py
```diff
@@ -0,0 +1,40 @@
+from text_query import TextQuery
+import external_search
+import full_text_search
+import indexing
+
+
+class Searcher(object):
+    """A utility class for searching repositories."""
```
gimite: Maybe "searching person records inside a repository"? Currently it sounds like a class to find a repository.

app/search/searcher.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(
+        self, repo, external_search_backends, enable_fulltext_search,
```
gimite: Optional: Maybe we can deprecate and remove `external_search_backends`. Looks like we haven't used it for years. This command returns no values:

```
$ tools/console google.org/personfinder              
> config.get("external_search_backends")
> config.get_for_repo("japan", "external_search_backends")
```

We used that to provide better Japanese search quality in 2011 which wasn't possible with AppEngine back then. But we haven't use it after that.

nworden:
It just occurred to me that the mock library probably doesn't belong in requirements.txt or app/vendors, since there's no need to upload it to AppEngine. Maybe I should have a separate test_requirements.txt file, and put testing libraries in tests/vendors? mox.py is just included directly in our tests directory, but that doesn't seem like the cleanest option (especially since, unlike mox, mock isn't developed by Google, at least not afaik).

tests/test_searcher.py
```diff
@@ -0,0 +1,79 @@
...
+        pass
+
+    @patch('external_search.search')
+    def test_external_search_backend_results(self, external_search_mock):
+        """Return external search backend results when available."""
+        external_search_mock.return_value = (
+            SearcherTests.EXTERNAL_SEARCH_RETURN_VALUE)
+        external_search_value = 'abc external search'
+        searcher = Searcher(
+            SearcherTests.REPO_NAME, external_search_value, False,
```
gimite: enable_full_text_search=False would be clearer.

tests/test_searcher.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def setUp(self):
+        pass
+
+    @patch('external_search.search')
+    def test_external_search_backend_results(self, external_search_mock):
+        """Return external search backend results when available."""
+        external_search_mock.return_value = (
+            SearcherTests.EXTERNAL_SEARCH_RETURN_VALUE)
+        external_search_value = 'abc external search'
```
gimite: Should be `external_search_backends` or `external_search_backends_value`. Also I believe it is a list, not a string. It may not matter here though.

tests/test_searcher.py
```diff
@@ -0,0 +1,79 @@
...
+        searcher = Searcher(
+            SearcherTests.REPO_NAME, external_search_value, False,
+            SearcherTests.MAX_RESULTS)
+        assert (searcher.search({'name': 'matt'}) ==
+                SearcherTests.EXTERNAL_SEARCH_RETURN_VALUE)
+        assert 1 == len(external_search_mock.call_args_list)
```
gimite: Swap the left hand side and the right hand side for consistency (actual == expected).

tests/test_searcher.py
```diff
@@ -0,0 +1,79 @@
...
+        self, external_search_mock, full_text_search_mock):
+        """Fall back to full_text_search.search results.
+
+        When external search backends don't return any results and full-text
+        search is enabled, fall back to full_text_search.search.
+        """
+        external_search_mock.return_value = []
+        full_text_search_mock.return_value = SearcherTests.FULLTEXT_RETURN_VALUE
+        searcher = Searcher(
+            SearcherTests.REPO_NAME, 'any value', False,
```
gimite: Should be True instead?

nworden:
> > It just occurred to me that the mock library probably doesn't belong in requirements.txt or app/vendors, since there's no need to upload it to AppEngine. Maybe I should have a separate test_requirements.txt file, and put testing libraries in tests/vendors? mox.py is just included directly in our tests directory, but that doesn't seem like the cleanest option (especially since, unlike mox, mock isn't developed by Google, at least not afaik).
> 
> Sounds good to me.

On second thought, I guess we don't really need it in our repo at all. I'll just add it to the GettingStarted doc.

tests/test_searcher.py
nworden: Yeah, you're right, this test shouldn't have been passing. The mock.patch behavior seems to be a little bit weird when you're patching multiple module functions with the same name. It works as I'd expect if I don't do it as a decorator, so I changed the tests to set mocks within the test functions.

tests/test_searcher.py
nworden: Done. What the value is here doesn't really matter, I just need to confirm that it passes the value through to the external_search.search function.

app/results.py
```diff
@@ -72,29 +73,20 @@ def max_word_length(query_words):
...
+            self.repo, self.config.external_search_backends,
+            config.get('enable_fulltext_search'), MAX_RESULTS)
+        results = searcher.search(
+            query_dict['name'],
+            query_dict['location'] if 'location' in query_dict else None)
```
gimite:
```python
query_dict['location'] if 'location' in query_dict else None
```
can be replaced by:
```python
query_dict.get('location')
```

app/search/searcher.py
```diff
@@ -0,0 +1,46 @@
...
+        self._external_search_backends = external_search_backends
+        self._enable_fulltext_search = enable_fulltext_search
+        self._max_results = max_results
+
+    def search(self, query_name, query_location=None):
+        """Get results for a query.
+
+        Args:
+          query_name: A name to query for.
+          query_location: A location to query for (optional).
```
gimite: Optional: May be nice to clarify that this is also a string.

tests/test_searcher.py
gimite: That's true. I just wondered if it can still confuse readers whether it takes a string or a list. Up to you though.

PR #492, opened 2019-01-23
Start of the create record page.

nworden:
This isn't done -- there's obviously a lot of TODOs, the biggest one being the work for the location fields. However, I'm pretty sure it's the last of the React code I've already written, so, if I get and my existing PRs in, the master branch will have a React app that at least runs. Plus, it's already over 600 lines of JS so it doesn't seem like a bad idea to split up the review for it.


ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+      showProfilePageOptions: false,
+    });
+  }
+
+  /*
+   * Removes the given profile page field from the form.
+   */
+  removeProfilePageField(index) {
+    var newProfileList = this.state.formProfilePages.slice(0, index);
+    if (index < this.state.formProfilePages.length - 1) {
```
gimite: I believe this condition is not necessary because slice() safely returns an empty array if index is the last element.

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+          [['site', site], ['value', '']])),
+      showProfilePageOptions: false,
+    });
+  }
+
+  /*
+   * Removes the given profile page field from the form.
+   */
+  removeProfilePageField(index) {
+    var newProfileList = this.state.formProfilePages.slice(0, index);
```
gimite: var -> let

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+    defaultMessage: 'Upload photo',
+    description: 'Label for a button for users who want to upload a photo.',
+  },
+});
+
+/*
+ * The Create page has two tabs: one for adding yourself and one for adding
+ * another person. This enum is used to refer to the two cases.
+ */
+const CREATE_TAB_INDICES = Object.freeze({
```
gimite: You really don't need CREATE_ for this and below because they are private in this file unless you export them.

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+import Tab from '@material/react-tab';
+import TabBar from '@material/react-tab-bar';
+import TextField, {HelperText, Input} from '@material/react-text-field';
+
+import Footer from './../components/Footer.js';
+import LoadingIndicator from './../components/LoadingIndicator.js';
+import PFNotchedOutline from './../components/PFNotchedOutline.js';
+import RepoHeader from './../components/RepoHeader.js';
+
+const messages = defineMessages({
```
gimite: Should have commented this earlier but shouldn't this be named MESSAGES because it's a constant?

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+        <FormattedMessage {...CREATE_PROFILE_PAGE_SITES[site]} />
+      </li>
+    );
+  }
+
+  /*
+   * Render the profile page link fields (if any) and the "Add site" button.
+   */
+  renderProfilePageFields() {
+    var profilePageFields = this.state.formProfilePages.map((page, index) => (
```
gimite: var -> const. Same for below.

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+      <div
+        className='create-profilefieldwrap'
+        key={this.profileFieldIdCounter++}>
```
gimite: How about just using `index` as `key`?

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+        <TextField
+          label={this.props.intl.formatMessage(
+              CREATE_PROFILE_PAGE_SITES[page.get('site')])}
+          outlined
+        >
+          <Input
+            value={this.state.formProfilePages[index].value}
```
gimite: Can be just `page.value`?

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+   * Render the profile page link fields (if any) and the "Add site" button.
+   */
+  renderProfilePageFields() {
+    var profilePageFields = this.state.formProfilePages.map((page, index) => (
+      <div
+        className='create-profilefieldwrap'
+        key={this.profileFieldIdCounter++}>
+        <TextField
+          label={this.props.intl.formatMessage(
+              CREATE_PROFILE_PAGE_SITES[page.get('site')])}
```
gimite: Can be `page.site`?

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+        >
+          <Input
+            value={this.state.formProfilePages[index].value}
+            onChange={(e) => {
+              // A component's state shouldn't be mutated directly. We need to
+              // update a map in an array, so we copy the map and do some
+              // splicing to produce a new array with the new map.
+              var newPageMap = new Map(this.state.formProfilePages[index]);
+              newPageMap['value'] = e.target.value;
+              const newPagesArr = this.state.formProfilePages.slice(0, index)
```
gimite: Maybe consider using Immutable.js which makes this kind of operations easier?:
https://facebook.github.io/immutable-js/docs/#/List
It provides immutable `set` method (and `delete`, `inset` as well).

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+        className='create-profilefieldwrap'
+        key={this.profileFieldIdCounter++}>
+        <TextField
+          label={this.props.intl.formatMessage(
+              CREATE_PROFILE_PAGE_SITES[page.get('site')])}
+          outlined
+        >
+          <Input
+            value={this.state.formProfilePages[index].value}
+            onChange={(e) => {
```
gimite: Would be better to extract this function as a method instead of embedding it inside HTML.

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+
+  renderIdentifyingInfoFields() {
+    var moreFields = this.state.showAllIdInfoFields ?
+        this.renderMoreIdentifyingInfoFields() :
+        this.renderMoreFieldsButton();
+    return (
+      <div className='create-formsectionwrapper'>
+        <div className='create-formgroupwrapper'>
+          <span className='mdc-typography--overline'>
+            Identifying information
```
gimite: I18n this text.

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+            label={this.props.intl.formatMessage(
+                messages.familyOrSurameRequired)}
+            outlined
+          >
+            <Input
+              name='family_name'
+              value={this.state.formSurname}
+              onChange={(e) => this.setState({formSurname: e.target.value})} />
+          </TextField>
+          <TextField
```
gimite: Since there are so many of them with the same structure, you can factor it out as a method (or even a component):
```js
this.renderTextField(
  'formGivenName',
  'given_name',
  messages.givenNameOrFirstNameRequired)
```

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+
+  renderAboutMeForm() {
+    return (
+      <div className='create-formwrapper'>
+        {this.renderIdentifyingInfoFields()}
+      </div>
+    );
+  }
+
+  handleSubmit(e) {
```
gimite: Can be a TODO, but don't we want some loading indicator while submitting?

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+              <span className='mdc-tab__text-label'>
+                <FormattedMessage {...messages.someoneElse} />
+              </span>
+            </Tab>
+          </TabBar>
+        </div>
+        <form onSubmit={this.handleSubmit}>
+          <input
+            type='hidden'
+            name='addType'
```
gimite: `add_type` to be consistent with other fields?

ui/src/pages/Create.js
nworden: Yeah, true. Done for this file here; I'll fix the others in a separate PR.

ui/src/pages/Create.js
nworden: I considered that at first, and it did seem to work in practice, but the React docs call it a last resort and warn that it'll break things if items can change order (we don't change the order of these, but it makes me nervous about using index if we just happen to fit into a niche case where it works):
https://reactjs.org/docs/lists-and-keys.html#keys
I also came across a couple articles warning against it as an anti-pattern:
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318
https://menubar.io/react-keys-index
so it just seemed safer to do it this way.

gimite: It is true that use of an index is not recommended. But I believe always using a new ID like this is even worse.

See https://reactjs.org/docs/reconciliation.html#recursing-on-children for details how keys are used. Couple of things to note:
* (In my understanding) The key only affects performance. So whatever key you give, it won't break the app. It just makes it less efficient.
* React tries to reuse DOM elements with the same key when it re-renders the page. So it helps performance if you give the same key for a component whose content is the same or similar.
* So drawback of using an index as a key is that it cannot reuse the existing DOM elements efficiently when the order is swapped or a component is inserted or removed. But it isn't a big deal for small components like this case.
* If you always give a new ID as a key, the situation gets worse, that it can never reuse existing DOM elements.

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+      </PFNotchedOutline>
+    );
+  }
+
+  /*
+   * Adds a profile page field for the given site to the form.
+   */
+  addProfilePageField(site) {
+    this.setState({
+      formProfilePages: this.state.formProfilePages.concat(new Map(
```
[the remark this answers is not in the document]
nworden: No, not really. Changed.

ui/src/pages/Create.js
nworden: Ok, I changed it to index with a comment warning that we need to be careful about it.

ui/src/pages/Create.js
nworden: Nice, that makes it much cleaner!

ui/src/pages/Create.js
nworden: Using immutable.List now, so resolving the thread.

ui/src/pages/Create.js
```diff
@@ -0,0 +1,645 @@
...
+    this.profileFieldIdCounter = 0;
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.addProfilePageField = this.addProfilePageField.bind(this);
+    this.removeProfilePageField = this.removeProfilePageField.bind(this);
+    this.photoUploadInput = React.createRef();
+  }
+
+  componentDidMount() {
+    const apiUrl = '/' + this.repoId + '/d/repo';
+    fetch(apiUrl)
```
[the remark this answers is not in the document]
nworden: I agree that's probably a good idea, but I'd rather do it in a separate PR since the function would be used by each /pages file.

PR #494, opened 2019-01-23
Minimal backend code for React UI.

nworden:
I think I underestimated how much was gonna be involved in refactoring to avoid code duplication, so some of that's still in progress, but getting this skeleton for the handlers (with some hard-coded dummy data) in will unblock anyone else who wants to work on the React UI.


app/frontendapi.py
```diff
@@ -0,0 +1,112 @@
...
+import photo
+import search
+import simplejson
+import utils
+
+
+class FrontendApiHandler(utils.BaseHandler):
+    """Base handler for frontend API handlers."""
+
+    def __init__(self, request, response, env):
```
gimite: Remove this? Looks it does nothing.

app/frontendapi.py
```diff
@@ -0,0 +1,112 @@
...
+        utils.BaseHandler.__init__(self, request, response, env)
+
+    def _return_json(self, data):
+        self.response.headers['Content-Type'] = (
+            'application/json; charsetutf-8')
```
gimite: `charset=utf-8`

app/frontendapi.py
```diff
@@ -0,0 +1,112 @@
+from django.http import JsonResponse
```
gimite: Is this used?

gimite: Optional: Maybe better naming this file as frontend_api.py (with an underscore)?

gimite: Optional: Can be done later, but if we expect each of the class here gets somewhat complex, may be better to split each class into its own file?

app/frontendapi.py
```diff
@@ -0,0 +1,112 @@
...
+            ]
+        self._return_json(json)
+
+
+class Results(FrontendApiHandler):
+
+    repo_required = True
+
+    def __init__(self, request, response, env):
+        FrontendApiHandler.__init__(self, request, response, env)
```
gimite: Optional: Alternatively this can be written as:
```python
super(Results, self).__init__(self, request, response, env)
```
I guess this is more common and it's clearer that it's calling a superclass?

app/frontendapi.py
```diff
@@ -0,0 +1,112 @@
...
+        self._searcher = search.Searcher(
```
gimite: Optional: I believe an instance of a handler class isn't really reused (as you see request/response in parameters of the initializer). So benefit of initializing searcher here is limited compared to just make it a local variable of get().

app/main.py
```diff
@@ -146,6 +146,11 @@ def repo_options(self):
 HANDLER_CLASSES['tasks/clean_up_in_test_mode'] = 'tasks.CleanUpInTestMode'
 HANDLER_CLASSES['tasks/notify_many_unreviewed_notes'] = 'tasks.NotifyManyUnreviewedNotes'
 HANDLER_CLASSES['tasks/thumbnail_preparer'] = 'tasks.ThumbnailPreparer'
+if config.get('enable_react_ui'):
+    HANDLER_CLASSES['d/create'] = 'frontendapi.Create'
```
gimite: Just curious, what does `d` here stand for? :)

app/main.py
```diff
@@ -573,6 +578,16 @@ def serve(self):
                         (env.repo, env.charset), get_vars)
                 response.out.write(content)
 
+        if config.get('enable_react_ui'):
+            if env.repo == 'static':
+                self.serve_static_content(self.env.action)
```
gimite: Optional: Static contents are served as /global/static/... in the existing app. Maybe the new UI can do the same so that it doesn't require this special handling? Can be done later though.

app/main.py
```diff
@@ -573,6 +578,16 @@ def serve(self):
...
+            elif not env.action.startswith('d/'):
```
gimite: Exclude these as well: api/, admin/, feeds/, sitemap, sitemap/, tasks

app/frontendapi.py
nworden: My goal is to have basically no complexity in this file itself: ideally all the logic will be factored out and shared with the code for the Lite/AMP UI (and the API too, if needed), like with the search code.

nworden: No, removed.

app/main.py
nworden: Done in PR #495.

nworden: Added an explanation to the readme in PR #495; there's a TODO here to make it /global/static.

PR #495, opened 2019-01-23
tools/ui script

nworden:
Used for:
* Running the React app locally
* Building bundles for prod and putting them in the right place in the Django app's directory

### DIFF
--- a/tools/ui
+++ b/tools/ui
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+if [ "$1" == "run" ]; then
+    npm run --prefix ui start
+elif [ "$1" == "buildtoae" ]; then
+    npm run --prefix ui build
+    mkdir -p app/resources/static/ui_bundles
+    mv ui/dist/*-bundle.js app/resources/static/ui_bundles/
+    mv ui/dist/bundle.css app/resources/static/ui_bundles/bundle.css
+    cp -r --no-clobber ui/static/static/* app/resources/static/
+fi
--- a/tools/ui
+++ b/tools/ui
@@ -20,5 +20,6 @@ case "$1" in
     ;;
 *)
     echo "Specify either run or buildtoae."
+    exit 1
     ;;
 esac
--- a/tools/ui
+++ b/tools/ui
@@ -1,11 +1,24 @@
 #!/bin/bash
 
-if [ "$1" == "run" ]; then
+# This script is to help with developing the React UI. It does two things:
+# * Runs a local server (tools/ui run)
+# * Does a production build of the React UI and places the JS and CSS bundles in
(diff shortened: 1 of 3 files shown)


tools/ui
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ "$1" == "run" ]; then
```
gimite: Optional: Use of `case` statement may be better especially when you add more actions later.

gimite: Would be helpful if you add comment to explain what this is and maybe also usage examples.

gimite: Would be better to add this line at the beginning:
```sh
set -eu
```
so that it aborts its execution when one of the commands fails, and makes it an error to use undefined variables.

tools/ui
```diff
@@ -0,0 +1,11 @@
...
+    npm run --prefix ui start
+elif [ "$1" == "buildtoae" ]; then
+    npm run --prefix ui build
+    mkdir -p app/resources/static/ui_bundles
+    mv ui/dist/*-bundle.js app/resources/static/ui_bundles/
+    mv ui/dist/bundle.css app/resources/static/ui_bundles/bundle.css
+    cp -r --no-clobber ui/static/static/* app/resources/static/
```
gimite: Out of scope of this pull request but just curious, why are icons in a directory ui/static/static/icons? Why not ui/static/icons?

nworden: Well, the Webpack dev server expects static files to be in a static directory, and serves whatever's in there from the root directory. But we also serve static files from the /static directory (and the HTML the React app produces needs to point there, because it's what will happen in prod). So, if we put the static files in a static directory within Webpack's static directory, it'll serve files from the same URLs prod uses.
I think there's a way around it by setting up a proxy (as we do with /d to have requests sent to the Django app at port 8000), but I haven't been able to get it set up right yet.

tools/ui
```diff
@@ -0,0 +1,24 @@
...
+    ;;
+"buildtoae")
+    npm run --prefix ui build
+    mkdir -p app/resources/static/ui_bundles
+    mv ui/dist/*-bundle.js app/resources/static/ui_bundles/
+    mv ui/dist/bundle.css app/resources/static/ui_bundles/bundle.css
+    cp -r --no-clobber ui/static/static/* app/resources/static/
+    ;;
+*)
+    echo "Specify either run or buildtoae."
```
gimite: Optional: Also `exit 1`

tools/ui
gimite: I see. It isn't too bad, but maybe good to leave comment somewhere (in static/README.md?) to explain the reason.

PR #496, opened 2019-01-24
Style/consistency updates for GlobalHome.js.

nworden:
"Assigning" this to Eliah, one of the SG engineers who offered to help out with reviews so that Hiroshi doesn't have to review everything ("assigning" in quotes because I don't think he's joined the Google GitHub org yet and that seems to mean I can't officially assign reviews to him. I guess we'll find out if he's able to add review comments without joining.)

* Use single-quotes, per the style guide.
* Call the goTo function something more specific, as we're doing in
  other files.


eliahhecht:
LGTM

PR #505, opened 2019-01-28
Python 3-compatibilify text_query and its tests.

nworden:
"Assigning" this review to Cyrus.

### DIFF
--- a/app/text_query.py
+++ b/app/text_query.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import absolute_import
+import six
 __author__ = 'eyalf@google.com (Eyal Fink)'
 
 import unicodedata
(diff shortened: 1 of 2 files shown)


app/text_query.py
```diff
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import absolute_import
```
gimite: Looks a bit weird that `__author__` is sandwiched with `import`s. Any reason to put those imports before `__author__`?

nworden: Yeah, python-modernize expects it to be at the very top, which kinda makes sense (since it's something that affects the imports themselves), but it is a little bit silly that it thinks it matters where __author__ is. It's not just a matter of where it puts it by default, it'll also complain if you fix it manually, so if we want to use python-modernize to avoid regressions we have to tolerate this weirdness.

gimite: I see. Is it still possible to put a blank line before `__author__` to make it look a bit less weird?

Or maybe we can just remove `__author__`. We don't add it for recent files...

nworden: I added a blank line above `__author__` in PR #553.

PR #506, opened 2019-01-28
Add six library to requirements.txt.

nworden:
This is kinda just a formality, since we already have dependencies that
depend on it (which is why nothing's getting added to app/vendors), but
since we're now depending on it directly it belongs in requirements.txt.


christopherestif:
lgtm ^.^

PR #509, opened 2019-01-28
Add Python 3 compatibility check.

nworden:
The python-modernize command will fail if one of the files in the list isn't Python 3 compatible. My plan is to add files to that list as I update them, so that once a file is made compatible we don't accidentally break its compatibility later.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN apt-get update && apt-get install -y \
 	&& rm -rf /var/lib/apt/lists/*
 
 RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && /usr/bin/python2.7 get-pip.py
-RUN pip install pytest==3.7.4 lxml cssselect pillow==4.1.0 mock
+RUN pip install pytest==3.7.4 lxml cssselect pillow==4.1.0 mock modernize
 
 # Install app engine
 WORKDIR   /opt/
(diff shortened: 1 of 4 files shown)


tools/p3_compatibility_check
```diff
@@ -0,0 +1,6 @@
+python-modernize --enforce \
```
gimite: Maybe add a comment why it doesn't check all files (yet).

gimite: Add copyright header and a comment explaining what this command does.

Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN apt-get update && apt-get install -y \
 	&& rm -rf /var/lib/apt/lists/*
 
 RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && /usr/bin/python2.7 get-pip.py
-RUN pip install pytest==3.7.4 lxml cssselect pillow==4.1.0 mock
+RUN pip install pytest==3.7.4 lxml cssselect pillow==4.1.0 mock modernize
```
gimite: Can you add instruction to install this on
https://github.com/google/personfinder/wiki/GettingStarted
as well for developers not using Docker?

tools/p3_compatibility_check
gimite: Add:
```
#!/bin/bash
```

gimite: Maybe better to spell out the file name as python3_compatibility_check? May not be obvious what p3 means.

PR #511, opened 2019-01-29
Assorted UI fixes

nworden:
I wrote a bunch of code and then tried to split it up on separate branches to send for review. It was messy and I broke some stuff in the review/merging process. This fixes those things.


ui/src/components/PFNotchedOutline.js
```diff
@@ -24,12 +24,12 @@ import React, {Component} from 'react';
  * pretty sparse and b) I'm not sure Material supports using their
  * NotchedOutline with anything other than TextField and Select components.
  */
-const PFNotchedOutline = () => (
+const PFNotchedOutline = (props) => (
```
cyrusboadway: Parameter type?

nworden: From what I can tell from other React code, it's just an object (props, along with state, is kinda a React thing; we don't create it ourselves). Added a param comment.

PR #512, opened 2019-01-29
Upgrade lxml to version 3.7.3.

nworden:
This isn't the latest version of lxml, but it is the latest version that App
Engine supports (we can't app/vendors it because it's a wrapper around C
libraries). It's listed as supporting up to Python 3.5, which hopefully means
it'll work on 3.7 too (version 2.3 only officially supports up to Python 3.2).

Assigning to Avi.


nworden:
Ignore this PR for now, it looks like it really does break Travis.

PR #523, opened 2019-02-01
Implement view page (partially)

gimite:
Will add the rest of the fields later.


gimite:
@nworden I will be on vacation on the next Monday, and I may not have time to continue working on this in the rest of the next week due to a summit. In case this work blocks your work, feel free to take over this pull request and make changes by yourself.

ui/src/pages/View.js
```diff
@@ -17,9 +17,106 @@
...
+const MESSAGES = defineMessages({
+  age: {
+    id: 'View.age',
+    defaultMessage: 'Age',
+    description: 'A label for a person\'s age.',
+  },
+  authorOfThisRecord: {
+    id: 'View.authorOfThisRecord',
+    defaultMessage: 'Author of this record',
+    description: ('The title of a section in a person record which contains '
```
nworden: Since these descriptions are going to get sent to translators, it might be good to avoid descriptions that refer to "person records" since a translator's unlikely to know quite what that means. Maybe something like "The title of a section with information about the source/author of the rest of the information on the page."

ui/src/pages/View.js
```diff
@@ -17,9 +17,106 @@
...
+    id: 'View.identifyingInformation',
+    defaultMessage: 'Identifying information',
+    description: 'The title of a section which contains a person\'s '
+        + 'identifying information (name, age, etc.).'
+  },
+  notesForThisPerson: {
+    id: 'View.notesForThisPerson',
+    defaultMessage: 'Notes for this person',
+    description: ('The title of a section which contains notes for the person '
+        + 'record.'),
```
nworden: Maybe "which contains notes with updates or additional information about the status of the person described on the page"?

ui/src/pages/View.js
```diff
@@ -29,16 +126,21 @@ class View extends Component {
       repo: null,
       person: null
     };
+    this.handleSearch = this.handleSearch.bind(this);
+    this.renderField = this.renderField.bind(this);
```
nworden: My understanding is that the .bind(this) call is only needed for event handlers or other functions that get passed as arguments. I'm pretty sure handleSearch is the only function you need to do it for.

ui/src/pages/View.js
```diff
@@ -56,14 +158,148 @@ class View extends Component {
...
+      });
+  }
+
+  renderField(labelMessage, value) {
+    return (
+      <div>
+        <span className='view-fieldname'>
+          <FormattedMessage {...labelMessage} />:
+        </span>
+        <span> </span>
```
nworden: Curious: why a span here? Is it bad to use &nbsp; to force a space?

ui/src/pages/View.js
```diff
@@ -56,14 +158,148 @@ class View extends Component {
...
+              <FormattedMessage {...MESSAGES.description} />
+            </h2>
+            <div className='view-fieldvalue'>{this.state.person.description}</div>
+          </div>
+
+          <div className='mdc-typography--body1'>
+            <h2 className='mdc-typography--subtitle2'>
+              <FormattedMessage {...MESSAGES.profilePages} />
+            </h2>
+            <div>TBD</div>
```
nworden: Put a TODO here (and below)?

ui/src/css/pages/_View.scss
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+.view-body {
+  /* The 100px bottom margin ensures that the page can be scrolled far enough to
+   * show the footer text without it being covered by the new record FAB. */
+  margin: 16px 0 100px;
+  text-align: center;
+}
+
+.view-card {
```
nworden: Any reason not to use the Material React library's [Card component](https://github.com/material-components/material-components-web-react/tree/master/packages/card) for this? We're using it on the global homepage (for repo link cards and the third-party devs/responders cards).

gimite: I basically copied implementation of _Results.scss because this card looks similar to the search result card (though I made some modification). So can I ask you back: Why didn't you use Card component in Results.js?

nworden: Ha, fair question.
IIRC, the reason I didn't use a Card for Results.js is because the results page doesn't have a separate card for each result, it just has a card-shaped wrapper around everything with horizontal rules between them. It didn't seem right to use a Card for a list of items, and [the MDC web docs](https://github.com/material-components/material-components-web/tree/master/packages/mdc-card) say that "Cards contain content and actions about a single subject." That means the wrapper around results shouldn't be a card because it's about many subjects, but we could use it for an individual result if we wanted.
How much of the _Results.scss CSS did you copy, and what did you need to change? If it'd be practical for them to share some of their CSS, that seems like a good reason to do it that way instead of using Card component, but if they're not sharing classes then IMO it should probably use a Card.

ui/src/pages/View.js
nworden: I see, yeah, that was what I was suggesting.

ui/src/css/pages/_View.scss
gimite: I needed these changes:
* Support of sections without a photo:
thgvqfgsrhs
* Rename of classes (since it's not a list)
* (A change in HTML not CSS but) Use div instead of ul/li (because ul/li is weird here)

So I believe I can rewrite results page with CSS classes defined here (but not vice versa). Does that sound good? If so, can I submit this PR for now and make that change later?

ui/src/pages/View.js
gimite: FYI Looks like this is one of the common ways to do that according to StackOverflow (https://stackoverflow.com/a/41142436/4339963).

ui/src/css/pages/_View.scss
nworden: Yeah sure.

PR #524, opened 2019-02-05
Upgrade to Django 1.11.

nworden:
I went through the release notes for 1.10 and 1.11 and couldn't find
anything that required code changes in our app (which is a little
surprising but not shocking, since we don't use Django for database
access).

One of our dependencies, oauth2client, uses the old location for
django.core.urlresolvers (it's now django.urls). The old location is
deprecated in 1.10 but apparently not yet removed. So, oauth2client
might eventually be a blocker for upgrading to Django 2.x, but it's
still fine for Django 1.11.

### DIFF
--- a/app/app.yaml
+++ b/app/app.yaml
@@ -42,6 +42,6 @@ handlers:
 
 libraries:
 - name: django
-  version: "1.9"
+  version: "1.11"
 - name: lxml
   version: "3.7.3"


nworden:
Well, this certainly breaks a lot of tests in Docker. Oddly, they all pass if I run them directly on my machine, but it also doesn't seem to have anything to do with the AppEngine SDK (like the issues with lxml from PR #512). I'll reopen this when I figure this out.

Issue #478, opened 2019-01-21
Implement UI v3 with React


gimite:
@nworden I created this because I couldn't find an issue for the overall UI refresh effort. But let me know if it is already somewhere.

I arbitrarily called it "v3" because the current UI was called "v2". Feel free to give it your preferred name :)

yashLadha:
Hey I would like to work on this issue and work alongside with @nworden . @gimite

nworden:
Hi Yash,
I'm sorry for my delay in replying (I was offline yesterday). Thanks for your offer to help!
Are you already familiar with React? If not, the tutorial on their website is pretty good so I'd suggest starting there. If you're comfortable with React already, let me know and we can find a piece of the UI work to break off for you.
- Nick

yashLadha:
@nworden  Yes, I am comfortable with React and had created a couple of personal projects in react+redux as well. So I would be happy to contribute to this project.

nworden:
Ok, that's great. Here are a few options that might be good places to start:
- A component for our static pages. We have a few pages that display static information (e.g., [the page for responders](https://google.org/personfinder/global/responders.html)), and we'll need to support them in the React UI. The implementation for each of these pages will probably be very similar, so I think the way to do it would be to create a general StaticPage component and subclass it for specific pages. [Here's a mock of one of the pages.](https://user-images.githubusercontent.com/9059304/52650900-d19a6d00-2ef3-11e9-9fad-8e742efbd590.png)
- Support a turndown message. When a repository for an event is getting less traffic and gets scheduled to be turned down, we want to announce the upcoming turndown to users. This would involve some backend work too, to pass the fact of the impending deactivation to the frontend.

Does one of those sound like something you'd be interested in doing? Don't hesitate to let me know if you have any questions!

yashLadha:
Sounds great. I will start working on the general static component ASAP and can you provide me with some more insight on the second task, as it will be a good task for learning purpose. @nworden . I am creating a separate issue for the first point. Also how about using typescript instead of javascript for frontend in React. This can possibly avoid a lot of bugs in production.

nworden:
[...]

PR #526, opened 2019-02-13
Fixes #525: Corrected Repo Header Link to go to Homepage

### LINKED ISSUE #525: Back button in RepoHeader doesn't takes back
When on clicking the back button on repo header, it doesn't go to the previous page and remains on the same page itself.

I am sending the patch for this.

yashLadha:
Fixes #525: Fixed the Repo Header link to go to home page.

### DIFF
--- a/ui/src/pages/RepoHome.js
+++ b/ui/src/pages/RepoHome.js
@@ -118,7 +118,7 @@ class RepoHome extends Component {
       <div>
         <RepoHeader
           repo={this.state.repo}
-          backButtonTarget={`/${this.repoId}`}
+          backButtonTarget={'/'}
         />
         <div className='repohome-body'>
           <SearchBar


yashLadha:
I signed it!

yashLadha:
@nworden Please review this

nworden:
Hey Yash! Thanks for this fix!
I'll talk with our UXer about whether we ought to have a back button on the homepage at all: during crises, people often follow links from social media, etc. that go directly to the repo homepage rather than the global homepage, so maybe we ought to leave the back button off the homepage altogether, but that's a UX decision. If we keep the back button, having it go to the global homepage makes more sense to me than having it go nowhere, so I'll merge in this PR.

yashLadha:
Sound's good @nworden . Also, I would like to know the perspective on static pages as well. Like should I create a HOC for all these components or create an interface from which they all get extended.

nworden:
Just FTR, I continued the discussion on issue #478.

PR #530, opened 2019-02-15
add .editorconfig file

### LINKED ISSUE #529: add .editorconfig file
Defing style guide rules in a **.editorconfig** file, DOS/Windows style line endings cause errors in running the project, **editorconfig** file tells editors to use linux style line endings in all files by default.

mohitkyadav:
Fixes #529

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,4 @@
+# https://editorconfig.org/
 root = true
 
 [*]
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,6 +13,3 @@ max_line_length = 80
 [*.py]
 indent_style = space
 indent_size = 4
-
-[*.md]
-trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,24 +3,15 @@ root = true
 [*]
 trim_trailing_whitespace = true
 end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = space
+indent_size = 2
+max_line_length = 80
 
 [*.py]
 indent_style = space
 indent_size = 4
-charset = utf-8
-insert_final_newline = true
-max_line_length = 120
 
-
-[*.{js,css}]
-indent_style = space
-indent_size = 2
-charset = utf-8
-insert_final_newline = true
-max_line_length = 120
-
-[*.html]
-indent_style = space
-indent_size = 2
-charset = utf-8
-insert_final_newline = true
+[*.md]
+trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+root = true
+
+[*]
+trim_trailing_whitespace = true
+end_of_line = lf
+
+[*.py]
+indent_style = space
+indent_size = 4
+charset = utf-8
+insert_final_newline = true
+max_line_length = 120
+
+
+[*.{js,css}]
+indent_style = space
+indent_size = 2
+charset = utf-8
+insert_final_newline = true
+max_line_length = 120
+
+[*.html]
+indent_style = space
+indent_size = 2
+charset = utf-8
+insert_final_newline = true


mohitkyadav:
@gimite Please review the rules.

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+[*]
+trim_trailing_whitespace = true
+end_of_line = lf
+
+[*.py]
+indent_style = space
+indent_size = 4
+charset = utf-8
+insert_final_newline = true
+max_line_length = 120
```
nworden: We have a max line length of 80 characters for Python (per the [Google Python style guide](http://google.github.io/styleguide/pyguide.html#32-line-length)).

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+
+[*]
+trim_trailing_whitespace = true
+end_of_line = lf
+
+[*.py]
+indent_style = space
+indent_size = 4
+charset = utf-8
+insert_final_newline = true
```
nworden: Are there file types for which we wouldn't want these charset and insert_final_newline settings, or could this be in the \[\*\] block above instead of copied in each of these filetype blocks?

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+max_line_length = 120
+
+
```
nworden: Nit: there's an extra newline between these blocks.

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+[*.{js,css}]
+indent_style = space
+indent_size = 2
+charset = utf-8
+insert_final_newline = true
+max_line_length = 120
```
nworden: 80 characters here too (per [Google JS style guide](https://google.github.io/styleguide/jsguide.html#formatting-column-limit)).

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+
+[*.py]
+indent_style = space
+indent_size = 4
+charset = utf-8
+insert_final_newline = true
+max_line_length = 120
+
+
+[*.{js,css}]
```
nworden: The style guides we use combine HTML and CSS into one style guide, and have a separate one for JS, so I suggest making this block just for JS and putting CSS together with HTML below.

.editorconfig
mohitkyadav: changed

.editorconfig
mohitkyadav: I cmmon rules combined into one.

.editorconfig
mohitkyadav: updated

mohitkyadav:
Updated, @nworden please take a look.

.editorconfig
```diff
@@ -0,0 +1,17 @@
+root = true
```
gimite: Can you add a comment what this file is (with a pointer to https://editorconfig.org/), assuming .editorconfig allows putting comments in it?

.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+indent_style = space
+indent_size = 2
+max_line_length = 80
+
+[*.py]
+indent_style = space
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false
```
gimite: Just curious, why don't you trim trailing whitespaces in .md? Are there cases where it shouldn't be removed?

mohitkyadav: @gimite removing trailing white space from .md can cause to break the formatting after rendering the markdown because markdown uses 4 spaces to indicate a new line and white spaces are  part of the language itself.

.editorconfig
mohitkyadav: sure.

.editorconfig
nworden: I don't think we use spaces-for-newlines anywhere. Is there ever a reason to prefer 2/4 spaces for newlines instead of newlines or a backslash? The advantage of those options is it's more readable if you're looking at the raw .md file, since you wouldn't see the trailing whitespace. (Or, is there anywhere else Markdown considers trailing whitespace meaningful?)

mohitkyadav: @nworden I'm not suggesting to use spaces, but removing trailing whitespace from markdowns files limits the language itself. 
I also think it's more readable if we look at the raw .me file without trailing whitespace.
Do you want me to remove the `trim_trailing_whitespace`?

gimite: @mohitkyadav Can you maybe give us an example Markdown where it makes difference by removing trailing whitespace?

mohitkyadav: Here's the screenshot of trimmed whitespace and with whitespace.
without trimmed
trimmed

gimite: I see. I believe what @nworden means is that we won't write in that way, and instead we should write like this:
```md
Astralis wins \
NaVi wins
```

So it should be safe to trim trailing whitespaces.

mohitkyadav: No problem I'll remove that.

mohitkyadav:
@nworden @gimite, changes made, please take a look.

PR #531, opened 2019-02-19
Set up Jest test runner

nworden:
Jest is FB's JS testing library which we're using for testing our React app. We might need to add Enzyme or something like that later too, but I'll figure that out later.
This PR also includes tests for Utils.js, because running jest with 0 tests makes it sad.


nworden:
Assigning the review to Cyrus.

Dockerfile
```diff
@@ -25,6 +25,11 @@ RUN apt-get update && apt-get install -y \
 RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && /usr/bin/python2.7 get-pip.py
 RUN pip install pytest==3.7.4 lxml cssselect pillow==4.1.0 mock modernize
 
+# Just apt-get installing nodejs doesn't work; these commands come from:
+# https://askubuntu.com/questions/720784/how-to-install-latest-node-inside-a-docker-container
+RUN curl -sL https://deb.nodesource.com/setup_11.x | bash
```
cyrusboadway: Looks like there's a 12 now: https://github.com/nodesource/distributions/blob/master/deb/setup_0.12

Poking around it looks like there's a `setup` and `setup_dev` alias which point to v10 and v12 respectively. If these are the 'stable' and 'beta/edge' versions, would it be worthwhile to point directly to `setup`?

ui/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "start": "webpack-dev-server --config webpack.dev.js",
     "build": "webpack --config webpack.prod.js",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "jest"
```
cyrusboadway: If you'd like to use the `_test.js` suffix instead of `.test.js`, it looks like you can add [configuration](https://jestjs.io/docs/en/configuration#testmatch-array-string), either as a separate file, or directly in `package.json`:

```json
{
  "jest": {
    "testMatch": [...],
  },
  "scripts": {
    "test": "jest",
  },
  "devDependencies": {
    "jest": "^24.1.0",
  }
}
```

Dockerfile
nworden: It looks like setup_0.12 is for v0.12 (i.e., pre-v1) rather than v12. `setup` and `setup_dev` appear to point to 0.10 and 0.12, which makes me think they're not used anymore. v10 does seem to be the stable version though, so I'll use setup_10.x instead of 11.x.

ui/package.json
nworden: Cool, done.

PR #532, opened 2019-02-19
Changed link schema from http to https

adityaxdiwakar:
Changed the schema for the link to use https, so that when a user clicks on the link through this README, they use SSL to ensure a trusted exchange.


nworden:
Thanks Aditya!

PR #539, opened 2019-02-25
Tests for GlobalHome.

nworden:


nworden:
Sent the review to Cyrus.

nworden:
As discussed offline, I updated it to use await/async in componentDidMount, which (long story short) means it's not necessary to use setImmediate in the tests. The babel plugin stuff is to support the use of await/async.
I'll update the other pages similarly as I add tests for them.
I don't really get why it was necessary to define asyncMountWithIntl -- it seems like I should be able to just call `await mountWithIntl` directly, but that didn't work.

ui/src/testing/enzyme-intl-helper.js
```diff
@@ -69,3 +69,7 @@ export function mountWithIntl(node, { context, childContextTypes, ...additionalO
         }
     );
 }
+
+export async function asyncMountWithIntl(node, { context, childContextTypes, ...additionalOptions } = {}) {
```
cyrusboadway: I'm surprised calling `await mountWithIntl` didn't work. Maybe chuck a comment on here for future explorers.

ui/src/pages/GlobalHome_test.js
```diff
@@ -0,0 +1,103 @@
...
+        });
+  });
+
+  test('snapshot test for GlobalHome', async (done) => {
+    await runPageTest(done,
+        async () => {
+          // We don't use setupPageWrapper here because we need to avoid passing
+          // a history object: it will generate random keys that mess up the
+          // snapshot.
+          fetch.mockResponseOnce(JSON.stringify([
```
cyrusboadway: Optional, maybe extract the JSON string to a const instead of redeclaring.

ui/src/pages/GlobalHome_test.js
```diff
@@ -0,0 +1,103 @@
...
+  );
+  return [wrapper, history];
+}
+
+describe('testing GlobalHome', () => {
+  beforeEach(() => {
+    fetch.resetMocks();
+  });
+
+  test('repo cards are rendered', async (done) => {
```
cyrusboadway: The `async` and `done` are generally mutually exclusive.

If you're not testing anything asynchronous, you don't need the test method to `async` and you don't need done. Return nothing from the method and the framework will assume the test has completed when the method ends.

If you're testing async stuff (as in this case), you can:

1. Return a promise which will resolve when all async behaviour and assertions have been made. The framework will know the test is done when the promise has completely resolved.
2. Make the test method `async`. The framework will assume the test is done when the async method completes. (This is equivalent to the above).
3. Accept the `done` param in its signature and call it when the test is done. In most cases, you don't need this because 1 or 2 are sufficient, but there are edge cases where this is still handy (e.g. the application code is hiding the async behaviour, but you're able to pass it the `done` as some kind of completing callback).

nworden: I ended up going back to using done. I'm not sure it's worth it to include the dependencies I would need to deal with the regeneratorRuntime stuff and support async/await.

ui/src/pages/GlobalHome_test.js
nworden: Done.

ui/src/testing/utils.js
```diff
@@ -0,0 +1,25 @@
+/**
```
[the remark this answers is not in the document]
nworden: I think it's still worth it now that setImmediate is staying :/

nworden:
I figured out the problem we were talking about this morning: Enzyme's `mount` call doesn't wait for `componentDidMount` (which, in hindsight, makes perfect sense: `componentDidMount` would be expected to run after `mount`). The problem is that this leaves me with no handle on the `componentDidMount` call, so there's no direct way for me to wait for it. The three options I can turn up are:
* Call `componentDidMount` directly myself. This would solve my immediate problem, but it would sorta compromise the test because it'd mean `componentDidMount` gets called twice (which is *probably* fine here, but in general could have implications around the logic of the code under test), so I don't want to do that.
* Wait x milliseconds.
* Use `setImmediate`.

So I went back to using `setImmediate`. I reset my branch back to the earlier point because that was the easiest way to clean out the various libraries I'd needed to support async/await (which doesn't seem worth it if I'm using setImmediate anyway) and then updated the code from that point to address your other comments.

ui/src/pages/GlobalHome_test.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+describe('testing GlobalHome', () => {
+  beforeEach(() => {
+    fetch.resetMocks();
+  });
+
+  test('repo cards are rendered', (done) => {
+    runPageTest(
+        () => {
```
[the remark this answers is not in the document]
nworden: As discussed, I dropped `runPageTest`, so now it's not getting passed anyway.

ui/src/pages/GlobalHome_test.js
```diff
@@ -0,0 +1,102 @@
...
+    const [wrapper, _] = setupPageWrapper();
+    flushPromises().then(() => {
+      wrapper.update();
+      const expectedDisplayCategories = [0, 1, 2, 3, 4, 0, 1];
+      for (let i = 0; i < expectedDisplayCategories.length; i++) {
+        expect(wrapper.find('GlobalHome').state().repos[i].displayCategory)
+            .toBe(expectedDisplayCategories[i]);
+      }
+      wrapper.unmount();
+      done();
```
cyrusboadway: You can return the promise from the test method, and it'll be considered done when the promise resolves.

ui/src/pages/GlobalHome_test.js
```diff
@@ -0,0 +1,102 @@
...
+      for (let i = 0; i < expectedTitles.length; i++) {
+        expect(repoCardTitles.at(i).text()).toBe(expectedTitles[i]);
+      }
+      wrapper.unmount();
+      done();
+    });
+  });
+
+  test('assigns display categories to repos', (done) => {
+    const [wrapper, _] = setupPageWrapper();
```
cyrusboadway: You can omit the unused param: `const [wrapper] = setupPageWrapper()`

ui/src/testing/utils.js
```diff
@@ -0,0 +1,11 @@
...
+ * Returns a promise to run resolve after existing promises are flushed.
+ *
+ * Long story short, we can't properly test components until componentDidMount
+ * has run, but, when componentDidMount has async calls (e.g., to fetch data and
+ * then use it to set state), React/Enzyme leaves us without any good ways to
+ * wait for componentDidMount. We use setImmediate to get around this.
+ */
+export function flushPromises() {
+  return new Promise((res) => setImmediate(res));
```
cyrusboadway: You might be able to write this as `return new Promise(setImmediate);`

ui/src/pages/GlobalHome_test.js
nworden: Oh, nice.

ui/src/pages/GlobalHome_test.js
nworden: Done.

ui/src/testing/utils.js
nworden: I can.

PR #540, opened 2019-02-25
Support legacy /global/home.html URL.

nworden:


ui/src/App.js
```diff
@@ -29,13 +29,14 @@ import './css/all.scss';
...
     <div>
-      {/* TODO(nworden): include support for legacy homepage URL:
-          global/home.html */}
       <Route exact path='/' component={GlobalHome} />
       <Route exact path='/:repoId' component={RepoHome} />
       <Route exact path='/:repoId/create' component={Create} />
       <Route exact path='/:repoId/results' component={Results} />
       <Route exact path='/:repoId/view' component={View} />
+      <Route exact path='/global/home.html' render={() => (
+            <Redirect to='/'/>
```
eliahhecht: Is this a temporary, or a permanent redirect? We probably want permanent.

nworden: Thanks, good point. It looks like react-router doesn't support redirects with any particular status codes, which I imagine is because they're not really HTTP redirects (once the React app is served up, the client has gotten the status code they're going to get), so I should really be doing this on the server side. In the course of looking into that, I found some behavior I didn't expect (see issue #542), which I want to check with Hiroshi about before I proceed. I'll close this PR for now, and reopen it/ping you once I've implemented the redirect on the server side instead.

PR #544, opened 2019-02-27
Install requests and requests-toolbelt.

nworden:
We're going to switch from urlfetch to requests at some point to prepare
for the migration to the Python 3 runtime (because urlfetch is going
away).


nworden:
I'm splitting this out from PR #543, because all those app/vendors files are gonna make it annoying to review anything else.

PR #549, opened 2019-02-28
New sitemap implementation. Fixes #456.

### LINKED ISSUE #456: Sitemap and sitemap ping not working
I found that sitemap and sitemap ping of Person Finder are not working.

And it looks like it is failing for long, at least for 7 years.

Maybe we can just remove it because Google probably does a good job finding pages quickly without sitemap, and we actually have seen no issues in these 7 years? WDYT? @nworden 

One potential counterargument is that it may still help Google to find a new repository even more quickly.

I noticed that because "sitemap ping" cron job is failing in our production. Also https://google.org/personfinder/japan/sitemap returns an error.

I found multiple problems:
* The URL specified in the cron job, /sitemap/ping?search_engine=google, is 404. It misses a repository name.
* /sitemap fails due to out of memory.
* The datastore contained two entities of StaticSiteMapInfo, both created at 2011, while the code expects only one entity. I have just deleted one of them.

nworden:
There's a few things going on in this PR:
* Changes the sitemap to only point to repo pages, not individual person pages (which was broken and also not desired behavior today).
* Includes links to alternate language pages in the sitemap, as described [here](https://support.google.com/webmasters/answer/189077#sitemap).
* Takes the sitemap ping task out of cron.yaml and does it when repos are modified instead.
* Use the requests library in place of urlfetch in the sitemap handler (other existing uses of urlfetch are untouched for now). We need to switch everything to requests eventually as part of the Python 3 migration.
* Removes an unused import in `tests/test_send_mail.py`.

### DIFF
--- a/app/sitemap.py
+++ b/app/sitemap.py
@@ -58,7 +58,7 @@ class SiteMapPing(BaseHandler):
         'google': 'http://www.google.com/ping?sitemap=%s',
     }
 
-    GET_PARAM_SEARCH_ENGINE = 'search_engine'
+    _GET_PARAM_SEARCH_ENGINE = 'search_engine'
 
     repo_required = False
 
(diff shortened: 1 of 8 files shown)


nworden:
Sending to Eliah.

app/sitemap.py
```diff
@@ -15,142 +15,74 @@
...
+from google.appengine.api import taskqueue
+
+import const
+from model import Repo
+from utils import BaseHandler
+
+
+# Use the App Engine Requests adapter. This makes sure that Requests uses
+# URLFetch.
+# TODO(nworden): see if we should condition this on the runtime (Python 2 vs. 3)
```
eliahhecht: Did you mean to leave this TODO in here?

app/sitemap.py
```diff
@@ -15,142 +15,74 @@
...
+
+    @staticmethod
+    def add_ping_tasks():
+        for search_engine in SiteMapPing._INDEXER_MAP:
+            name = 'sitemapping-%s-%s' % (search_engine, int(time.time()*1000))
+            taskqueue.add(name=name, method='GET', url='/global/sitemap/ping',
+                          params={'search_engine': search_engine})
 
     def get(self):
         search_engine = self.request.get('search_engine')
```
eliahhecht: Nit: maybe add a string constant for this 'search_engine' parameter, to protect against typos?

app/sitemap.py
nworden: Yeah.

app/sitemap.py
nworden: Done.

PR #550, opened 2019-02-28
Issue/525

yashLadha:
General Static Component for rendering static pages

### DIFF
--- a/ui/src/components/GeneralStatic.js
+++ b/ui/src/components/GeneralStatic.js
@@ -1,0 +1,24 @@
+import React, { Component } from 'react';
+import Footer from './Footer';
+
+/**
+ * General static component for rendering pages
+ *
+ * @class StaticComp
+ * @extends {Component}
+ */
+const StaticPage = ({ subtitle, children }) => (
+  <div className='staticpage-wrapper'>
+    {/* TODO(nworden): Refactor repoheader and use it here */}
+    <div id='repoheader-info'>
+      <p className='mdc-typography--subtitle1'>{subtitle}</p>
+    </div>
+    <div className='staticpage-contents'>
+      { children }
+    </div>
+    <Footer />
+  </div>
+);
+
+
+export default StaticPage;
--- a/ui/src/pages/RepoHome.js
+++ b/ui/src/pages/RepoHome.js
@@ -118,7 +118,7 @@ class RepoHome extends Component {
       <div>
         <RepoHeader
           repo={this.state.repo}
-          backButtonTarget={`/${this.repoId}`}
+          backButtonTarget={'/'}
         />
         <div className='repohome-body'>
           <SearchBar


yashLadha:
@nworden  Can you please review and suggest changes?

nworden:
I haven't gotten a chance to do a review yet, but just FYI that test failure looks to me like a flake. If it continues failing on later commits, we can worry about it then, but for now I don't think you need to look into it.

ui/src/components/GeneralStatic.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from "react";
+import { FormattedMessage } from "react-intl";
+
+/**
+ * General static component for rendering pagesS
+ *
+ * @class StaticComp
+ * @extends {Component}
+ */
+const StaticComp = ({ subtitle, children }) => (
```
nworden: Maybe call it `StaticPage` instead.

ui/src/components/GeneralStatic.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from "react";
+import { FormattedMessage } from "react-intl";
```
nworden: I don't think you need to import this.

ui/src/components/GeneralStatic.js
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * General static component for rendering pagesS
```
nworden: Nit: there's an extra 'S' at the end of this line

ui/src/components/GeneralStatic.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @class StaticComp
+ * @extends {Component}
+ */
+const StaticComp = ({ subtitle, children }) => (
+  <div>
+    <div id='repoheader-info'>
+      <p className='mdc-typography--subtitle1'>{subtitle}</p>
+    </div>
+    { children }
```
nworden: Please wrap the children in a div with a unique class name (`staticpage-contents` might be a good choice). That way we can apply spacing or stuff like that we might need.

nworden: I think we'll want to include the footer on every page, so please include it after the children (the footer component is in `components/Footer.js`).

ui/src/components/GeneralStatic.js
```diff
@@ -0,0 +1,20 @@
...
+import { FormattedMessage } from "react-intl";
+
+/**
+ * General static component for rendering pagesS
+ *
+ * @class StaticComp
+ * @extends {Component}
+ */
+const StaticComp = ({ subtitle, children }) => (
+  <div>
```
nworden: We'll probably need a class name to style this div. Could you call it `staticpage-wrapper` or something like that?

ui/src/components/GeneralStatic.js
```diff
@@ -0,0 +1,20 @@
...
+    <div id='repoheader-info'>
```
nworden: You can definitely go ahead with this PR without doing the `RepoHeader` refactor if you want, but could you add a to-do for me to do it? I can never remember for sure how to do JSX comments, but I think it'd be something like this:
`{//TODO(nworden): refactor RepoHeader and use it here}`

ui/src/components/GeneralStatic.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from "react";
```
nworden: Nit: for the sake of consistent style, please use single quotes here, and for `./Footer`  on the next line.

PR #551, opened 2019-03-01
Only ping sitemap index servers in prod.

nworden:
The proximate reason for this is that PR #549 caused admin config tests to really try to ping Google and Bing's index servers, which is bad for the obvious reason and also seems to be causing test flakes. I decided to fix it by having the sitemap ping handler check if it was in prod, rather than mocking stuff out in the admin config tests, because it's probably a good idea to stop local servers or dev instances from really pinging sitemap indexers too (plus, this way we won't have to worry about remembering it when we write new admin tests).

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -597,6 +597,10 @@ def strip_url_scheme(url):
 def is_dev_app_server():
     return os.environ['APPLICATION_ID'].startswith('dev~')
 
+
+def is_prod_server():
+    return os.environ['APPLICATION_ID'] == 'googlepersonfinder'
+
 # ==== Struct ==================================================================
 
 class Struct:
--- a/app/utils.py
+++ b/app/utils.py
@@ -597,17 +597,6 @@ def strip_url_scheme(url):
 def is_dev_app_server():
     return os.environ['APPLICATION_ID'].startswith('dev~')
 
-
-def is_prod_server():
-    """Returns True if this is the production server.
-
-    This is not the simple opposite of is_dev_app_server (above), because it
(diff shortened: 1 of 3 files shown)


app/utils.py
```diff
@@ -597,6 +597,17 @@ def strip_url_scheme(url):
...
+
+def is_prod_server():
+    """Returns True if this is the production server.
+
+    This is not the simple opposite of is_dev_app_server (above), because it
+    only returns True if this is "the" prod server (i.e., the one that serves
+    google.org/personfinder). It won't return True for other servers that happen
+    to be running on App Engine (e.g., people's dev instances).
+    """
+    return os.environ['APPLICATION_ID'] == 'googlepersonfinder'
```
gimite: Since we allow third parties to run their own Person Finder, this hard-coded condition is not 100% accurate.

I'm OK with this for now, but can you leave TODO to improve it in the future? Maybe add a config to enable sitemap ping (default to False) or more generally introduce a config for "production mode"?

nworden: Done with a config setting for enable pinging. That way it can also be useful for situations where you wouldn't want even the production server to ping index servers (like if you wanted to run a somewhat private PF).

gimite: Thanks! Can you document the config at:
https://github.com/google/personfinder/wiki/DeploymentGuide#step-8-configure-the-app-for-production-service
https://github.com/google/personfinder/wiki/Config#list-of-configs

It would be nicer if the config is added to the admin page too, but it can be done later.

nworden: Updated the docs, and filed issue #556 for adding the setting to the admin page.

PR #553, opened 2019-03-03
Set up Python 3 testing. Fixes #508.

### LINKED ISSUE #508: Set up Python 3 testing
Most of our tests will fail right now if we try to run them with Python 3, so we'll need to set up something that runs a specific subset of tests; we can add to that set as we make more code Python 3-compatible.

nworden:
Just includes test_text_query and test_jautils for now. I'll add more files to the PY3_TEST_FILES list as I get them working with Python 3.

I didn't add `app/jautils.py` to `tools/python3_compatibility_check` because it produces a false positive (running `python-modernize -w` produces no changes); it seems to be [a known bug](https://stackoverflow.com/questions/40240229/2to3-says-no-changes-needed-then-files-that-need-to-be-modified) related to strings with escapes. That's likely to be an issue with a bunch of our files, so I might consider dropping the python3 compatibility check (it's kinda redundant with the Python 3 tests), but that's a question for another pull.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: python
 
 python:
   - 2.7
+  - 3.7
 
 services:
   - docker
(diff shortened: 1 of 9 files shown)


nworden:
I'll send this to someone in NYC for review.

PR #554, opened 2019-03-04
 Remove external_search_backends feature.

nworden:
Hiroshi suggested deprecating this in the discussion on pull #491: it
hasn't been used since 2011, and he checked that it's not enabled in
prod.


nworden:
Sent to Avi.

PR #555, opened 2019-03-04
XSRF tool.

nworden:


nworden:
Christophe already reviewed this separately.

app/utils.py
```diff
@@ -1165,3 +1167,36 @@ def delete(self, *args):
...
+
+# ==== XSRF protection =========================================================
+
+class XsrfTool(object):
+
+    # XSRF tokens expire after 4 hours.
+    TOKEN_EXPIRATION_TIME = 60 * 60 * 4
+
+    def __init__(self):
+        self._key = config.get('xsrf_token_key').encode('utf-8')
```
gimite: Can you add initialization logic of this config somewhere so that new instances will have a valid key? e.g., Initialize it with a random string when it is not set.

app/admin_delete_record.py
```diff
@@ -31,10 +31,22 @@ class Handler(BaseHandler):
     admin_required = True
 
     def get(self):
+        xsrf_tool = XsrfTool()
```
gimite: No concrete idea, but I wonder if it's possible to share this logic between different admin pages instead of repeating it in all pages (in a common base class of admin pages?). It would also make sure that the logic is also applied to any new admin pages in the future. Or at least report an error when the page does not apply XSRF tool.

app/utils.py
nworden: Yeah, good idea. Done with the tests in PR #557.

app/admin_delete_record.py
nworden: Yeah, I agree that'd be useful. It might make sense to do it as part of the migration off webapp2 (which I just realized a couple days ago doesn't support Python 3), since that'll involve a lot of changes/refactoring of the handlers anyway.

gimite: Sounds good.

PR #557, opened 2019-03-05
Tests for XSRF tool

nworden:
Also sets a token in `XsrfTool.__init__` if none exists (suggested by Hiroshi on PR #555), and includes XSRF tokens in the API Key view table (which the server expects to see but currently doesn't get).

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -26,6 +26,7 @@ import logging
 import os
 import random
 import re
+import string
 import sys
 import time
 import traceback
@@ -1177,7 +1178,16 @@ class XsrfTool(object):
     TOKEN_EXPIRATION_TIME = 60 * 60 * 4
 
     def __init__(self):
-        self._key = config.get('xsrf_token_key').encode('utf-8')
+        configured_key = config.get('xsrf_token_key')
+        if configured_key:
+            # config.get returns unicode, but hmac is going to want a str
+            self._key = configured_key.encode('utf-8')
+        else:
+            configured_key = ''.join([
+                random.choice(string.letters + string.digits)
+                for i in range(20)])
+            config.set(xsrf_token_key=configured_key)
+            self._key = configured_key
 
     def generate_token(self, user_id, action_id):
         action_time = get_utcnow_timestamp()
--- a/app/utils.py
+++ b/app/utils.py
@@ -1177,6 +1177,9 @@ class XsrfTool(object):
     # XSRF tokens expire after 4 hours.
     TOKEN_EXPIRATION_TIME = 60 * 60 * 4
 
+    # Characters with which to choose a key if it's not set yet.
+    TOKEN_CHARACTER_SET = string.letters + string.digits
+
     def __init__(self):
         configured_key = config.get('xsrf_token_key')
         if configured_key:
@@ -1184,8 +1187,7 @@ class XsrfTool(object):
             self._key = configured_key.encode('utf-8')
         else:
             configured_key = ''.join([
-                random.choice(string.letters + string.digits)
-                for i in range(20)])
+                XsrfTool.TOKEN_CHARACTER_SET for _ in range(20)])
             config.set(xsrf_token_key=configured_key)
             self._key = configured_key
 
--- a/tests/test_xsrftool.py
+++ b/tests/test_xsrftool.py
@@ -1,0 +1,29 @@
+import datetime
+import unittest
+
+import config
+import utils
+
+
+class XsrfToolTests(unittest.TestCase):
+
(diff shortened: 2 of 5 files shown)


app/utils.py
```diff
@@ -1177,7 +1178,16 @@ class XsrfTool(object):
...
     def __init__(self):
-        self._key = config.get('xsrf_token_key').encode('utf-8')
+        configured_key = config.get('xsrf_token_key')
+        if configured_key:
+            # config.get returns unicode, but hmac is going to want a str
+            self._key = configured_key.encode('utf-8')
+        else:
+            configured_key = ''.join([
+                random.choice(string.letters + string.digits)
+                for i in range(20)])
```
christopherestif: depending on how fancy you want to be, I'd suggest:

option a: no [] and no i:
ALPHANUMERIC = string.letters + string.digits
def __init....
configured_key = ''.join(
                random.choice(ALPHANUMERIC)
                for _ in range(20))

option b: no for loop and yet another library:
ALPHANUMERIC = list(string.letters + string.digits)
def __init....
  configured_key = ''.join(numpy.random.choice(ALPHANUMERIC, 20, replace=True))

tests/test_xsrftool.py
```diff
@@ -0,0 +1,50 @@
...
+        tool = utils.XsrfTool()
+        timestamp = utils.get_timestamp(XsrfToolTests.TEST_NOW)
+        self.assertFalse(tool.verify_token(
+            'NotTheRightDigest/%f' % timestamp, 12345, 'test_action'))
+
+    def testRejectsExpiredToken(self):
+        config.set(xsrf_token_key='abcdef')
+        tool = utils.XsrfTool()
+        token = tool.generate_token(12345, 'test_action')
+        utils.set_utcnow_for_test(datetime.datetime(2010, 1, 31, 22, 1, 0))
```
christopherestif: if you want to save the reader the math homework:
utils.set_utcnow_for_test(TEST_NOW + datetime.timedelta(hours=4, minutes=1))

app/utils.py
nworden: Thanks, good points! We're not using numpy for anything else so I went with option A.

tests/test_xsrftool.py
nworden: Done.

tests/test_xsrftool.py
```diff
@@ -0,0 +1,50 @@
...
+    TEST_NOW = datetime.datetime(2010, 1, 31, 18, 0, 0)
+
+    def setUp(self):
+        utils.set_utcnow_for_test(XsrfToolTests.TEST_NOW)
+
+    def testGenerateAndVerifyGoodToken(self):
+        config.set(xsrf_token_key='abcdef')
+        tool = utils.XsrfTool()
+        token = tool.generate_token(12345, 'test_action')
+        self.assertTrue(tool.verify_token(token, 12345, 'test_action'))
```
christopherestif: just in case it's relevant:
self.assertIs(True, <expr>) actually tests that <expr> is True
whereas
self.assertTrue(<extr>) tests that bool(<expr>) is True

so if you want to detect functions suddenly returning [] instead of False, you can use assertIs(True/False)
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertTrue

nworden: Thanks, that's good to know, but I don't think it's a concern here: the XSRF tool's job is just to give a thumbs-up/thumbs-down (even if it did start returning an empty list or something, its callers would still interpret it the same way the test does).

app/utils.py
```diff
@@ -1176,8 +1177,19 @@ class XsrfTool(object):
...
+
     def __init__(self):
-        self._key = config.get('xsrf_token_key').encode('utf-8')
+        configured_key = config.get('xsrf_token_key')
+        if configured_key:
+            # config.get returns unicode, but hmac is going to want a str
+            self._key = configured_key.encode('utf-8')
+        else:
+            configured_key = ''.join([
+                XsrfTool.TOKEN_CHARACTER_SET for _ in range(20)])
```
gimite: Maybe you forgot to call random choice function here? Looks like this always generates the same key, 20 time repeat of 'abc...789'. Also, consider using utils.generate_random_key instead:
https://github.com/google/personfinder/blob/17f793f3fee56eb6b5cdfa6027d5262d2d8e12de/app/utils.py#L615

nworden: Whoa, yeah, you're right, I dropped random.choice by accident when I set up the character set constant. Thanks for catching that! I changed it to call generate_random_key (PR #559).

PR #559, opened 2019-03-06
Change default XSRF key generation.

nworden:


app/utils.py
```diff
@@ -1188,8 +1188,7 @@ def __init__(self):
             # config.get returns unicode, but hmac is going to want a str
             self._key = configured_key.encode('utf-8')
         else:
-            configured_key = ''.join([
-                XsrfTool.TOKEN_CHARACTER_SET for _ in range(20)])
```
gimite: Remove definition of TOKEN_CHARACTER_SET which is no longer used.

nworden: Done in PR #562.

PR #560, opened 2019-03-06
Remove user arg from call to reveal.html template.

nworden:
As suggested in the comment, someone already did put a Turing test
there instead. The user param is no longer used.


nworden:
Sent to Avi.

PR #561, opened 2019-03-07
Add more fields to the view page

gimite:


ui/src/pages/View.js
```diff
@@ -164,8 +194,11 @@ class View extends Component {
   }
 
   renderField(labelMessage, value) {
+    if (value == null || value === '') {
+      return null;
+    }
     return (
-      <div>
+      <div key={labelMessage.id}>
```
nworden: What is this key used for? It feels a little awkward to me to use the message ID as a key, and it might mean that they're not always unique: for example, if someone had multiple Twitter accounts, the Twitter message ID would be used as the key for two of these.

ui/src/pages/View.js
```diff
@@ -179,12 +212,30 @@ class View extends Component {
...
       return this.props.intl.formatMessage(
           SEX_VALUE_MESSAGES[this.state.person.sex]);
     }
   }
+  
+  getSourceDateValue() {
+    if (this.state.person.source_date == null) {
+      return null;
+    } else {
+      return new Date(this.state.person.source_date).toLocaleString();
```
nworden: We might want to consider doing this formatting server-side and returning a string from the FE API instead of a raw timestamp. If someone selects a language in PF that's different from what their browser thinks they want, I think it'd be better if our date formatting matched the rest of what we serve them.

ui/src/pages/View.js
```diff
@@ -234,30 +285,41 @@ class View extends Component {
...
           </div>
 
-          <div className='mdc-typography--body1'>
-            <h2 className='mdc-typography--subtitle2'>
-              <FormattedMessage {...MESSAGES.profilePages} />
-            </h2>
-            {/* TODO(gimite): Implement this. */}
-            <div>TBD</div>
-          </div>
+          {this.state.person.profile_pages.length > 0 &&
```
nworden: Is this a common thing to do in JS (to use && in place of an if statement)?

ui/src/Utils.js
```diff
@@ -26,4 +51,15 @@ class Utils {
   }
 }
 
+/**
+ * A mapping from a profile page site ID to a MessageDescriptor of the name of
+ * the site.
+ */
+Utils.PROFILE_PAGE_SITES = Object.freeze({
```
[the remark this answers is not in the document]
nworden: SGTM!

ui/src/pages/View.js
nworden: Ok, cool.

gimite:
@nworden Do you have any further comments, or is this approved?

gimite:
> Oh, sorry for the delay, I thought you meant you were going to create the utils directory in this PR.

Oh I see, I have forgotten about that part. I will make the change.

PR #563, opened 2019-03-08
Set Content Security Policy for the React UI.

nworden:
Using the policy from [this example](https://csp.withgoogle.com/docs/strict-csp.html).

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -104,6 +104,7 @@ class MainTests(unittest.TestCase):
             assert 'Content-Security-Policy' in response.headers
             assert ('nonce-totallyrandomkey' in
                     response.headers['Content-Security-Policy'])
+            assert 'nonce="totallyrandomkey"' in response.body
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


tests/test_main.py
```diff
@@ -91,6 +91,16 @@ def test_default_language(self):
...
 
+    def test_content_security_policy_for_react(self):
+        """Verify CSP is set when the React UI is enabled."""
+        config.set(enable_react_ui=True)
+        request = setup_request('/')
+        response = webapp.Response()
+        handler = main.Main(request, response)
+        handler.get()
+        assert 'Content-Security-Policy' in response.headers
+        assert 'nonce-' in response.headers['Content-Security-Policy']
```
cyrusboadway: It might be worthwhile to assert the full nonce in the header matches the response body, since that's what the browser will work against. Perhaps also verify the nonce is not just `"nonce-"`.

nworden: Thanks, done. Since mocking in Python is (usually) ridiculously easy I mocked out generate_random_key and have it verifying the specific value.

PR #568, opened 2019-03-12
Cease collecting and displaying street addresses

nworden:
We'll keep it as part of the API, to be true to the PFIF standard, but we won't collect street addresses ourselves nor display them.


app/locale/af/LC_MESSAGES/django.po
```diff
@@ -1371,9 +1371,6 @@ msgstr "Status van persoon"
 msgid "Status of this person:"
 msgstr "Status van hierdie persoon:"
 
-msgid "Street name"
-msgstr "Straatnaam"
-
 msgid "Street name only, no number"
```
christopherestif: just checking, should we also remove "Street name only, no number" ?

nworden: Yeah, that string was a hint for the street name field, so we don't need it anymore.

PR #572, opened 2019-03-12
Python 3 tests, in shared Docker container.

### LINKED ISSUE #508: Set up Python 3 testing

nworden:
Similar to PR #553, but does everything within a single Docker container rather than trying to have Travis handle the different Python versions (which didn't work because Docker's actually the one deciding what version(s) of Python get installed).
Fixes #508.


nworden:
(sigh) this worked locally. I'll close this PR until I figure it out.

PR #573, opened 2019-03-13
Use Django utilities to validate URLs and emails.

nworden:
Three changes:
- I was wrong last week: it turns out Django has an email validation thing these days, so use that.
- Use Django to validate URLs before display (in `sanitize_urls`, which is already called by the various view functions) instead of our own function.
- Also validate URLs when records are created.


nworden:
Assigning to Eliah. I'm not sure I've sent you a file with django.po files before, but, just so you know, they're automatically generated.

app/create.py
```diff
@@ -102,6 +103,21 @@ def post(self):
         expiry_date = days_to_date(self.params.expiry_option or
                                    self.config.default_expiry_days)
 
+        profile_urls = []
+        if self.params.profile_url1:
+            profile_urls.append(self.params.profile_url1)
+        if self.params.profile_url2:
+            profile_urls.append(self.params.profile_url2)
+        if self.params.profile_url3:
+            profile_urls.append(self.params.profile_url3)
```
eliahhecht: Maybe you could do this more concisely with a list comprehension? Something like:

`profile_urls = [url for url in [self.params.profile_url1, self.params.profile_url2, self.params.profile_url3] if url]`

app/utils.py
```diff
@@ -293,8 +294,7 @@ def validate_image(bytestring):
         return False
 
 
-EMAIL_PATTERN = re.compile(r'(?:^|\s)[-a-z0-9_.%$+]+@(?:[-a-z0-9]+\.)+'
-                           '[a-z]{2,63}(?:\s|$)', re.IGNORECASE)
+EMAIL_VALIDATOR = EmailValidator()
```
eliahhecht: Do we still need this field?

app/utils.py
```diff
@@ -382,16 +378,34 @@ def get_app_name():
...
+        except ValidationError:
             setattr(record, field, None)
     # Multi-line URLs.
     for field in ['profile_urls']:
         urls = (getattr(record, field, None) or '').splitlines()
-        sanitized_urls = [url for url in urls if url and url_is_safe(url)]
+        sanitized_urls = []
+        for url in urls:
+            if not url:
+                continue
```
eliahhecht: not sure if this is a python thing but in general, I think I would prefer:
```py
for url in urls:
  if url: 
   # validate the url
```

nworden: It's not a Python thing, changed.

app/utils.py
nworden: We need EMAIL_VALIDATOR, yes. It's what we're using in place of the regex.

app/create.py
nworden: I got fancy and used `filter`.

PR #578, opened 2019-03-14
Code for moving data files

ashutoshnarayan:
Code for moving data files from app/ directory


ashutoshnarayan:
@gimite I have raised the PR #578  please review

app/admin_resources.py
```diff
@@ -20,7 +20,7 @@
 import mimetypes
 
 import config
-import const
+from utils import const
```
nworden: Did you intend to include all these changes related to const.py in this PR (the title says it's just for moving the data files)?

app/locale/af/LC_MESSAGES/django.po
```diff
@@ -695,6 +695,10 @@ msgstr "Vergelyk hierdie rekords"
 msgid "Contact information"
 msgstr "Kontakinligting"
 
+#, python-format
+msgid "Content purported to be compressed with %s but failed to decompress."
```
nworden: I'm not sure exactly why, but it looks like messages from files in `app/vendors` is getting added to the django.po files. That's not necessary -- we only need our own messages in here.

app/config.py
```diff
@@ -19,6 +19,9 @@
 See here for usage examples:
 https://github.com/google/personfinder/wiki/DeveloperFaq
 """
+import sys
```
nworden: I'm not sure exactly what this is for (?), but it shouldn't be necessary to modify `sys.path` within our code.

tests/test_importer.py
```diff
@@ -16,13 +16,18 @@
 
 import datetime
 import unittest
+# added by Ashutosh Narayan
```
nworden: Although there are names from the earlier Person Finder authors still in the codebase, in recent years we haven't been adding individual names to the source code (I think I've only seen gimite's name once or twice, and I've never added my name anywhere); the commit log is the record of who did what. If you'd like us to start an `AUTHORS` file, feel free to open an issue for that and assign it to me (Google might have rules about how exactly I do it, I'd have to check), but I don't think it's practical to have names inline in the source considering how many people have contributed different pieces over time.

app/config.py
ashutoshnarayan: @nworden sys.path was present since directory named handlers and utils were having config.py in it and it was supposed to be imported.

tests/test_importer.py
ashutoshnarayan: @nworden I have removed name from source code ; it was added as part of testing to remove later if there are any failures due to lines appended by me. Yes, we can have an AUTHORS file

app/admin_resources.py
ashutoshnarayan: @nworden Initially in PR #450 ( closed ) I moved files related to handlers and utils into respective directories. This PR actually reverts them to app/ and copies only contents related to app/data/

app/locale/af/LC_MESSAGES/django.po
ashutoshnarayan: Even am trying to understand how messages are getting added to django.po files. Neither any of regular expression in my code is appending so. What exactly is the ask here ?

app/admin_resources.py
ashutoshnarayan: > Did you intend to include all these changes related to const.py in this PR (the title says it's just for moving the data files)?

@nworden It was actually the part of handlers which were supposed to be moved from app directory. Now since handlers are kept under app/ we can keep const.py here itself.

ashutoshnarayan: > The [diffs I'm looking at](https://github.com/google/personfinder/pull/578/files) are your current PR compared to our master branch, and there are a lot of changes unrelated to the data files. In particular, in many places it changes imports of `const` from `import const` to `from utils import const` (along with changes to sys.path and the django.po files).

@nworden I have reverted that change to import const

app/config.py
```diff
@@ -19,12 +19,13 @@
...
 
 from google.appengine.ext import db
 import UserDict, model, random, simplejson
 import logging
 import datetime
-import utils
+#import utils
```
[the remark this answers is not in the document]
ashutoshnarayan: @nworden have uncommented it in this commit

tests/test_utils.py
```diff
@@ -16,6 +16,10 @@
 
 """Tests for utils."""
 
+import sys
```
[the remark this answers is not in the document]
ashutoshnarayan: @nworden have removed sys path in this commit

PR #582, opened 2019-03-15
Python 3 testing.

nworden:
This sets up Python 3 testing for both local development and Travis CI, for the (currently quite limited) set of tests that we expect to pass with Python 3 (listed in `tests/unit_tests.py`).
As part of doing this, I changed the way our Travis tests run by taking them out of Docker and running on Travis directly. As far as I know there wasn't a particular reason to run them in a Docker container; I guess once you have a Docker setup it might just be the easiest path to Travis tests, but it's kinda limiting if you need to run tests in multiple environments. We'll keep the Docker setup around though (it's useful for other things).
Also, minor change: we no longer specify the version of pytest to install, so pip will just get the latest version. We might be able to do the same with Pillow later, but I haven't looked into it (App Engine might be using an older version, so maybe we should be too).
Fixes issue #508.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
       unzip
   - wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.62.zip -O /tmp/appengine.zip
   - unzip -qq /tmp/appengine.zip
-  - pip install cssselect lxml mock modernize pillow==4.1.0 pytest
+  - pip install --user cssselect lxml mock modernize pillow==4.1.0 pytest
   - npm --prefix ui install
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 sudo: required
 
+language: python
+
 matrix:
   include:
   - name: "Python 2 tests"
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
 
 install:
   - |
-    apt-get install -y \
+    sudo apt-get install -y \
       build-essential \
       curl \
       gettext \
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -2,26 +2,31 @@ sudo: required
...
-
-before_install:
-  - sudo rm /usr/local/bin/docker-compose
-  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
-  - chmod a+x docker-compose
-  - sudo mv docker-compose /usr/local/bin
-  - docker-compose up -d
-
-before_script:
+matrix:
```
christopherestif: knock knock, Neo...

tools/common.sh
```diff
@@ -37,6 +37,18 @@ for python in \
     fi
 done
 
+for python3 in \
+    "$PYTHON3" \
```
christopherestif: in general, I don't know if we care to use curly brackets for variables (like ${var}) consistently or not

tools/py3_unit_tests
```diff
@@ -0,0 +1,19 @@
...
+
+# To run all tests we expect to pass with Python 3:
+#
+#     tools/py3_unit_tests
+#
+# To run just the tests in tests/test_model.py and tests/test_indexing.py:
+#
+#     tools/py3_unit_tests test_model test_indexing
+
+pushd "$(dirname $0)" >/dev/null && source common.sh && popd >/dev/null
```
christopherestif: do we care about a "|| exit 1" or something special here?

nworden: I think it's ok as-is. This is consistent with its Python 2 counterpart (`tools/unit_tests`), and, if `common.sh` doesn't get run for whatever reason, it'll fail anyway, because `common.sh` is what adds a bunch of dependencies (most notably the App Engine SDK) to the Python path.

tools/common.sh
nworden: It works like this; this code is basically the same some code above which sets the $PYTHON variable for Python 2.

.travis.yml
nworden: :)

Issue #575, opened 2019-03-13
Upgrade WebOB (and webapp2 to be safe)


nworden:
Our current Docker setup is running Python v2.7.12 (because that's what's in the Ubuntu Xenial repository). It breaks with Python 2.7.15 (I haven't tried versions in between).

The reason is that the newer version of `xml.sax.expatreader.ExpatParser` (used by `app/pfif.py`) expects to be able to call `close` on its input file, and the input object provided by WebOB v1.1.1 (a `LimitedLengthFile`) doesn't have a `close` function. It does have that function in WebOB v1.2.3, so we should upgrade to that. It might be a good idea to upgrade to webapp2 v2.5.2 at the same time (since they're closely linked it seems better to run versions from similar times).

nworden:
Once I get things working with Python 2.7.15, I'll see if I can set up Travis to run with two environments: one that uses 2.7.12 (because that seems to be [what App Engine currently uses](https://cloud.google.com/appengine/docs/standard/python/release-notes) for its 2.7 runtime) and one for the latest thing I can find (to be safe if/when App Engine upgrades its runtime).

nworden:
TL;DR for @gimite: is there a partner or anyone else you know of who's using the `charsets` URL parameter?

This is turning out to be a big can of worms, because of the way WebOB 1.2.3 supports non-UTF8 requests. You can't just assign the charset field anymore, you need to manually call `request.decode()`, which doesn't sound so bad except `decode()` doesn't modify the request in place; instead it builds a brand new request and returns that. If your original request object is actually a subclass of the WebOB `BaseRequest` class (say, a webapp `Request` object), any fields from the subclass (which WebOB's `BaseRequest.decode` doesn't know about) get lost. webapp uses at least one of those (the `route` field) for critical functionality.

webapp isn't being maintained, so it's never going to call `Request.decode` in the right place in its own code to deal with it under the hood (which I think is the theoretically correct way this should be handled).

This isn't a long-term problem for us since we're abandoning webapp2 anyway, but it does leave us in a bad position in the meantime. I think there's three options:
1. Do nothing. Hope that App Engine doesn't start using Python 2.7.15 before we've migrated to Django.
2. Abandon support for non-UTF8 requests to make it easier to move to WebOB 1.2.3 (I think we could still support non-UTF8 responses if we wanted).
3. Basically fork webapp2 to address the WebOB change ourselves.

[...]

PR #583, opened 2019-03-15
Run Travis with both Python 2.7.12 and a recent 2.7.

### LINKED ISSUE #576: Include latest (of 2.7.x) version of Python in Travis tests
Spun off of issue #575.
Currently we test on an older version of Python 2.7, just by default based on what's in the Ubuntu package repository. That's ok, because [right now App Engine uses 2.7.12](https://cloud.google.com/appengine/docs/standard/python/release-notes) anyway, which was released in 2016. However, it'd be good to also test with the latest 2.7.x version, in case App Engine upgrades its 2.7 runtime.

nworden:
We recently discovered that we would break with Python 2.7.15 (see issue
#575). We'd been ok because App Engine uses Python 2.7.12 for its Python
2.7 runtime. However, it seems possible that they could upgrade to a
more recent version at any time, without giving us a heads-up (since no
one generally expects a patch version to be a breaking change). So, I
want to run our Travis tests with both Python 2.7.12 (because it's what
we currently get in prod) and a recent version (because we might get
something more recent in prod at any time).
Fixes #576.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,14 @@ language: python
 
 matrix:
   include:
-  - name: "Python 2 tests"
+  # As of 2019-03-15, App Engine's Python 2.7 runtime uses 2.7.12.
+  - name: "Python 2.7.12 tests"
+    python: 2.7.12
+  # To be safe if/when App Engine upgrades, also run tests on a recent version
+  # of Python 2.7.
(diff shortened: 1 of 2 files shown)


nworden:
Sent to Eliah.

PR #584, opened 2019-03-17
Stop using jQuery.

nworden:
We're not going to use jQuery with the React UI, and it seems like
overkill to use it for a lite mode, and I'd rather not load it if we
don't have to (especially since we have to avoid using a CDN to support
zero-rating). We're also not really using it much currently anyway: I
started seeing what it would take to upgrade to jQuery 3, and it turned
out to be easier to just rewrite the jQuery parts to use vanilla JS.

This does mean losing the sliding animations for opening sections on
create/add_note pages (e.g., the "more information to identify the
person" section), but that seems ok, since I'm not sure we'd really want
JS animations in lite mode. (If we really wanted to keep them though,
we could rewrite them in vanilla JS as well.)


app/resources/results.html.template
```diff
@@ -46,16 +46,20 @@
...
-            url: repo_url + '/third_party_search',
-            data: {
-                query: query,
-                lang: lang,
-                query_type: query_type,
-                search_engine_id: id
-            },
-            dataType: 'json'
-        }).done(function(r) {
+        fetch(repo_url + '/third_party_search', {
```
gimite: I'm wondering if it is safe to use fetch() and ES6 syntax e.g., `=>` here because they are relatively new and not supported by all browsers:
https://caniuse.com/#feat=promises,fetch
https://caniuse.com/#search=ecmascript
and we currently do not use ES6 here (except for the new React code which is transpiled). It might be OK since they are supported by all modern browsers. But may be safer to avoid them and use good old XMLHTTPRequest and `function(){}`? What do you think?

One alternative is to put this (removing jQuery) on hold until we launch the new React UI. Then this kind of HTML templates are only used by Lite mode and AMP where JS is disabled. So we can simply remove all JS code in HTML templates.

app/resources/results.html.template
```diff
@@ -64,15 +68,18 @@
                   'not a string');
               return;
             }
-            jQuery('#third_party_search_result_link' + id).attr('href', r.url);
-            jQuery('#third_party_search_result' + id).show();
+            document.querySelector(
+                '#third_party_search_result_link' + id).href = r.url;
+            document.querySelector(
+                '#third_party_search_result' + id).style.display = 'inline';
```
gimite: Should be 'block' because it's a P tag.

app/resources/results.html.template
gimite: Have you checked third party search works fine after your change, just in case? If not, I can probably help you to test it since it's a bit tricky to set up an environment to test it...

app/resources/static/forms.js.template
```diff
@@ -367,22 +367,25 @@ function validate_fields() {
...
-  var expandedPart = jQuery(expandedSection).find("div.expanded-part");
-  var expandButton = jQuery(expandedSection).find("a.expand-button");
-
-  expandedPart.hide();
-  expandButton.click(function() {
-    if (expandedPart.is(":visible")) {
-      expandedPart.slideUp("slow");  
-      expandButton.removeClass(
-        'expand-button-icon-up').addClass('expand-button-icon-down');
+function make_form_expandable(expandedSectionSelector) {
```
gimite: Have you confirmed that this part works fine after your change?

app/resources/results.html.template
```diff
@@ -46,16 +46,20 @@
...
-            data: {
-                query: query,
-                lang: lang,
-                query_type: query_type,
-                search_engine_id: id
-            },
-            dataType: 'json'
-        }).done(function(r) {
+        fetch(repo_url + '/third_party_search', {
+          method: 'POST',
```
gimite: Just curious, why did you change this from GET to POST?

app/third_party_search.py
```diff
@@ -29,7 +29,7 @@
 
 class Handler(utils.BaseHandler):
     """Proxy to perform search with third-party search engine."""
-    def get(self):
+    def post(self):
```
gimite: Since you changed the JS code to send the parameters as JSON in POST body instead, I guess you also need to change parameter extraction logic here e.g., self.params.query?

app/resources/static/forms.js.template
nworden: Yes.

app/resources/results.html.template
nworden: Done.

app/resources/results.html.template
nworden: Good point, changed to XMLHTTPRequest.
I think it's worth doing the work to get off jQuery now, since we're not sure when the new UI will launch. Also maybe we can consider having JS in lite mode as long as it's not critical functionality (e.g., the expanding/collapsing sections could still be used in lite mode as long as they're expanded by default when there's no JS).

app/resources/results.html.template
nworden: I got some kind of error (can't remember what it was) trying to send the data with fetch as a GET request. Doesn't seem to be an issue with XMLHTTPRequest though (plus now that I think about it it doesn't really make sense for it to be a POST request), so changed it back to GET.

app/resources/results.html.template
nworden: I tested the frontend changes, but I did it by hacking the backend to return hard-coded values. Is it easy to find a query that will return results (once you configure it to point at the real third-party) or do you test it some other way?

app/resources/results.html.template
```diff
@@ -46,37 +46,33 @@
...
-          },
-        })
-        .then(r => r.json())
-        .then(function(r) {
-          console_log('Third-party search response: ', r);
-          if (r.status == 'OK') {
-            if (typeof(r.url) != 'string') {
-              console_error(
+        var req = new XMLHttpRequest();
+        const url = (
```
gimite: May be better to avoid using `const` as well because it is part of ES6.

app/resources/results.html.template
gimite: That would probably be sufficient, thanks! "安否　一郎" is a sample query which always returns the result, so we can try the query after pushing this change in production.

app/resources/results.html.template
```diff
@@ -46,37 +46,33 @@
...
+            repo_url + '/third_party_search?query=' + encodeURIComponent(query)
+            + '&lang=' + encodeURIComponent(lang) + '&query_type='
+            + encodeURIComponent(query_type) + '&search_engine_id='
+            + encodeURIComponent(id));
+        req.open('GET', url);
+        req.setRequestHeader(
```
gimite: Since you switched back to a GET request, I guess you don't need Content-type here?

nworden: Done.

app/resources/results.html.template
nworden: Oh, I hadn't realized that. Changed to var.

nworden:
Thanks for the sample query! I'll be sure to try it when I do the release.

PR #587, opened 2019-03-18
Store photos from user-provided URLs

nworden:
Currently, if a user wants to provide a photo, they can either upload one or give us a link. In the latter case, we'll hotlink it. This changes it to pull the image from the given URL and store it in our DB, same as we do for uploaded photos. I'll set up a cron job or something later to do the same with image links we get from API partners.

The reasons for this change are that:
- Hotlinking images is a problem for zero-rating (which requires that everything be served off the same domain).
- It lets us shrink large images and generate thumbnails for the results page.
- The current arrangement allows a way around moderation -- we wouldn't want people changing images after the record's been reviewed.

### DIFF
--- a/app/photo.py
+++ b/app/photo.py
@@ -54,6 +54,15 @@ def create_photo_with_url(handler, photo_upload, photo_url):
     return a Photo object and a URL with which to serve it.
 
     If neither parameter is provided, returns (None, None).
+
+    Args:
+      handler: a request handler (needed to generate URLs)
+      photo_upload: optional; an images.Image for an uploaded photo
+      photo_url: optional; a user-provided URL for a photo
+
+    Returns:
+      A tuple with an Image and URL with which to serve it, or (None, None) if
+      neither an uploaded photo nor a URL was provided.
     """
     if photo_upload is not None:
         return create_photo(photo_upload, handler)
--- a/app/photo.py
+++ b/app/photo.py
@@ -16,6 +16,8 @@
 """Handler for retrieving uploaded photos for display."""
 
 import os
+import requests
+import requests_toolbelt.adapters.appengine
 
 import model
 import utils
@@ -23,6 +25,11 @@ import utils
 from django.utils.translation import ugettext_lazy as _
 from google.appengine.api import images
 from google.appengine.runtime.apiproxy_errors import RequestTooLargeError
+
+# Use the App Engine Requests adapter. This makes sure that Requests uses
+# URLFetch.
+# TODO(nworden): see if we should condition this on the runtime (Python 2 vs. 3)
+requests_toolbelt.adapters.appengine.monkeypatch()
 
 MAX_IMAGE_DIMENSION = 300
 MAX_THUMBNAIL_DIMENSION = 80
--- a/app/photo.py
+++ b/app/photo.py
@@ -55,7 +55,7 @@ def create_photo_with_url(handler, photo_upload, photo_url):
 
     If neither parameter is provided, returns (Note, None).
     """
-    if photo_upload:
+    if photo_upload is not None:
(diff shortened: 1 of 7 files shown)


app/photo.py
```diff
@@ -40,6 +47,24 @@ class SizeTooLargeError(PhotoError):
                 'Please upload a smaller one.')
 
 
+def create_photo_with_url(handler, photo_upload, photo_url):
```
cyrusboadway: The design might be more general-purpose to have two methods, or one method with only the handler param and an 'image' param and do some kind of type test to determine which flow to follow.

Python's weird. Parametric polymorphism would be nice.

Also, perhaps a pydoc explaining what hander is, because it wasn't clear to em.

nworden: Added pydoc.
The reason for having two separate parameters is that all the callers will have variables for an uploaded photo and a URL, but won't know which (if either) is actually populated. I'd prefer not to have a bunch of duplicated code checking that stuff and calling the appropriate method (I'd already need two sets of it in create.py alone, for the note and record photos). So I think it does make sense to have that logic in a function, but maybe it doesn't belong in `photo.py`. What if I moved that part into `create.py`, so that `photo.py` was just about photos and `create.py` was responsible for sorting out user input and sending it to the right `photo.py` function?

cyrusboadway: Yea, that makes sense. A little helper in `create.py` and the generic functions in `photos.py` makes sense to me.

nworden: Done. The helper in create.py is outside the class so it can be used by other things that need to create stuff later -- the various places that create things should probably share a bunch of code, but that's a refactor for another day, so for now I put it in create.py with a todo.

app/photo.py
```diff
@@ -40,6 +47,20 @@ class SizeTooLargeError(PhotoError):
                 'Please upload a smaller one.')
 
 
+def create_photo_from_url(handler, photo_url):
```
cyrusboadway: `create_photo` and `create_photo_from_url` have `handler` in different positions. For consistency consider swapping to handler always first.

nworden: Done.

app/photo.py
```diff
@@ -40,6 +47,20 @@ class SizeTooLargeError(PhotoError):
...
 
 
+def create_photo_from_url(photo_url, handler):
+    """Creates a photo from a URL.
+
+    Returns:
+      A tuple with an Image and URL with which to serve it, or (None, None) if
+      the image is invalid.
+    """
+    response = requests.get(photo_url)
```
gimite: Just curious:

* Maybe we should set the user agent header to something appropriate here?
* Maybe we should check robots.txt to see if the photo allows fetch by robots?

nworden: Are robots.txt files applicable to things other than crawlers? Particularly since we're essentially fetching the file on behalf of a human user, I'm not sure we need to worry about robots.txt files (?).
Any suggestions on the user agent? Can we just make one up (i.e., something with "PersonFinder" in the name)?

gimite: > Are robots.txt files applicable to things other than crawlers? Particularly since we're essentially fetching the file on behalf of a human user, I'm not sure we need to worry about robots.txt files (?).

I'm not really sure about it, but probably you are right:

http://www.robotstxt.org/faq/what.html
> A robot is a program that automatically traverses the Web's hypertext structure by retrieving a document, and recursively retrieving all documents that are referenced.

Since we are not **traversing** or **recursively** retrieving documents.

> Any suggestions on the user agent? Can we just make one up (i.e., something with "PersonFinder" in the name)?

Yeah I guess we can just make up. Usually they contain the URL of the site e.g.:

Google Person Finder (https://google.org/personfinder)

Issue #349, opened 2017-03-29
Consider putting all configs into a single entry


gimite:
Currently we put each config (key-value pair) into a single ConfigEntry entry in datastore:
https://github.com/google/personfinder/blob/master/app/config.py#L123

This causes a lot of datastore access. We should probably consider putting all configs for a repository into a single entry, which should reduce the number of datastore access significantly, and make the app load faster.

Note that datastore access time may be significantly improved by #346 which will use memcache internally. But I believe this change is still preferred after #346 because it can make the app load even faster.

Note that there should be a limit of the size of a single entry (1MB?). But config for a single repository should fit in the limit, though we need to verify. We should probably keep separating the entry per repository, because there can be many repositories over time and the total size of config in all repositories might exceed 1MB.

waseem18:
@gimite I'll look into it along with #346 :+1:

gimite:
Thanks for the interest, but @githaruo may be looking into this. @githaruo Do you want to take this?

githaruo:
@gimite I'm okay to pass #346 to @waseem18.

githaruo:
Sorry, I'm saying #349 (this thread).

gimite:
OK, thanks. Then @waseem18 feel free to take this issue too!

waseem18:
@gimite Happy to look into that! :)

PR #590, opened 2019-03-18
Reduce datastore queries for config settings

nworden:
The main change is to have Configuration.\_\_init\_\_ get all config settings.
Currently, a separate query is made each time a config entry needs to be
accessed. With this change, all the config entries are pulled at once in
just two queries (one for the repo config, one for the global config)
when the Configuration object is initialized. Individual entries can
then be accessed without additional datastore queries. It's pretty
similar to what gimite suggested on issue #349, except I didn't change
the datastore model (there's still a separate ConfigEntry object for
each setting, just now we get them all at once).

According to Stackdriver this reduces /datastore_v3.Get RPCs for the
repo homepages by about half (from 146 to 62). Datastore queries seem to
be the biggest contributor to latency, so this speeds things up a
noticeable amount (from ~1500-1700ms to ~700-800ms).

This means skipping the cache codepath, but the config cache seems to
be disabled in prod anyway.

There are some places where a Configuration object is unavailable (e.g.,
when HANDLER_CLASSES is built), and for them I've just left things
as they are for now.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -125,6 +125,10 @@ class ConfigEntry(db.Model):
     value = db.TextProperty(default='')
 
 
+# If calling from code where a Configuration object is available (e.g., from
+# within a handler), prefer Configuration.get. Configuration objects get all
+# config entries when they're initialized, so they don't need to make an
+# additional Datastore query.
 def get(name, default=None, repo='*'):
     """Gets a configuration setting from cache if it is enabled,
        otherwise from the database."""
@@ -145,6 +149,10 @@ def set(repo='*', **kwargs):
            value=simplejson.dumps(value)) for name, value in kwargs.items())
     cache.delete(repo)
 
+# If calling from code where a Configuration object is available (e.g., from
+# within a handler), prefer Configuration.get. Configuration objects get all
+# config entries when they're initialized, so they don't need to make an
+# additional Datastore query.
 def get_for_repo(repo, name, default=None):
     """Gets a configuration setting for a particular repository.  Looks for a
     setting specific to the repository, then falls back to a global setting."""
--- a/app/config.py
+++ b/app/config.py
@@ -195,5 +195,8 @@ class Configuration(UserDict.DictMixin):
             return self.global_config[name]
         return None
 
+    def get(self, name, default=None):
+        return self[name] or default
+
     def keys(self):
         return self.entries.keys()
--- a/app/config.py
+++ b/app/config.py
@@ -204,7 +204,13 @@ class Configuration(UserDict.DictMixin):
         return None
 
     def get(self, name, default=None):
-        return self[name] or default
+        # UserDict.DictMixin.get isn't going to do what we want, because it only
+        # returns the default if __getitem__ raises an exception (which ours
+        # never does).
+        res = self[name]
+        if res is None:
+            return default
+        return res
 
     def keys(self):
         return self.entries.keys()
--- a/app/config.py
+++ b/app/config.py
@@ -163,6 +163,13 @@ def set_for_repo(repo, **kwargs):
 class Configuration(UserDict.DictMixin):
     def __init__(self, repo):
         self.repo = repo
+        db_entries = model.filter_by_prefix(
+            ConfigEntry.all(), self.repo + ':')
+        self.entries = {
+            entry.key().name().split(':', 1)[1]: simplejson.loads(entry.value)
+            for entry in db_entries
+        }
+        self.global_config = None if repo == '*' else Configuration('*')
 
     def __nonzero__(self):
         return True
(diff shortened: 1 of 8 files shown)


app/create.py
```diff
@@ -41,8 +42,11 @@ def days_to_date(days):
 class Handler(BaseHandler):
     def get(self):
         self.params.create_mode = True
-        profile_websites = [add_profile_icon_url(website, self)
-                for website in self.config.profile_websites or []]
+        profile_websites = [
+            # add_profile_icon_url is going to modify website, so we make a copy
+            # of the config setting to avoid modifying the original.
+            add_profile_icon_url(website, self)
```
gimite: Can be a TODO, but maybe add_profile_icon_url() should rather perform deep copy internally so that it doesn't cause modification to its parameter?

app/main.py
```diff
@@ -298,17 +298,18 @@ def setup_env(request):
...
 
     # Internationalization-related stuff.
     env.charset = select_charset(request)
     env.lang = select_lang(request, env.config)
     env.rtl = env.lang in const.LANGUAGES_BIDI
 
     # Determine the resource bundle to use.
-    env.default_resource_bundle = config.get('default_resource_bundle', '1')
+    env.default_resource_bundle = (
+        env.config.get('default_resource_bundle') or '1')
```
gimite: Can't you write env.config.get('default_resource_bundle', '1') here?

gimite:
Thanks for doing this, and glad to see significant improvement of performance!

app/config.py
```diff
@@ -180,8 +187,11 @@ def __getattr__(self, name):
     def __getitem__(self, name):
         """Gets a configuration setting for this repository.  Looks for a
         repository-specific setting, then falls back to a global setting."""
-        return get_for_repo(self.repo, name)
+        if name in self.entries:
+            return self.entries[name]
+        elif self.global_config:
+            return self.global_config[name]
```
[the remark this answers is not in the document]
nworden: It won't raise an exception. Since this function overrides accesses like `config[name]`, line 193 is basically a recursive call into this same function for the global config object. The global config won't have a pointer to any other config (i.e., `self.global_config` will evaluate as false) and it'll jump down to 'return None`.
I added a test for this to be safe (`test_no_exception_for_unset_key` in `server_test_cases/config_tests.py`).

app/config.py
```diff
@@ -163,6 +163,13 @@ def set_for_repo(repo, **kwargs):
 class Configuration(UserDict.DictMixin):
     def __init__(self, repo):
         self.repo = repo
+        db_entries = model.filter_by_prefix(
```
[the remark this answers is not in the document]
nworden: Done.

app/config.py
```diff
@@ -180,8 +187,11 @@ def __getattr__(self, name):
     def __getitem__(self, name):
         """Gets a configuration setting for this repository.  Looks for a
         repository-specific setting, then falls back to a global setting."""
-        return get_for_repo(self.repo, name)
```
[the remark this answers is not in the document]
nworden: I couldn't find any get_for_repo() calls from within a handler (i.e., where a Configuration object is available).

app/create.py
nworden: Done.

app/main.py
nworden: Done.

app/config.py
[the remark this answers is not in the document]
gimite: Oh I see, it's a Config instance not a dict.

app/utils.py
```diff
@@ -614,6 +615,7 @@ def get_url(request, repo, action, charset='utf-8', scheme=None, **params):
 
 
 def add_profile_icon_url(website, handler):
+    website = copy.deepcopy(website) # avoid modifying the original
```
gimite: Nit: One more whitespace before "#".

tests/server_test_cases/config_tests.py
```diff
@@ -104,6 +104,21 @@ def test_config_namespaces(self):
...
 
+    def test_no_exception_for_unset_key(self):
+        # Tests that a config will return None, and not throw an exception, when
+        # asked for the value of an unknown key.
+        config.set_for_repo('*', good_key_1='abc', good_key_2='def')
+        config.set_for_repo('foo', good_key_1='ghi')
+        cfg = config.Configuration('foo')
+        assert cfg['good_key_1'] == 'ghi'
+        assert cfg['good_key_2'] == 'def'
+        assert cfg['unknown_key'] == None
```
gimite: Should be `is None`.

app/config.py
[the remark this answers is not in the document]
gimite: Thanks! Maybe can you add a comment in config.get_for_repo() and config.get() to warn that they are not as efficient as instance methods of Config class?

It may also be better to update https://github.com/google/personfinder/wiki/Config to suggest using Config object in the handlers. The doc currently only mentions config.get() and config.get_for_repo().

tests/server_test_cases/config_tests.py
nworden: Done.

app/utils.py
nworden: Done.

nworden:
It's failed twice on the Python 2.7.15 tests, but the failures were on different assertions, so this must just be noise.

app/config.py
```diff
@@ -193,5 +203,8 @@ def __getitem__(self, name):
             return self.global_config[name]
         return None
 
+    def get(self, name, default=None):
+        return self[name] or default
```
gimite: I believe this returns `default` when the value is 0 or ''. Can you return `default` only when the value is None?

It may also be helpful to document why you need to explicitly define this method because it is also provided by UserDict.DictMixin (I guess because `__getitem__` doesn't throw an exception for non-existing keys?).

nworden: Done. And yeah, `DictMixin.get` returns the default only if an exception is raised (it was kinda hard to find documentation of its behavior, but I tested it out).

PR #593, opened 2019-03-21
Fuzzify ages

nworden:
For data we collect from users, fuzzify ages immediately; we don't
ever want to display or share the exact age. For data exchanged
between partners via our API, we want to be an honest broker so we
don't modify their data, but we do fuzzify it when it's displayed on
our site.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -237,7 +237,7 @@ def validate_approximate_date(string):
     return ''
 
 
-AGE_RE = re.compile(r'^\d+(-\d+)?$')
+AGE_RE = re.compile(r'^(\d+)(-(\d+))?$')
 # Hyphen with possibly surrounding whitespaces.
 HYPHEN_RE = re.compile(
     ur'\s*[-\u2010-\u2015\u2212\u301c\u30fc\ufe58\ufe63\uff0d]\s*',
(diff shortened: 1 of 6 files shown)


nworden:
Sent to Avi.

app/utils.py
```diff
@@ -367,6 +367,33 @@ def validate_cache_seconds(string):
...
+
+# The range should be no more specific than a five year period.
+MIN_AGE_RANGE = 5
+
+def fuzzify_age(value):
+    if not value:
+        return None
+    parse = AGE_RE.match(value)
+    if not parse:
+        raise ValueError('Invalid age value.')
```
gimite: Drop-by comment: Just curious, do we have validation somewhere to verify that the age is in a valid format, for both records from the web UI and API? I'm wondering if it is safe to just thrown an exception here.

nworden: Oh, good point. We do have validation for input through the web UI, but not the API afaict. Changed it to return None for stuff it can't parse.

app/utils.py
```diff
@@ -367,6 +367,33 @@ def validate_cache_seconds(string):
     return 1.0
 
 
+# ==== Fuzzification functions =================================================
+
+# The range should be no more specific than a five year period.
+MIN_AGE_RANGE = 5
+
+def fuzzify_age(value):
```
gimite: Add doc comment including behavior when the input is in bad format.

app/utils.py
gimite: Just curious, do you know how None is rendered in the template? I just want to make sure that we don't render "Age: None" literally.

nworden: I changed the "if person.age" check to check fuzzified_age instead, so it won't do anything if it's None.

PR #598, opened 2019-03-26
moved authorization param list to constants in const.py

yashLadha:


yashLadha:
@nworden @gimite

nworden:
Hi Yash,

I'm not sure I understand the motivation for moving this list to `const.py`. `const.py` is meant for values that are shared between multiple modules or handlers (for example, `ROOT_URL` is needed in multiple places across the codebase, so there's no other single file that would be a particularly good place for it). However, the authorization param list is only used in `admin_api_keys.py`, so I don't think we need to move it to `const.py`.

All that said, I definitely agree that it would be better for it to be a constant instead of being embedded in the function. How about just moving it back from `const.py` (and putting it alongside `API_KEY_LENGTH` and `KEYS_PER_PAGE` in `admin_api_keys.py`)?

yashLadha:
Yaa sounds good @nworden  , updating it

app/admin_api_keys.py
```diff
@@ -28,23 +28,25 @@
 API_KEY_LENGTH = 16
 KEYS_PER_PAGE = 50
 
+# Authorization params list
```
nworden: Nit: I think you can delete this comment. It's not needed since it's redundant with the name of the constant itself.

yashLadha:
Done @nworden

app/admin_api_keys.py
yashLadha: :+1:

yashLadha:
Done @nworden

PR #600, opened 2019-03-26
 Never show external images inline.

nworden:


nworden:
The Travis run [finished (and succeeded)](https://travis-ci.org/google/personfinder/builds/511601604) at least 10 minutes ago, not sure why GitHub still has this yellow but it's good to go.

tests/test_utils.py
```diff
@@ -275,6 +275,14 @@ def test_error_message(self):
         assert 'Invalid language tag' in response.body
         assert '<script' not in response.body
 
+    def test_should_show_inline_photo(self):
+        _, _, handler = self.handler_for_url('/haiti/create')
+        # localhost is the base URL for handlers in the test environment
+        assert handler.should_show_inline_photo(
+            'http://localhost/photo.jpg') == True
```
gimite: Omit `== True` which is redundant.

tests/test_utils.py
```diff
@@ -275,6 +275,14 @@ def test_error_message(self):
...
+        assert handler.should_show_inline_photo(
```
gimite: `assert not handler.should_show_inline_photo(...)`

PR #602, opened 2019-03-26
Get pfif-tools tests passing.

nworden:
For context: this is a standalone app, separate from Person Finder itself, that's used for validating data in PFIF format, which runs at [pfif-tools.appspot.com](http://pfif-tools.appspot.com).

The tests currently depend on older (no-longer-default) versions of WebOB, so we need to specifically add the older version to the Python PATH. Also, the server tests don't seem to exist so just drop that.

The version of WebOB (v0.9) it wants is no longer supported on App Engine (according to [this doc](https://cloud.google.com/appengine/docs/standard/python/tools/built-in-libraries-27)). The site's still running, and I'm not sure if that's because it's somehow grandfathered into running v0.9 or because the problems with later WebOB versions only apply to the tests. Either way, I don't really care, because we're not going to push it again until it's migrated to Django (and possibly not until it's migrated to Python 3).


nworden:
Sent to Eliah.

nworden:
Decided to skip this step and migrate it to Django now, which was a lot easier than I expected.

PR #603, opened 2019-03-27
Run pfif-tools on Django.

nworden:
Also drops the server_tests file because it seems that no server tests actually exist.

Most of the work was updating the tests, which I guess is a good sign. Next step will be changing it to use Django templates for the dynamic pages, and probably serving the static files off Django too just to be consistent.


tools/pfif-tools/app/settings.py
```diff
@@ -0,0 +1,67 @@
+import os
+
+
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+# If we actually did anything that used the secret key we'd need to set it to
+# some constant value and find a way to secretly store it. However, pfif-tools
+# doesn't use it for anything, we just need to set it to something to make
+# Django happy.
+SECRET_KEY = os.urandom(30)
```
gimite: Optional: If it isn't used for anything, maybe it can even be an empty string or a constant string like 'dummy'?

nworden: Just to be safe in case we end up accidentally using it in the future. Added to the comment in the file.

PR #605, opened 2019-03-27
Use a Django template for pfif-tools validation results

nworden:
For context: pfif-tools is a small standalone app, separate from Person Finder itself, running at pfif-tools.appspot.com. It's a tool to help partners validate their data is in proper PFIF format.

It currently has a bunch of code to produce HTML piece-by-piece. This moves one of the dynamic pages into a template instead (the code it used to use can't be deleted yet though, because it's shared with another page I'm doing later).

### DIFF
--- a/tools/pfif-tools/app/pfif_validator.py
+++ b/tools/pfif-tools/app/pfif_validator.py
@@ -762,7 +762,7 @@ class PfifValidator:
           link_field = linking_note.find(
               self.tree.add_namespace_to_tag('linked_person_record_id'))
           messages.append(self.make_message(
-              'There is an asymmetric linked record.  That is, a note has a'
+              'There is an asymmetric linked record.  That is, a note has a '
               'linked_person_record_id to another person, but that person '
               'does not link back.',
               record=linking_note, element=link_field, is_error=False))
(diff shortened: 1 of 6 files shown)


tools/pfif-tools/app/pfif_validator.py
```diff
@@ -762,7 +762,7 @@ def validate_linked_records_matched(self):
           link_field = linking_note.find(
               self.tree.add_namespace_to_tag('linked_person_record_id'))
           messages.append(self.make_message(
-              'There is an asymmetric linked record.  That is, a note has a'
+              'There is an asymmetric linked record.  That is, a note has a '
```
ajhoffma: Only 1 space after a period.
```suggestion
              'There is an asymmetric linked record. That is, a note has a '
```

tools/pfif-tools/app/resources/validate_results.html
```diff
@@ -0,0 +1,89 @@
...
+
+<head>
+  <meta charset="utf-8">
+  <title>{{ page_title }}</title>
+  <link rel="stylesheet" type="text/css" href="/static/style.css">
+</head>
+
+<body>
+
+{% if error %}
```
ajhoffma: Maybe just render a different file on error, so you can avoid a file-spanning if/else block?

tools/pfif-tools/app/pfif_validator.py
nworden: Done.

tools/pfif-tools/app/resources/validate_results.html
nworden: Done.

PR #606, opened 2019-03-27
Add pfif-tools tests to Travis script.

nworden:


nworden:
Never mind, there's stuff to sort out here that's going to change with Python 3 anyway. Since the only development we're doing on this anytime soon is to move to Python 3, it doesn't make sense to put much effort into getting the tests running on Python 2.

Issue #226, opened 2015-12-02
Flaky server test: PersonNoteTests.test_seeking_someone_regular


gimite:
_________________ PersonNoteTests.test_seeking_someone_regular _________________
tests/server_test_cases/person_note_tests.py:597: in test_seeking_someone_regular
    self.run_test_seeking_someone_regular()
tests/server_test_cases/person_note_tests.py:582: in run_test_seeking_someone_regular
    'Expiry date of this record:': 'Jan 11, 2001, 12:00:05 AM UTC'})
tests/server_test_cases/person_note_tests.py:170: in verify_details_page
    assert fields[label].text.strip() == value
E   AssertionError: assert 'Jan 11, 2010...:00:00 AM UTC' == 'Jan 11, 2001, 12:00:05 AM UTC'
E     - Jan 11, 2010, 12:00:00 AM UTC
E     ?            -         ^
E     + Jan 11, 2001, 12:00:05 AM UTC
E     ?           +          ^

nworden:
I'm going to try splitting up this test to start with (it's pretty big). I don't think it'll solve anything, but it could simplify debugging, and if nothing else it'll ensure we test the rest of the stuff in this test even though the flakes will probably continue.

nworden:
Something I'd overlooked earlier is that it's not just the times that don't match, it's also the year. I think that rules out the possibility that it's something going on within that particular test; it has to be another test (one that uses dates in 2010) interfering with it somehow.

Since the server tests set the time by passing a CGI parameter, if we have multiple tests running in separate threads against the same server could have this problem; I'm not sure exactly how Travis runs our tests, but I've never seen this test flake locally. That would also explain why the frequency increased when we started running tests with two versions of Python 2.

Long-term, when we've migrated off webapp2 to Django, we can move most of our tests over to the Django testing framework instead of using dev_appserver for everything; then I think we can mock out the calls to get time instead of setting it via a CGI parameter. In the meantime, maybe splitting it up will reduce the flake frequency, since right now this test does a lot of stuff in between setting the time on the server and checking the expiry date (so there could be a lot of time in between for another test to come along and set its own time).

PR #607, opened 2019-03-28
Eliminate extra dependencies on django_setup.py

nworden:
It needs to be imported once, but not more than once. Since main.py runs before everything else, the import there is all we really need. Getting rid of the extra imports will make it simpler to change stuff to get it running on Django in place of webapp2.

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -18,8 +18,6 @@
 
 __author__ = 'kpy@google.com (Ka-Ping Yee)'
 
-import django_setup  # always keep this first
-
 import calendar
 import csv
 import datetime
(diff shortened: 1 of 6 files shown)


nworden:

app/api.py
```diff
@@ -34,6 +32,7 @@
 import uuid
 
 import django.utils.html
+from django.utils.translation import activate
```
gimite: I didn't know it, but looks like it is discouraged to directly import a specific function like this:
https://github.com/google/styleguide/blob/gh-pages/pyguide.md#22-imports
You should instead do:
```python
from django.utils import translation
```

That style makes sense to me. It is hard to understand what is activated by just reading `activate(query_lang)`.

IMHO one line below here can be considered an exception because it's a common pattern.

nworden: Ok, done. There's other places where we import functions, but I wouldn't want to make it worse.

nworden:
Test failures are noise (issue #226); it fails on different lines for different runs.

PR #608, opened 2019-03-28
Show notes in the record view page

gimite:


ui/src/components/Note.js
```diff
@@ -0,0 +1,169 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, {Component} from 'react';
+import {FormattedMessage, defineMessages, injectIntl} from 'react-intl';
+
+const MESSAGES = defineMessages({
+  unspecified: {
```
nworden: Could you put the messages in alphabetical order?

ui/src/components/Note.js
```diff
@@ -0,0 +1,169 @@
...
+ * Usage: <Note note={note} />
+ */
+class Note extends Component {
+  renderStatus() {
+    return <FormattedMessage {...STATUS_MESSAGES[this.props.note.status]} />;
+  }
+
+  render() {
+    return (
+      <div className='note-card'>
```
nworden: Is there a reason not to use the MDC Card component here? Since each note has its own card it seems like it'd be a good candidate for the Card component. IIRC, you didn't use MDC Cards for the person record itself to share code with the results list, but I can't think of anything this code would be shared with.

gimite:
Sorry to take this long to get back on this, but I finally resolved the comments. PTAL.

ui/src/components/Note.js
```diff
@@ -0,0 +1,170 @@
...
+  'believed_dead': MESSAGES.believedDead,
+};
+
+/**
+ * A component which shows a single note record.
+ *
+ * Usage: <Note note={note} />
+ */
+class Note extends Component {
+  renderStatus() {
```
nworden: Maybe consider inlining this? It's only called once.

PR #609, opened 2019-03-28
Use Django template for pfif-tools diff results page

nworden:
It turns out the pfif-tools code also supports a command-line interface, so I couldn't delete as much code as I'd hoped, but at least all the HTML-outputting code is gone.


nworden:
Sent to Avi.
Test failures look like they're just noise (issue #226).

ajhoffma:
One test is failing.

ajhoffma:
I agree that probably is noise.

PR #612, opened 2019-03-28
Only get auth data when needed (for API and feeds).

nworden:
Currently the global base handler is setting the auth field for every
request, but most of them don't need it. This creates parent classes for
API and feed stuff and sets the auth value there.


nworden:
Motivation for this is that I want to break up [the `auto_params` monolith](https://github.com/google/personfinder/blob/master/app/utils.py#L735). Having `BaseHandler` access parameters that only some handlers need is kinda an issue, and I want to push some responsibility down to the handlers. Later on I'll have the handlers that call `set_auth` pass in the key.

PR #614, opened 2019-03-31
Set up Django app to run alongside the webapp2 app.

nworden:
This is not meant for submission as-is. Our prod instance serves from a
subdirectory (google.org/personfinder), which makes anything related to
URLs challenging to test locally. If "personfinder" is considered the
script prefix, I think we should be able to handle URLs the same locally
and in prod, as long as we stick to using request.path_info (instead of
request.path) when we need to access URLs directly, as described here:
https://docs.djangoproject.com/en/1.11/ref/request-response/#django.http.HttpRequest.path_info

However, I'm not 100% sure about that, and I don't see a way to verify
it without trying it out in prod. This change contains configuration for
a Django app that serves a couple test pages that display information
about various variables and functions from the Django HttpRequest
object, so we can see for sure how it works. My plan is to try this in
prod, but not actually merge this PR into the master branch. Later I'll
get a real page running with Django and send that out as a separate PR
(and then I'll start gradually moving more pages to Django until we're
done with webapp2 altogether).


nworden:

nworden:
The test failure is a flake (issue #226). The Python 2.7.12 tests all passed.

app/django_setup.py
```diff
@@ -13,87 +13,37 @@
...
+The move is complicated by the fact that the original implementation uses Django
+for templating and therefore requires some Django configuration. However, Django
+can't be initialized more than once. If we try to treat the webapp2 app (with
+its Django templating setup) and the new Django app totally independently, we'll
+have a problem: whichever one happens to serve first from an instance will work
+fine, and then the next one will crash when it tries to intialize Django a
+second time. So, we want to have a single set of Django settings, intialized by
+the Django app regardless of which app serves first from an instance. The
+webapp2 app imports this, this imports the Django app's entrypoint (wsgi), and
+it initializes Django.
```
eliahhecht: unclear what "it" refers to in this line

nworden: Clarified.

nworden:
Update: this did not behave in the way I'd hoped: `path` and `path_info` had the same value (prefixed with `/personfinder`), that prefix is included in what the URL router uses, and you can't use `build_absolute_uri` with an absolute path. Setting [`FORCE_SCRIPT_NAME`](https://docs.djangoproject.com/en/2.1/ref/settings/#force-script-name) didn't help (it seems to be meant for people with sorta the opposite problem). So we'll have to do continue doing some manual URL manipulation.

PR #616, opened 2019-04-01
Use defusedxml in pfif.py

nworden:
Uses the [defusedxml](https://pypi.org/project/defusedxml/) package to prevent [billion laughs attacks](https://en.wikipedia.org/wiki/Billion_laughs_attack).


nworden:
Eliah asked offline why this seems to add the certifi library (this PR adds a bunch of files under `app/vendors/certifi-2019.3.9.dist-info`), which initially makes it seem like the defusedxml package depends on certifi for no apparent reason. I checked it out and it turns out it has nothing to do with defusedxml. We [already depend on certifi](https://github.com/google/personfinder/tree/master/app/vendors/certifi) because [the google-auth library](https://github.com/googleapis/google-auth-library-python/blob/a10b15e34447a35cedd0f73654b13404999141b5/tox.ini#L6) depends on it (we depend on google-auth by way of [google-api-python-client](https://github.com/googleapis/google-api-python-client/blob/master/tox.ini#L10)). certifi happened to get a new version since the last time anyone ran an app/vendors update, so files from that update got added. It looks like this happens a lot: [we have a bunch of .dist-info directories from old versions](https://github.com/google/personfinder/tree/master/app/vendors), and I don't exactly know what we need them for (in theory, some of our dependencies might depend on different specific versions of a library, but I don't think that could explain all of it).

IMO we should consider adding the contents of `app/vendors` to `.gitignore` rather than keeping them in the repo, and otherwise we might want to do a cleanup of the directory (maybe as simple as deleting everything and running a fresh pip install).

PR #618, opened 2019-04-02
Drop the AdminEnv class.

nworden:
This is only used in one place, and each field is only referenced once,
so I don't think we need it anymore. It's kinda a problem because it
causes a shared template to depend directly on the request object. This
is particularly an issue now because it's an obstacle to the incremental
migration from webapp2 to Django (which has different request objects).

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -1157,8 +1157,8 @@ class BaseHandler(webapp.RequestHandler):
         # If we use it, user can't sign out
         # because the error page of "login: admin" doesn't have sign-out link.
         if self.admin_required:
-            user = users.get_current_user()
-            if not user:
+            self.env.user = users.get_current_user()
+            if not self.env.user:
                 login_url = users.create_login_url(self.request.url)
                 webapp.RequestHandler.redirect(self, login_url)
                 self.terminate_response()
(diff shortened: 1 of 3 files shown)


app/utils.py
```diff
@@ -1168,6 +1168,21 @@ def __init__(self, request, response, env):
                 self.render('not_admin_error.html', logout_url=logout_url, user=user)
                 self.terminate_response()
                 return
+            self.env.logout_url = users.create_logout_url(self.request.url)
+            try:
+                # This is different from env.repo_options because this contains
+                # all repositories including deactivated ones.
+                self.env.all_repo_options = [
```
ajhoffma: Why would this throw an exception? It's not clear what's dangerous here.

nworden: Hm, yeah I don't know; the original code did this but it's not clear why. Also, now that I think about it, the reason to log the exception like this doesn't apply anymore because this is now evaluated at the same time as the rest of the code here runs (rather than from within the template). I've removed the try-except.

nworden:
The test failure is a flake (issue #226 again). Only happened on one of the 2.7 test runs.

gimite:
LGTM. Just as a context, I believe I originally introduced AdminEnv class just because I wanted to lazy-load repo_options (but I couldn't just make env.repo_options lazy-loaded because env is a struct). So your alternative approach to load it only in admin pages sounds good.

PR #623, opened 2019-04-04
Tests for access restriction of (Django-served) admin pages.

nworden:
This sets up tests to check that admin pages in the Django server are access-controlled. It doesn't test admin pages for the non-Django server (which is currently all but one of them), but we're going to move everything to Django over time, so it doesn't seem worth it to set up a way to test stuff on the old web framework.

My goals were to:
1) Ensure that nobody can forget to add tests to check that an admin page is access-restricted.
2) Make it readable.
3) Make it easy to add those tests.

I think I got 1 and 3; I'm not sure this is going to continue to be readable once I get into similar tests for XSRF checks, but hopefully it'll be ok.


nworden:
The large number of commits is because I forked this off my branch for creating a Django app in the first place (PR #620) and started working on this before the Django app was actually in. Now GitHub thinks all those commits are part of this PR (probably because we squash things before we merge them into master, so the commit that went in is its own thing).

tests/views/test_access_restriction.py
```diff
@@ -0,0 +1,73 @@
...
+
+import django.urls
+
+import urls
+import views
+
+import django_tests_base
+
+
+class AccessRestrictionTests(django_tests_base.DjangoTestsBase):
```
ajhoffma: Also test that the ones mapped to False are available to non-admins by checking that the status_code is not 403? (could be 400 because you won't set up parameters correctly)

nworden: Done.

PR #625, opened 2019-04-05
Move admin create-repo page to Django server.

nworden:
We set up a Django app (with the admin stats page) in PR #620. Now I just need to migrate all the other handlers. The tests don't verify XSRF enforcement because I plan to add XSRF stuff to the general automatic security tests (in `views/test_access_restriction.py`) separately.
I've skipped adding the view file to the new linter tool because pylint has some complaints related to the use of the env field. For unrelated reasons, Hiroshi suggested making env a class, which will solve the pylint issues; I'll do that in a separate PR and add the view to the linter then.


nworden:
Sent to Avi.

PR #627, opened 2019-04-06
Store activation status on repo objects, step 1

nworden:
TL;DR: this should let us avoid a bunch of datastore queries (helps with issue #248)

Currently, the status of a repo (whether it's launched and/or deactivated) is stored in a pair of ConfigEntries: one boolean for whether it was launched, and one for whether it was deactivated. There's a third ConfigEntry for whether the repo's in test mode.

This means that if you want to get a list of active repositories, you're likely to make a lot of datastore queries. The [`Repo.list_launched` function](https://github.com/google/personfinder/blob/047c9683b31aef990531ccd084962ddb25e88be5/app/model.py#L126) makes at least one query for each repo to see if the repo was launched; if it was launched (which most have been), it makes another query to see if it was deactivated. The `get_repo_options` function in `main.py` [calls `list_launched`](https://github.com/google/personfinder/blob/047c9683b31aef990531ccd084962ddb25e88be5/app/main.py#L201), and `get_repo_options` is called for every page [as part of `setup_env`](https://github.com/google/personfinder/blob/047c9683b31aef990531ccd084962ddb25e88be5/app/main.py#L291). Additionally, for every launched repo, `get_repo_options` will make [an additional query to check if it's in test mode](https://github.com/google/personfinder/blob/047c9683b31aef990531ccd084962ddb25e88be5/app/main.py#L206).

So, that's at least one datastore query per repo in the datastore (two queries for most of them), for serving every page. If we store this data directly on the Repo objects instead, we'll be able to do a filter in the original query. (I don't think using the Config objects would work, because those are designed to contain information about a single specific repo, whereas these queries involve every repo).

This PR is just to add the new fields to the Repo model class. My plan is to get this into prod, set the values on existing Repos in the datastore (I'll probably just do that manually), and then make the changes to actually use the new fields in a separate PR.


nworden:
Test failure is yet another instance of issue #226. One of the Python 2 test runs succeeded so I don't think it's anything to worry about.

app/model.py
```diff
@@ -104,10 +104,22 @@ class Repo(db.Model):
     existence of a repository.  Key name: unique repository name.  In the UI,
     each repository behaves like an independent instance of the application."""
 
-    # No properties for now; only the key_name is significant.  The repository
-    # title and other settings are all in ConfigEntry entities (see config.py).
-    # The per-repository 'deactivated' setting blocks UI and API access to the
-    # repository, replacing all its pages with a deactivation message.
+    class Status(object):
```
gimite: Can you document meaning of each enum value? FYI The repository admin page has detailed documentation of what each state implies.

Then you can delete the comment below:

> The per-repository 'deactivated' setting...

which is now a bit weird to be placed there.

app/model.py
```diff
@@ -104,10 +104,22 @@ class Repo(db.Model):
...
+        """An enum for the launch/activation status of the repo."""
+        UNLAUNCHED = 0
```
gimite: "Staging" is the terminology we use in the admin page now. Are you intentionally changing it? I personally prefer "staging" because "unlaunched" can mean that it was launched and then unlaunched.

app/model.py
gimite: Optional: ActivationStatus may be a better class name which is consistent with the field name and more specific.

app/model.py
```diff
@@ -104,10 +104,22 @@ class Repo(db.Model):
...
+        ACTIVE = 1
+        DEACTIVATED = 2
+
+    # Whether the repository is unlaunched, active, or deactivated.
+    activation_status = db.IntegerProperty(
+        required=False, default=Status.UNLAUNCHED)
+
+    test_mode = db.BooleanProperty(default=False)
```
gimite: Can you document what this is? Again, the admin page has a good documentation.

app/model.py
```diff
@@ -104,10 +104,22 @@ class Repo(db.Model):
...
-    # The per-repository 'deactivated' setting blocks UI and API access to the
-    # repository, replacing all its pages with a deactivation message.
+    class Status(object):
+        """An enum for the launch/activation status of the repo."""
+        UNLAUNCHED = 0
+        ACTIVE = 1
+        DEACTIVATED = 2
+
+    # Whether the repository is unlaunched, active, or deactivated.
+    activation_status = db.IntegerProperty(
```
gimite: Add TODO comment to actually use this field.

app/model.py
nworden: Done.

app/model.py
nworden: Changed to STAGING.

app/model.py
nworden: Done.

app/model.py
nworden: Done.

app/model.py
nworden: Done.

app/model.py
```diff
@@ -104,16 +104,28 @@ class Repo(db.Model):
...
     activation_status = db.IntegerProperty(
-        required=False, default=Status.UNLAUNCHED)
+        required=False, default=ActivationStatus.STAGING)
 
+    # Whether the repository is in test mode; meant for use with evergreen
+    # repositories when they're not needed. Records for repos in test mode are
+    # automatically deleted after 24 hours.
     test_mode = db.BooleanProperty(default=False)
 
     # Few properties for now; the repository title and other settings are all in
```
gimite: Delete this sentence which is now covered by the comment above:
> The per-repository 'deactivated' settings...

nworden: Done.

PR #628, opened 2019-04-06
Split up run_test_seeking_someone_regular

nworden:
It flakes frequently, and I'm hoping that splitting it up (specifically, keeping calls to set the server time close to the things that depend on them) will reduce the flakes. There's some further explanation in the comments on issue #226.

Even if it doesn't end up helping the flakes, I'd like to split it up anyway though, because it's huge and tests a really wide variety of things.

It's mostly the same tests just split up, except I've dropped a call to verify_note_form (that function's already called three other places and I can't see anything different about the call in `run_test_seeking_someone_regular`, so it just seems redundant).

Also fixes the admin_create_repo tests, which got broken in a merge.


nworden:

tests/server_test_cases/person_note_tests.py
```diff
@@ -497,34 +497,56 @@ def assert_params(url=None):
             full_name='_test_given_name _test_family_name',
             details={'Author\'s name:': '_test_author_name'})
 
-        # Now the search should yield a result.
+    @parameterized.expand([(True,), (False,)])
+    def test_search_with_result(self, use_full_text_search):
+        config.set(enable_full_text_search = use_full_text_search)
+        def assert_params(url=None):
+            self.assert_params_conform(
+                url or self.s.url, {'role': 'seek'}, {'ui': 'small'})
```
eliahhecht: maybe move this def closer to where it's being used (on line 511)?

tests/server_test_cases/person_note_tests.py
```diff
@@ -497,34 +497,56 @@ def assert_params(url=None):
...
                                  some_have=(['_test_given_name']),
                                  status=(['Unspecified']))
         self.verify_click_search_result(0, assert_params)
-        # set the person entry_date to something in order to make sure adding
-        # note doesn't update
+
+    def test_note_doesnt_update_entry_date(self):
+        self.go('/haiti/create?query=&role=seek&given_name=&family_name=')
+        self.submit_minimal_create_form(self.s.doc.cssselect_one('form'))
         person = Person.all().filter('given_name =', '_test_given_name').get()
```
eliahhecht: Assuming this is a Django model. Is there a dfiference between `Person.all().filter` and `Person.objects.filter`?

tests/server_test_cases/person_note_tests.py
```diff
@@ -497,34 +497,56 @@ def assert_params(url=None):
...
-        person.entry_date = datetime.datetime(2006, 6, 6, 6, 6, 6)
-        db.put(person)
+        entry_date = person.entry_date
         self.verify_details_page(num_notes=0)
+        self.advance_utcnow(days=5)
+        self.verify_update_notes(
+            False, '_test A note body', '_test A note author', None)
+        person = Person.all().filter('given_name =', '_test_given_name').get()
+        assert person.entry_date == entry_date
```
eliahhecht: Why `assert` rather than `self.assertEqual` and friends? My understanding is that the unit test assert methods produce better error messages.

tests/server_test_cases/person_note_tests.py
```diff
@@ -497,34 +497,56 @@ def assert_params(url=None):
...
+        self.go('/haiti/create?query=&role=seek&given_name=&family_name=')
+        self.submit_minimal_create_form(self.s.doc.cssselect_one('form'))
         person = Person.all().filter('given_name =', '_test_given_name').get()
-        person.entry_date = datetime.datetime(2006, 6, 6, 6, 6, 6)
-        db.put(person)
+        entry_date = person.entry_date
         self.verify_details_page(num_notes=0)
+        self.advance_utcnow(days=5)
+        self.verify_update_notes(
+            False, '_test A note body', '_test A note author', None)
```
eliahhecht: might be worth explicitly using keywords for these arguments, since it is otherwise difficult to tell what False and None are doing. (this applies a few other places in this file)

tests/server_test_cases/person_note_tests.py
nworden: We don't use the Django ORM, these are App Engine datastore calls.
There are two ways to start a query with the db library, shown here:
https://cloud.google.com/appengine/docs/standard/python/datastore/queryclass
(Person Finder uses `.all()` consistently.)

tests/server_test_cases/person_note_tests.py
nworden: Done.

tests/server_test_cases/person_note_tests.py
nworden: Done.

tests/server_test_cases/person_note_tests.py
nworden: Oh, I hadn't realized that. PF currently uses `assert` for basically all tests, but I've updated the use of `assert` that I introduced here.

PR #629, opened 2019-04-07
Make env a class and refactor params-reading

nworden:
This sets up a class (and an admin subclass) for environment. I also refactored the params-reading function; you brought up the fact that params can be different depending on when you read them, and this fixes that. I'd still prefer to keep the env-setting logic in the view/handler classes though. If we moved that into the env init functions, we'd have to give them a way to build URLs (I guess by passing them a reference to the view class) and probably other complexity later, and that seems kinda messy to me. IMO it's cleaner to keep env as simple containers.


app/views/base.py
```diff
@@ -37,6 +37,93 @@ class BaseView(django.views.View):
...
+            """Gets the action ID, an identifier for the page being served."""
+            return self._action
+
+        @action.setter
+        def action(self, value):
+            self._action = value
+
+        @property
+        def charset(self):
+            """Gets the character set being used to serve the page."""
```
gimite: Optional nit: Character **encoding** would be a more accurate name here (though it is named "charset").

app/views/base.py
```diff
@@ -37,6 +37,93 @@ class BaseView(django.views.View):
...
+            """Gets the URL for the global root."""
+            return self._global_url
+
+        @global_url.setter
+        def global_url(self, value):
+            self._global_url = value
+
+        @property
+        def lang(self):
+            """Gets the two-letter code for the language being used."""
```
gimite: Since it is not always 2-letter (e.g., "prs", "zh-TW"), may be better to just call it "language code" and refer to const.py.

app/views/base.py
```diff
@@ -37,6 +37,93 @@ class BaseView(django.views.View):
...
+            """Gets the repository ID, or None if it's a global page."""
+            return self._repo
+
+        @repo.setter
+        def repo(self, value):
+            self._repo = value
+
+        @property
+        def rtl(self):
+            """Gets whether the language a right-to-left language."""
```
gimite: Nit: the language **is** a right-to-left language

app/views/admin/base.py
```diff
@@ -42,6 +72,12 @@ def setup(self, request, *args, **kwargs):
             for repo in sorted(model.Repo.list())
         ]
 
+    def get_params(self):
```
gimite: Should each view override get_params() like this? It would be a bit tedious. How about making _GET_PARAMETERS and _POST_PARAMETERS methods, and let each view class only override them?:

```python
class MyView(...):
    def get_get_param_names(self):
        super(MyView, self).get_get_param_names() + ['param1', 'param2']
    def get(self, ...):
        params = self.get_params()  # Returns param1, param2 and inherited params
        ...
```

app/views/admin/base.py
```diff
@@ -29,9 +28,40 @@ class AdminBaseView(views.base.BaseView):
...
     }
 
+    class Env(views.base.BaseView.Env):
+        """Class to store environment information used by views and templates.
+        """
+        # pylint: disable=attribute-defined-outside-init
+
+        @property
+        def all_repo_options(self):
+            """Gets a list of Structs with repo IDs and URLs."""
```
gimite: Can you also document field names in the struct?

app/views/base.py
```diff
@@ -37,6 +37,93 @@ class BaseView(django.views.View):
         'lang': utils.strip,
     }
 
+    class Env(object):
+        """Class to store environment information used by views and templates.
```
gimite: Can you also document that a subclass of BaseView can define a nested `Env` class to be used as a replacement of this class? Since it is a bit unusual API.

app/views/base.py
```diff
@@ -37,6 +37,96 @@ class BaseView(django.views.View):
...
+
+        Subclasses of BaseView may define their own Env class (which must be
```
gimite: Can you also mention that, when a subclass defines its own Env class, it will automatically used to construct self.env in the subclass? That's the part I feel a bit tricky.

app/views/admin/base.py
gimite: That's a fair point. Though I personally feel more tedious with the current implementation where you always need to:

* Define MyView._GET_PARAMS
* Override get_params() which is mostly boilerplate

So you do need multiple definitions. But it is more like a personal preference, so I'm OK either ways.

But can you document in BaseView.get_params() how subclasses are supposed to override get_params(), maybe with sample code? It doesn't look obvious from the current document.

And optionally, maybe it's more concise by inlining the constants?:

```python
    def get_params(self):
        return views.base.read_params(
            super(AdminBaseView, self).get_params(),
            self.request,
            post_params={'xsrf_token': utils.strip})
```

gimite:
I haven't merged it because Travis is failing. Can you merge it yourself if you confirm that the failure is noise?

nworden:
It's pylint, it thinks config is a standard import rather than our own module for some reason, and complains that the import order is wrong. I don't want to block stuff on pylint so I just took the file involved (`app/views/base.py`) back out of `tools/lint` for now.

gimite:
> It's pylint, it thinks config is a standard import rather than our own module for some reason, and complains that the import order is wrong. I don't want to block stuff on pylint so I just took the file involved (`app/views/base.py`) back out of `tools/lint` for now.

Can't you just add a magic comment to disable the specific warning as you have done for other lint errors?

nworden:
Yeah I could have, but I've seen this before (specifically with `config`), and I want to figure out the reason if I can. Otherwise we'll have to have keep putting that skip comment in a ton of files.

gimite:
Sounds good to investigate first, thanks!

Issue #619, opened 2019-04-02
Automated style checking


nworden:
It'd be nice to have some kind of style check run automatically. [yapf](https://github.com/google/yapf/) seems like a good place to start. The docs say if you run it with `-d` you get a non-zero return code if changes are required, so you can integrate it into an automated script. It's not a complete solution (e.g., it doesn't seem to detect imports of individual functions, issue #615, even if you run it with `--style google`), but it's better than nothing.

We should also consider using pylint, which won't fix things automatically but is more thorough.

I don't think it would be practical to try to do this all at once, but we could probably do it the same way I'm handling Python 3 testing: have a list of specific files that should be checked, and add files to it as we update them to be in line with the style guide. I'll start by running pylint and yapf over the new Django view classes as I write them.

gimite:
Oh so yapf performs auto formatting like clang-format. That sounds nice! I like auto formatting (though some people hates it).

> I don't think it would be practical to try to do this all at once

Just curious, why is that? Is it possible that yapf breaks the code, not just changing formatting? Can't we apply it to all files at once if it safely reformats files without changing semantics?

Also just curious, do you know of anything similar for JS(X)?

nworden:
yapf isn't expected to change semantics, but it looks like there are going to be thousands of lines that need to change just under /app (not including app/vendors), so if we tried to do it all at once we'd have a high risk of merge conflicts and a huge PR to review. Maybe we could have a rule that if you change a file that's not already checked by yapf you should run yapf on it and add it to the list? Since every file is probably going to be affected by the Python 3 work, every file would get done this year (hopefully).

I'm not sure what to do about pylint, which will be a lot more work; I wouldn't want to block PRs on fixing pylint warnings.

It looks like there's a bunch of options for formatting JS, but none of them really dominate the market (for lack of a better way to put it). The Google JS style guide recommends clang-format (the Closure team has instructions [here](https://github.com/google/closure-library/wiki/Formatting-.js-with-clang-format#formatting-js-with-clang-format)), but I'd rather not make people go out of their way to install too much stuff when we can help it (pylint and yapf don't seem like a problem since we can just add them to the pip install command). Maybe we could use js-beautifier or prettier (both available as NPM packages).

gimite:
[...]

PR #630, opened 2019-04-07
First batch of yapf auto-formatting.

nworden:
This covers all our files under app/ which don't start with a, f, m, or v.

Starts addressing issue #619.


nworden:
I'll do tests separately. I skipped some letters because I'm doing work on admin handlers, you have a pending change for frontend_api.py, I have a pending change for model.py, and vintol mentioned he's working on views.py.

nworden:
Hm, I think I figured out why Jakub's PR had a lint failure: yapf had a release today, and I think it generated different output than before. I'm going to close this PR for now, and look into other auto-formatter options. yapf seemed like a good choice because it's from Google and would be sure to use the style guide we use, but maybe it'd be better to choose something more mature. Enforcing use of an auto-formatter with less-than-stable output seems like a real hassle.

gimite:
If we cannot find a good alternative, does it make sense to stick to a specific version of yapf (and perform project-wide batch update when we feel the version is too old)?

nworden:
Replied on issue #619

PR #633, opened 2019-04-08
Use HTTPS everywhere, except for tasks.

nworden:
I was initially hesitant to require HTTPS, because of performance concerns, but I did some research and here's what I found:
- All else held equal, HTTPS is slower than HTTP, generally by a couple hundred milliseconds.
- However, if you use HTTPS, it's possible to use HTTP/2 instead of HTTP/1.1, and (all else held equal) HTTP/2 is faster than HTTP/1.1. (Nobody really supports HTTP/2 without HTTPS.)

The top answer here was helpful in understanding it: https://stackoverflow.com/questions/36293889/why-might-an-https-page-request-be-faster-than-http

I had a hard time finding hard numbers comparing the two, but there's a post on the Akamai blog that says HTTP/2 generally makes up for the HTTPS overhead and comes out ahead of insecure HTTP/1.1:
https://blogs.akamai.com/2015/08/is-http2-worth-the-performance-price-of-tls.html

App Engine will use HTTP/2 automatically if you're serving over HTTPS:
https://cloudplatform.googleblog.com/2015/10/Full-Speed-Ahead-with-HTTP2-on-Google-Cloud-Platform.html

So, I'm sold on HTTPS. On the Django side, it's easy (there's a boolean you set in your settings module). For the short-term, I'll make force-HTTPs the default option for new repos (being done as part of PR #625) and make sure the webapp2 handlers are have https_required set to True (I don't want to spend time going out of my way to clean that up immediately; it'll get eliminated as we migrate to Django).

We can't use HTTPS for task pages, because App Engine makes HTTP requests for stuff you put in the task queue.

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -509,7 +509,7 @@ class Write(BaseApiHandler):
 
 
 class Search(BaseApiHandler):
-    https_required = False
+    https_required = True
 
     def get(self):
         if self.config.search_auth_key_required and not (
--- a/app/feeds.py
+++ b/app/feeds.py
@@ -54,7 +54,7 @@ class Repo(BaseFeedsHandler):
     TITLE = 'Person Finder Repository Feed'
 
     repo_required = False
-    https_required = False
+    https_required = True
     # For a deactivated repo, return an empty feed instead of an error page.
     ignore_deactivation = True
 
@@ -127,7 +127,7 @@ class Person(BaseFeedsHandler):
 
 class Note(BaseFeedsHandler):
     # SSL check is done in get() if person_record_id is not specified.
-    https_required = False
+    https_required = True
 
     def get(self):
         # SSL and auth key is not required if a feed for a specific person
--- a/app/const.py
+++ b/app/const.py
@@ -20,7 +20,7 @@ from __future__ import absolute_import
 from django.utils.translation import gettext_lazy as _
 
 # The root URL of this application.
-ROOT_URL = 'http://google.org/personfinder'
+ROOT_URL = 'https://google.org/personfinder'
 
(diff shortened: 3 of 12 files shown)


nworden:
Test failures are more of issue #226. One of the Python 2 runs succeeded so I'm not worried about it.

app/api.py
```diff
@@ -509,7 +509,7 @@ def write_status(self, type, written, skipped, total, id_field):
 
 
 class Search(BaseApiHandler):
-    https_required = False
+    https_required = True
```
gimite: I wonder if this can break existing clients, but it would probably be safe because our document points to HTTPS URLs?:
https://github.com/google/personfinder/wiki/DataAPI

We can probably just launch it to see if anyone complains.

app/feeds.py
```diff
@@ -54,7 +54,7 @@ class Repo(BaseFeedsHandler):
     TITLE = 'Person Finder Repository Feed'
 
     repo_required = False
-    https_required = False
+    https_required = True
```
gimite: Same as API.

app/api.py
nworden: Yeah, and I'd be kinda surprised if someone couldn't handle making HTTPS requests these days. If someone does break though, this should be an easy PR to roll back.

gimite: I believe it is still common that it requires an additional step to support HTTPS in HTTP client libraries, like pointing to the path to SSL certificates in the system. It is also possible that the request is not configured to handle HTTP redirects. Anyway hopefully the clients have specified HTTPS URLs from the beginning.

PR #636, opened 2019-04-09
Don't require HTTPS locally.

nworden:


nworden:
More of issue #226. The other Python 2 run succeeded.

PR #637, opened 2019-04-09
Automated XSRF testing.

nworden:
This refactors the admin access tests into a more general-purpose security test file, to accommodate XSRF testing too.
Also changes the XSRF utility to handle the case where a super-invalid token (missing the expected slash) is passed in.


nworden:
Sent to Avi.

nworden:
Added a test for task access restriction (related to P$ #640, but I think the merge would have been a mess if I'd added this test in that PR since the file's changing so much here).

app/utils.py
```diff
@@ -1253,11 +1253,14 @@ def generate_token(self, user_id, action_id):
             action_time)
 
     def verify_token(self, token, user_id, action_id):
```
ajhoffma: Is there no django or python utility to do this for you?

tests/views/test_auto_security.py
```diff
@@ -0,0 +1,251 @@
...
+    'PathTestInfo',
+    [
+        # Whether the path is expected to accept GET requests.
+        'accepts_get',
+        # Whether the past is expected to accept POST requests.
+        'accepts_post',
+        # Whether the page is restricted to admins.
+        'restricted_to_admins',
+        # Whether POST requests should require an XSRF token.
+        'requires_xsrf',
```
ajhoffma: will this every be false if accepts_post is true?

app/utils.py
nworden: Sent you an email about this.

tests/views/test_auto_security.py
nworden: Yeah, because we only use XSRF tokens for admin pages. The user-facing pages don't require login, so there's not much point in using XSRF.

PR #639, opened 2019-04-14
Content security policy for Django-served pages

nworden:
This adds a strict CSP for pages served by Django, based on the example in [this doc](https://csp.withgoogle.com/docs/strict-csp.html). The plan is to get additional templates in line with the CSP as the webapp2->Django migration progresses.

This could have been implemented the same way the CSP is currently implemented for React pages ([here](https://github.com/google/personfinder/blob/5f860a20afca38f328583868580ccc0127841574/app/main.py#L544)), but I prefer having it be part of the middleware, so that it can't be forgotten. If we put this in `BaseView.render` or something, we'd risk leaving it out if we later added a variant of the `render()` function or something like that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,10 @@ install:
       git \
       time \
       unzip
+  # The default version of pip will try to install the latest version of Django
+  # (apparently as part of installing django_csp), which doesn't work because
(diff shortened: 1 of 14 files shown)


nworden:

PR #640, opened 2019-04-15
Migrate sitemap and sitemap ping task to Django

nworden:
This PR:
1) Moves the sitemap view to Django. It also drops repo-specific sitemaps, which aren't a desired feature anymore and are kinda broken already I think.
2) Sets up a base Django "view" for task handlers. I didn't want to put these under `/views` because they don't seem very view-y to me. I'd like to put them in a "tasks" module, but tasks.py is currently using that name, so I'm temporarily calling it "tasksmodule". Once everything's moved out of tasks.py I'll rename the module to just "tasks."
3) Moves the sitemap ping task to a Django handler, subclassing the base task view.

This doesn't have tests for the required-GAE-task-header part, because I'd like to do that as part of our generic security tests, and that file is changing significantly in PR #637 (under review), so adding those tests in this PR would be messy.


tests/tasks/test_sitemap_ping.py
```diff
@@ -0,0 +1,44 @@
...
+        # We set it to "really" ping index servers for the test. Since we're
+        # also mocking out requests.get though, it won't really succeed in
+        # pinging anything.
+        config.set(ping_sitemap_indexers=True)
+        with mock.patch('requests.get') as requests_mock:
+            res = self.run_task('/global/tasks/sitemap_ping',
+                                {'search_engine': 'google'})
+            assert len(requests_mock.call_args_list) == 1
+            call_args, _ = requests_mock.call_args_list[0]
+            assert call_args[0] == ('https://www.google.com/ping?sitemap='
```
eliahhecht: use self.assert* ?

nworden: Oh, right, thanks for the reminder.

PR #643, opened 2019-04-19
Migrate delete_expired tasks to Django.

nworden:


nworden:
The recaptcha changes (in utils.py) here are to deal with issue #644 and keep Travis happy, but aren't really related to this PR, and are also being done separately in PR #645 to unblock other stuff. I'll leave issue #644 open until I really understand what's going on, but I can't see any way it might have been caused by migrating these tasks to Django.

PR #646, opened 2019-04-19
Migrate API key management pages to Django.

nworden:
This also:
- Gets the page in line with our new content security policy (required because we're having the Django server set it automatically on all responses).
- Reworks the templates to avoid generating HTML in the code.
- Adds the admin's IP address and the state of the key to the management log entities (using the fields added in PR #635).


app/resources/admin_api_key_form.html.template
```diff
@@ -204,10 +207,20 @@
 
   <p>
     <input type="hidden" name="key" value="{{target_key.key}}">
-    <input type="submit" name="submit" value="{{operation_name}}" id="submit"
-           onclick="return validate_fields();">
-    <div id="mandatory_field_missing"
-         class="mandatory" style="display: none">
+    {% if operation_type == 'update' %}
+      <input
+          type="submit"
```
eliahhecht: nit: this indentation is different than the following lines?

app/views/admin/api_keys.py
```diff
@@ -0,0 +1,225 @@
...
+_API_KEY_LENGTH = 16
+
+
+class ApiKeyListView(views.admin.base.AdminBaseView):
+    """The admin API key list view."""
+
+    ACTION_ID = 'admin/api_keys/list'
+
+    def get(self, request, *args, **kwargs):
+        """Serves get requests.
```
eliahhecht: maybe explain what the method is actually doing? i.e. what is returned (and then you can probably drop the 'Returns' clause).

app/views/admin/api_keys.py
```diff
@@ -0,0 +1,225 @@
...
+                message = t.ugettext(
+                    'A new API key has been created successfully.')
+            elif management_log.action == model.ApiKeyManagementLog.UPDATE:
+                message = t.ugettext(
+                    'The API key has been updated successfully.')
+            return self._render_form(management_log.authorization, message)
+        else:
+            return self._render_form()
+
+    def post(self, request, *args, **kwargs):
```
eliahhecht: this method is pretty long. would it be possible to break it up?

app/views/admin/api_keys.py
```diff
@@ -0,0 +1,225 @@
...
+            # Just override the existing one.
+            existing_authorization = db.get(self.params.get('key'))
+            if not existing_authorization:
+                return self.error(404, t.ugettext(
+                    'No such Authorization entity.'))
+            # This shouldn't happen unless an admin does something funny with
+            # URLs, but check just to be safe.
+            if existing_authorization.repo != repo:
+                return self.error(400, t.ugettext(
+                    'Authorization already exists for another repo!'
```
eliahhecht: nit: missing a space between the two sentences?

app/views/base.py
```diff
@@ -210,6 +210,9 @@ def build_absolute_path(self, path=None):
             path (str, optional): A path beginning with a slash (may include a
                 query string), e.g., '/abc?x=y'. If the path argument is not
                 specified or is None, the current request's path will be used.
+            repo (str, optional): A repo ID. If specified, the path will be
+                considered relative to the repo's route. If this is specified,
+                path must also be specified.
```
eliahhecht: currently if the user specifies repo but not path, the first branch of the if-statement will execute, which may be unexpected. maybe add an assertion that path must be present if repo is present?

app/views/admin/api_keys.py
nworden: Fixed.

app/views/admin/api_keys.py
nworden: A little? I factored out the Authorization-creating code and added a general XSRF enforcement function to the base admin handler (it works for 403s because we don't include any message with those, but I can't do the same with cases that involve a custom error message).

app/views/admin/api_keys.py
nworden: Done.

app/resources/admin_api_key_form.html.template
nworden: Fixed (it was the other lines that were wrong).

app/views/base.py
nworden: Done.

app/views/admin/api_keys.py
eliahhecht: :+1:

PR #648, opened 2019-04-20
Store activation status on repo objects, step 2

nworden:
Sequel to PR #627 (motivation explained there). This writes to the new fields when repos are created or edited.

### DIFF
--- a/app/admin.py
+++ b/app/admin.py
@@ -184,6 +184,10 @@ class Handler(BaseHandler):
                 break
 
         config.set_for_repo(repo, **self.__view_config_to_model_config(values))
+        if self.params.operation == 'save_repo':
+            # We don't want to try doing this for global saves; that wouldn't
+            # make sense.
+            self.__update_repo_status(values)
         return True
 
     def __model_config_to_view_config(self, model_config):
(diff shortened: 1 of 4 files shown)


nworden:

app/admin.py
```diff
@@ -225,3 +229,18 @@ def __view_config_to_model_config(self, view_config):
             else:
                 model_config[name] = value
         return model_config
+
+    def __update_repo_status(self, view_config):
+        repo_entity = model.Repo.get_by_key_name(self.env.repo)
+        launch_status = view_config['launch_status']
+        if launch_status == 'staging':
```
christopherestif: just by habbit, I'll ask if it's worth putting these hard-coded strings in some constants somewhere?

tests/views/test_admin_create_repo.py
```diff
@@ -54,13 +54,16 @@ def test_create_repo(self):
...
+        self.assertEqual(
+            repo.activation_status, model.Repo.ActivationStatus.STAGING)
+        self.assertFalse(repo.test_mode)
         # Check a couple of the config fields that are set by default.
         repo_conf = config.Configuration('idaho')
-        assert repo_conf.language_menu_options == ['en', 'fr']
-        assert not repo_conf.launched
-        assert repo_conf.time_zone_abbreviation == 'UTC'
+        self.assertEqual(repo_conf.language_menu_options, ['en', 'fr'])
+        self.assertFalse(repo_conf.launched)
```
christopherestif: supernit, self.assertIs(..., False) is more specific than self.assertFalse(...):
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertFalse
I don't know which one is more relevant to your case though

app/admin.py
nworden: Done.

tests/views/test_admin_create_repo.py
nworden: Thanks, changed to assertIs.

PR #650, opened 2019-04-22
Migrate admin delete-record page to Django.

nworden:


tests/views/test_admin_delete_record.py
```diff
@@ -0,0 +1,86 @@
...
+        get_doc = self.to_doc(self.client.get(
+            '/global/admin/delete_record/', secure=True))
+        xsrf_token = get_doc.cssselect_one('input[name="xsrf_token"]').get(
+            'value')
+        post_resp = self.client.post('/haiti/admin/delete_record/', {
+            'xsrf_token': xsrf_token,
+            'id': self.person.record_id,
+        }, secure=True)
+        # Check that the user's redirected to the repo's main admin page.
+        assert isinstance(post_resp, django.http.HttpResponseRedirect)
```
eliahhecht: self.assertIsInstance?

tests/views/test_admin_delete_record.py
```diff
@@ -0,0 +1,86 @@
...
+            home_neighborhood='Good Neighborhood',
+            author_name='Alice Smith',
+            author_phone='111-111-1111',
+            author_email='alice.smith@gmail.com',
+            source_url='https://www.source.com',
+            source_date=datetime.datetime(2010, 1, 1),
+            source_name='Source Name',
+            entry_date=datetime.datetime(2010, 1, 1),
+            expiry_date=datetime.datetime(2010, 2, 1),
+            other='')"""
```
eliahhecht: this seems like a lot of details that don't matter much to the test. would there be any value in extracting a utility function like "create_person", that could also be used by other tests?

nworden: Ha, oops, I did that in PR #649, merged it into this branch, used the shared function on the line below and commented this stuff while I checked that the tests still passed, and then forgot to delete this commented-out code.

tests/views/test_admin_delete_record.py
nworden: Done.

PR #652, opened 2019-04-23
Model for admin ACLs.

nworden:
As with tasks, I don't want to put this in the already-large `model.py`, but that file is occupying the name I really want to use, so I put this under `modelmodule` for now and I'll rename it whenever I clear out `model.py`.


app/modelmodule/logging.py
```diff
@@ -0,0 +1,36 @@
+"""Entities for logging permissions changes and other such things."""
```
christopherestif: GGive has been using SimpleHistory to track model changes, it was a super light-weight integration and does just the trick (adding it to the middleware too):
from simple_history.models import HistoricalRecords
class AdminPermission(db.Model):
  ...
  history = HistoricalRecords()

app/modelmodule/admin_acls.py
```diff
@@ -0,0 +1,65 @@
+"""Model for admin/moderator access control."""
+
+
+from google.appengine.ext import db
+
+
+class AdminPermission(db.Model):
+    """Entity class for tracking admin/moderator access.
+
+    There should be only one AdminPermission entity for any given repo/email
```
christopherestif: can that constraint be enforced by a 
class AdminPermission(db.Model):
  class Meta(object):
    unique_together = (('repo', 'email_address'),)

app/modelmodule/admin_acls.py
```diff
@@ -0,0 +1,65 @@
...
+    address pair.
```
christopherestif: in general, could there be an edge case where a user is eg a moderator until Jan 1, and later on they should be an administrator - but there's already a pair with their email and the repo, so they can't be added until the old entry is deleted?
Maybe that's working as intended - just wanted to flag it.

app/modelmodule/admin_acls.py
```diff
@@ -0,0 +1,65 @@
...
+            access_level=access_level,
+            expiration_date=expiration_date)
+
+    @staticmethod
+    def get(repo, email_address):
+        return AdminPermission.get_by_key_name(
+            AdminPermission._key_name(repo, email_address))
+
+    def permission_state(self):
+        """Returns a string representation of the permission's state.
```
christopherestif: just checking, would it be worth making it the __unicode__(self) method?

app/modelmodule/logging.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class AdminPermissionManagementLog(db.Model):
+    """An entity class for logging changes to admin permissions."""
+
+    # The email address of the admin making the change.
+    managing_admin_email = db.StringProperty(required=True)
+
+    # The IP address of the user making the change.
+    ip_address = db.StringProperty(required=True)
```
christopherestif: not entirely sure if the IP address gives more than the email of the admin making the change? (since IP addresses might not always be trustworthy)

app/modelmodule/admin_acls.py
nworden: Not exactly, because `unique_together` is part of the Django ORM, which we're not using. However, the key is defined by these two fields, and the key has to be unique, so it kinda works out the same.

app/modelmodule/admin_acls.py
nworden: I can't think of a situation where we'd need to support timed upgrades/downgrades, I think it'll be fine to require that someone just edit the existing entry.

app/modelmodule/logging.py
nworden: Yeah, but it's better than nothing. My thinking here is that if someone's credentials got stolen we'd at least have a record of when changes started coming from another IP address.

app/modelmodule/admin_acls.py
nworden: Since this method has a very specific purpose (logging) I wanted to use a different method than something general like unicode.

app/modelmodule/logging.py
nworden: I don't think we can use that since we don't use Django's ORM, but it does seem worth considering other more generalized options, since we have other stuff we want to log in a similar way (API key changes are already logged much like what I have here, and eventually I want to do the same thing for repo settings and the global config). I'm going to drop this file from this PR and just merge `admin_acls.py` in for now, and do the logging in a separate PR later.

PR #653, opened 2019-04-25
Page for adding/editing/removing admins

nworden:


tests/views/test_admin_acls.py
```diff
@@ -0,0 +1,171 @@
...
+        xsrf_token = get_doc.cssselect('input[name="xsrf_token"]')[1].get(
+            'value')
+        post_resp = self.client.post('/haiti/admin/acls', {
+            'xsrf_token': xsrf_token,
+            'email_address': 'j@mib.gov',
+            'expiration_date': '2019-04-25',
+            'level': 'moderator',
+            'revoke_button': 'Revoke',
+        }, secure=True)
+        acls = admin_acls_model.AdminPermission.all()
```
christopherestif: maybe .count() instead of .all() directly?

tests/views/test_admin_acls.py
```diff
@@ -0,0 +1,171 @@
...
+        self.assertEqual(acls.count(), 0)
```
christopherestif: just checking: is it worth asserting there was 1 entry between the setup and the post?

tests/views/test_admin_acls.py
```diff
@@ -0,0 +1,171 @@
...
+        self.data_generator.admin_permission(email_address='j@mib.gov')
+        get_doc = self.to_doc(self.client.get(
+            '/haiti/admin/acls/', secure=True))
+        xsrf_token = get_doc.cssselect('input[name="xsrf_token"]')[1].get(
+            'value')
+        post_resp = self.client.post('/haiti/admin/acls', {
+            'xsrf_token': xsrf_token,
+            'email_address': 'j@mib.gov',
+            'expiration_date': '2019-04-25',
+            'level': 'moderator',
```
christopherestif: it looks like 'expiration_date' and 'level' are not checked by the server during the deletion - should they be removed from the unit test?

tests/views/test_auto_security.py
```diff
@@ -117,6 +129,16 @@ class AutoSecurityTests(view_tests_base.ViewTestsBase):
...
 
+    def init_testbed_stubs(self):
+        """Initializes the App Engine testbed stubs.
+
+        The base class initializes a Datastore stub, which seems to cause
+        problems when we set another app ID (for the task tests). We don't
+        really need Datastore for this, so just override the base class and
+        stick to the user stub.
+        """
+        self.testbed.init_user_stub()
```
christopherestif: just checking, should there be a matching self.testbed.deactivate()?

app/modelmodule/admin_acls.py
```diff
@@ -48,6 +49,10 @@ def get(repo, email_address):
         return AdminPermission.get_by_key_name(
             AdminPermission._key_name(repo, email_address))
 
+    @staticmethod
+    def get_for_repo(repo):
+        return AdminPermission.all().filter('repo =', repo)
```
christopherestif: do we need .all() before .filter()?

nworden: Weirdly yes, `all()` is one of the two ways to get a `Query` object to call `filter` on (the other way is less counter-intuitive, but Person Finder code is entirely consistent about its use of `all()` so I stick to it).

tests/views/test_admin_acls.py
nworden: Unfortunately I need `all()` to get a query object.

tests/views/test_admin_acls.py
nworden: I don't think it's needed; if the data_generator isn't doing its job for some reason plenty of other tests will fail.

tests/views/test_admin_acls.py
nworden: The server always reads all of them at the beginning of the `post` function, so I need them to be present.

tests/views/test_auto_security.py
nworden: The base class does that (the base class also sets these up, just not quite the way we need for this test).

nworden:
FTR: not sure why GitHub doesn't seem to know it, but Travis finished a while ago and the tests passed.

PR #654, opened 2019-04-26
Enforce new admin ACLs in Django pages.

nworden:
In addition to the enforcement, this makes two changes:
- It rearranges the model to support three levels. We'd originally planned on giving the higher level to access to global admins, but that introduces some implementation complexity and makes the whole thing more rigid. By splitting it up like this instead, I think it'll be more flexible and better able to handle future feature requests.
- Adds a constant to the `site_settings` module for a root admin, so that someone setting up from scratch has a way to get in.

### DIFF
--- a/app/modelmodule/admin_acls.py
+++ b/app/modelmodule/admin_acls.py
@@ -32,7 +32,7 @@ class AdminPermission(db.Model):
         # An ordered list of the admin access levels, from lowest to highest
         # (i.e., each level has the permissions of everything before it in the
         # list).
-        ORDERED = [MODERATOR, MANAGER, SUPERADMIN]
+        ORDERING = [MODERATOR, MANAGER, SUPERADMIN]
 
     access_level = db.IntegerProperty(required=True)
 
@@ -74,9 +74,9 @@ class AdminPermission(db.Model):
         """
         if self.access_level == other_level:
             return 0
-        level_index = AdminPermission.AccessLevel.ORDERED.index(
+        level_index = AdminPermission.AccessLevel.ORDERING.index(
             self.access_level)
-        other_level_index = AdminPermission.AccessLevel.ORDERED.index(
+        other_level_index = AdminPermission.AccessLevel.ORDERING.index(
             other_level)
         if level_index < other_level_index:
             return -1
--- a/app/modelmodule/admin_acls.py
+++ b/app/modelmodule/admin_acls.py
@@ -29,6 +29,11 @@ class AdminPermission(db.Model):
         # Moderators can moderate user input.
         MODERATOR = 2
 
+        # An ordered list of the admin access levels, from lowest to highest
+        # (i.e., each level has the permissions of everything before it in the
+        # list).
+        ORDERED = [MODERATOR, MANAGER, SUPERADMIN]
+
     access_level = db.IntegerProperty(required=True)
 
     # The expiration date for the permission.
@@ -55,6 +60,28 @@ class AdminPermission(db.Model):
     @staticmethod
     def get_for_repo(repo):
         return AdminPermission.all().filter('repo =', repo)
+
+    def compare_level_to(self, other_level):
+        """Compares the level of this permission to the given level.
+
+        Args:
+            other_level (int): An AccessLevel value to compare to.
+
+        Returns:
+            int: A negative value if this permission has a lower level, a
+            positive value if this permission has a higher level, or 0 if the
+            levels are the same.
+        """
+        if self.access_level == other_level:
+            return 0
+        level_index = AdminPermission.AccessLevel.ORDERED.index(
+            self.access_level)
+        other_level_index = AdminPermission.AccessLevel.ORDERED.index(
+            other_level)
+        if level_index < other_level_index:
+            return -1
+        else:
+            return 1
 
     def permission_state(self):
         """Returns a string representation of the permission's state.
--- a/app/modelmodule/admin_acls.py
+++ b/app/modelmodule/admin_acls.py
@@ -20,11 +20,14 @@ class AdminPermission(db.Model):
     class AccessLevel(object):
         """An enum for the level of access the user has."""
 
-        # Admin access: the admin can edit the repo, add new moderators, etc.
-        ADMINISTRATOR = 0
-        # Moderator access: the admin can moderate user input, but not anything
-        # else.
-        MODERATOR = 1
+        # Superadmins can do anything for a repo, and global superadmins can
+        # create new repos.
+        SUPERADMIN = 0
+        # Managers have limited administrative access: e.g., they can edit
+        # custom messages for their repo, but not the title.
+        MANAGER = 1
+        # Moderators can moderate user input.
+        MODERATOR = 2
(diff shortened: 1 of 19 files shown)


app/resources/admin_acls.html.template
```diff
@@ -16,15 +16,16 @@
...
 
   <form method="post">
     <input type="hidden" name="xsrf_token" value="{{xsrf_token}}" />
     <p>
       Email address: <input type="text" name="email_address" /><br/>
       Access level:
       <select name="level">
-        <option value="administrator">Administrator</option>
+        <option value="superadmin">Superadmin</option>
+	<option value="manager">Manager</option>
```
christopherestif: (most insightful comment of the year)
maybe add spaces to line it with the other <option>s?

app/views/admin/acls.py
```diff
@@ -47,11 +53,20 @@ def get_params(self):
     def _render_form(self):
         existing_acls = admin_acls_model.AdminPermission.get_for_repo(
             self.env.repo)
+        editable_acls = [
+            acl for acl in existing_acls if
+            self.env.user_admin_permission.compare_level_to(acl.access_level)
+            >= 0]
```
christopherestif: just curious, would it help to define helper methods that encapsulate compare_to >= 0 / < 0, like ".is_at_least(acl...)" or ".is_less_than()"?

app/modelmodule/admin_acls.py
```diff
@@ -20,11 +20,19 @@ class AdminPermission(db.Model):
...
+        # Managers have limited administrative access: e.g., they can edit
+        # custom messages for their repo, but not the title.
+        MANAGER = 1
+        # Moderators can moderate user input.
+        MODERATOR = 2
+
+        # An ordered list of the admin access levels, from lowest to highest
+        # (i.e., each level has the permissions of everything before it in the
+        # list).
+        ORDERING = [MODERATOR, MANAGER, SUPERADMIN]
```
christopherestif: I'm clearly overthinking this - just curious if it's better to rely just on the integer value of the enum (which is used in the template {% if acl.access_level == 0 %}) instead of the array? It looks like duplicated information, and I'm not sure what would happen if either the enum values or the ordering changed in the future?

app/views/admin/base.py
```diff
@@ -61,13 +62,34 @@ def user(self):
         def user(self, value):
             self._user = value
 
+    def _get_user_admin_permission(self):
+        user_repo_admin_object = admin_acls_model.AdminPermission.get(
```
christopherestif: would .get() raise an exception if the object is not found? (instead of returning None) - would .filter(...).first() be more robust?

app/views/admin/base.py
```diff
@@ -82,6 +104,27 @@ def get_params(self):
             self.request,
             post_params={'xsrf_token': utils.strip})
 
+    def _enforce_admin_level(self, min_level):
+        if self.env.user_admin_permission is None:
+            raise django.core.exceptions.PermissionDenied
+        if self.env.user_admin_permission.compare_level_to(min_level) < 0:
+            raise django.core.exceptions.PermissionDenied
+
+    def enforce_moderator_admin_level(self):
```
christopherestif: if you wanted to be super fancy, maybe you could turn these methods into annotations (like @RequireJobOrSuperuser, not that I've seen that name in a codebase or something)

tests/views/test_admin_acls.py
```diff
@@ -33,22 +33,53 @@ class AdminAclsViewTests(view_tests_base.ViewTestsBase):
...
+        self.login_as_manager()
+        self.data_generator.admin_permission(
+            access_level=
+            admin_acls_model.AdminPermission.AccessLevel.SUPERADMIN)
+        self.data_generator.admin_permission(
+            email_address='m@mib.gov',
+            access_level=
+            admin_acls_model.AdminPermission.AccessLevel.SUPERADMIN)
+        res = self.client.get('/haiti/admin/acls/', secure=True)
+        self.assertEqual(len(res.context['editable_acls']), 0)
```
christopherestif: maybe assertEmpty?

tests/views/test_admin_acls.py
```diff
@@ -33,22 +33,53 @@ class AdminAclsViewTests(view_tests_base.ViewTestsBase):
...
+        self.assertEqual(len(res.context['fixed_acls']), 2)
```
christopherestif: maybe assertLen?

tests/views/test_admin_acls.py
```diff
@@ -89,24 +147,40 @@ def test_add_administrator(self):
             'xsrf_token': xsrf_token,
             'email_address': 'l@mib.gov',
             'expiration_date': '2019-04-25',
-            'level': 'administrator',
+            'level': 'superadmin',
         }, secure=True)
-        acls = admin_acls_model.AdminPermission.all()
+        acls = admin_acls_model.AdminPermission.all().filter('repo =', 'haiti')
```
christopherestif: same, I don't know if .all() is required before .filter()

tests/views/view_tests_base.py
```diff
@@ -10,28 +12,58 @@
...
-    TEST_USER_ID = 'k'
-
     def setUp(self):
         super(ViewTestsBase, self).setUp()
         self._xsrf_tool = utils.XsrfTool()
+        self.data_generator.admin_permission(
+            repo_id='global', email_address='z@mib.gov',
+            access_level=
+            admin_acls_model.AdminPermission.AccessLevel.SUPERADMIN,
+            expiration_date=datetime.datetime(2051, 1, 20))
```
christopherestif: does that mean that 32 years from now, tests will start failing? :^p

app/views/admin/base.py
nworden: Oh, yeah, done.

app/modelmodule/admin_acls.py
nworden: I want the order to be independent of the enum values because the enum values are going in the database. E.g., if we later want to add a new level between manager and superadmin, we can make its value 3 but just put it in the right place in `ORDERING` and have everything work.

app/views/admin/acls.py
nworden: If I find myself doing a lot more of this I probably will, but I think there's only one page where we have to vary behavior based on level, so it might be overkill. I'm going to keep it to one function for now.

app/views/admin/base.py
nworden: `get()` is our function, and it calls `get_by_key_name` under the hood, which returns None if nothing's found:
https://cloud.google.com/appengine/docs/standard/python/datastore/modelclass#Model_get_by_key_name
This is a common case in tests (since typically only one or the other of the repo and global permissions will exist), so even if the `get()` implementation gets changed someday we'll catch it.

tests/views/view_tests_base.py
nworden: Probably the `datetime` library will be deprecated or changed in a non-backwards-compatible fashion before then, so someone will have to change this anyway :)

PR #655, opened 2019-04-26
Less awkward params implementation.

nworden:
This changes the params stuff to be more straight-forward, require less boilerplate, and let us access values with `self.params.x` instead of `self.params.get('x')` in cases where we can't assume the value is set.
Also removes the in-class admin enforcement functions, which I replaced with decorators but forgot to delete.


nworden:

PR #663, opened 2019-04-29
Datacheck tasks.

nworden:
Taking a page out of ggive's book.


app/settings.py
```diff
@@ -34,11 +34,13 @@
...
 if os.environ.get('SERVER_SOFTWARE', '').startswith('Development'):
     DEBUG = True
+    DEBUG_PROPAGATE_EXCEPTIONS = True
     # If DEBUG is True and ALLOWED_HOSTS is empty, Django permits localhost.
     ALLOWED_HOSTS = []
     SECURE_SSL_REDIRECT = False
 else:
     DEBUG = False
+    DEBUG_PROPAGATE_EXCEPTIONS = False
     ALLOWED_HOSTS = site_settings.PROD_ALLOWED_HOSTS
```
christopherestif: just curious, is it worth checking the env variable for .startswith('Prod') before allowind PROD_ALLOWED_HOSTS?

app/tasksmodule/datachecks.py
```diff
@@ -0,0 +1,178 @@
...
+
+class PersonDataValidityCheckTask(DatachecksBaseTask):
+    """Checks that person records have valid data."""
+
+    BASE_NAME = 'person_data_validity_check'
+    TASK_PATH = 'check_person_data_validity'
+
+    def _check_person(self, person):
+        if not person.entry_date:
+            self.alert('A person record is missing an entry_date value.')
```
christopherestif: just checking: do we want to add troubleshooting info to the alert message? (like the record id) otherwise receiving the alert would require running the same task that the check performed essentially.

app/tasksmodule/datachecks.py
```diff
@@ -0,0 +1,178 @@
...
+        if not person.original_creation_date:
+            self.alert(
+                'A person record is missing an original_creation_date value.')
+        if (person.author_email and
+                not utils.validate_email(person.author_email)):
+            self.alert('A person record has an invalid author_email value.')
+
+    def post(self, request, *args, **kwargs):
+        del request, args, kwargs  # unused
+        q = model.Person.all(filter_expired=False).filter(
```
christopherestif: maybe add it to the filter() block?

app/tasksmodule/datachecks.py
```diff
@@ -0,0 +1,178 @@
...
+            'repo =', self.env.repo)
+        cursor = self.params.get('cursor', '')
+        if cursor:
+            q.with_cursor(cursor)
+        try:
+            for person in q:
+                next_cursor = q.cursor()
+                self._check_person(person)
+                cursor = next_cursor
```
christopherestif: seems like dark magic to me, but I'll take your word for it

app/tasksmodule/datachecks.py
```diff
@@ -0,0 +1,178 @@
...
+        except runtime.DeadlineExceededError:
+            self._schedule_task(self.env.repo, cursor)
+        except datastore_errors.Timeout:
+            self._schedule_task(self.env.repo, cursor)
+        return django.http.HttpResponse('')
```
christopherestif: maybe add something like 'success' to the response?

app/tasksmodule/datachecks.py
```diff
@@ -0,0 +1,178 @@
...
+                'A note record is missing an original_creation_date value.')
+        if not note.person_record_id:
+            self.alert('A note record is missing a person_record_id value.')
+        if (note.author_email and not utils.validate_email(note.author_email)):
+            self.alert('A note record has an invalid author_email value.')
+        if (note.email_of_found_person and
+                not utils.validate_email(note.email_of_found_person)):
+            self.alert(
+                'A note record has an invalid email_of_found_person value.')
+        if not model.Person.get(self.env.repo, note.person_record_id):
```
christopherestif: just checking, .get() isn't going to raise an exception while .filter(...).first() would return None?

app/tasksmodule/datachecks.py
```diff
@@ -0,0 +1,178 @@
...
+    """Checks that expired person records have been cleared."""
+
+    BASE_NAME = 'expired_person_record_check'
+    TASK_PATH = 'check_expired_person_records'
+
+    def _check_person(self, person):
+        # Check things that were expired yesterday, just in case this job is
+        # ahead of the deletion job.
+        yesterday = utils.get_utcnow() - datetime.timedelta(days=1)
+        if person.expiry_date and person.expiry_date < yesterday:
```
christopherestif: in case you wanted to reduce the nesting levels, you can turn that if block into an early return instead

app/tasksmodule/datachecks.py
```diff
@@ -0,0 +1,178 @@
...
+            for name, prop in person.properties().items():
+                if name not in ['repo', 'is_expired', 'original_creation_date',
+                                'source_date', 'entry_date', 'expiry_date',
+                                'last_modified']:
+                    if getattr(person, name) != prop.default:
```
christopherestif: maybe combine the two ifs into a single test?

tests/tasks/test_datachecks.py
```diff
@@ -0,0 +1,165 @@
...
+class PersonDataValidityCheckTaskTests(task_tests_base.TaskTestsBase):
+    """Tests the person record validity datacheck task.
+
+    We don't have a unit test for the missing entry date case, because
+    entry_date is required by the model (i.e., we cannot test it because we
+    cannot create a Person entity without it to test with).
+    """
+
+    def init_testbed_stubs(self):
+        self.testbed.init_user_stub()
```
christopherestif: just by habit, should we set testbed.tear_down too?

tests/tasks/test_datachecks.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def setUp(self):
+        super(ExpiredPersonRecordCheckTaskTest, self).setUp()
+        utils.set_utcnow_for_test(ExpiredPersonRecordCheckTaskTest._NOW)
+        self.data_generator.repo(repo_id='haiti')
+
+    def test_good_expired_record(self):
+        expiry_date = (
+            ExpiredPersonRecordCheckTaskTest._NOW -
+            datetime.timedelta(days=1, hours=1))
```
christopherestif: since you use it a few times, maybe make it another constant like _YESTERDAY or _PAST?

tests/testutils/data_generator.py
```diff
@@ -34,7 +34,7 @@ class TestDataGenerator(object):
         'home_postal_code': 'M4V 2C5',
         'home_neighborhood': 'Deer Park',
         'author_name': 'Alice Smith',
-        'author_photo': '111-111-1111',
```
christopherestif: good catch, that was a terrible photo

app/tasksmodule/datachecks.py
nworden: The `filter` method is part of the `db` library's implementation, so it isn't going to know what to do with that argument. The `all()` method here is ours.

app/settings.py
nworden: No. From a security standpoint we'd want to err the other direction (limit the allowed_hosts, set debug to false, etc. unless we're sure it's dev).

app/tasksmodule/datachecks.py
nworden: Heh, we do this because jobs can run over GAE's task time limit. The cursor variable should always point to a record at or before the last record we checked, so that the next task scheduled picks up at or before where the last one left off and we ensure that every record gets checked.

app/tasksmodule/datachecks.py
nworden: Oh, yeah, I surely do want to reduce the nesting.

app/tasksmodule/datachecks.py
nworden: I tried it but it got kinda hard to read.

tests/tasks/test_datachecks.py
nworden: `TaskTestsBase` does that already. This function just exists here because these tests require more stubs than some other tests.

app/tasksmodule/datachecks.py
nworden: Good idea, done.

app/tasksmodule/datachecks.py
nworden: I can't, it's not a parameter for the filter method.

app/tasksmodule/datachecks.py
nworden: We actually don't know if it's successful, because it might have overrun the limit and be a continuing task. Since it's meant to be run as a cron job I'm not too concerned about making these responses meaningful.

app/tasksmodule/datachecks.py
nworden: It'll return None if there's no result:
https://cloud.google.com/appengine/docs/standard/python/datastore/modelclass#Class_Methods

tests/tasks/test_datachecks.py
nworden: Done.

PR #674, opened 2019-05-10
Migrate the repo admin index page to Django.

nworden:
Previously the global admin page and the main repo admin page were in the same handler and Django template. This splits it up and moves the repo one to Django to start with.

In this PR I've also added enforcement based on the new ACL system for this page. That's more complex for this page than the others, because manager-level admins and superadmins should both be able to access it, but managers should only be able to change a subset of the fields (I considered splitting those fields off into their own page, but I think it'd be annoying for superadmins to have the language list and the custom messages on separate pages).

Lastly, this rewrites the JS for handling languages and custom messages on that page using React to simplify things. The JS for that page was kinda complex to start with, and it turned into a mess when I tried to handle the manager/superadmin separation (which is why this PR's branch name is adminindex*4*...). This is an isolated use of React (i.e., it's not connected to the new user-facing UI).

### DIFF
--- a/tests/views/test_admin_repo_index.py
+++ b/tests/views/test_admin_repo_index.py
@@ -204,6 +204,7 @@ class AdminRepoIndexViewTests(view_tests_base.ViewTestsBase):
         repo = model.Repo.get_by_key_name('haiti')
         self.assertTrue(repo.test_mode)
         repo_conf = config.Configuration('haiti')
+        self.assertIs(repo_conf.test_mode, True)
         self.assertEqual(
             repo_conf.updated_date,
             utils.get_timestamp(datetime.datetime(2019, 5, 10, 12, 15, 0)))
--- a/tests/views/test_admin_repo_index.py
+++ b/tests/views/test_admin_repo_index.py
@@ -1,0 +1,141 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for the main repo admin page."""
+
+import copy
+
+import django
+import django.http
+import django.test
+
+# pylint: disable=wrong-import-order
+# pylint sometimes thinks config is a standard import that belongs before the
+# django import. It's mistaken; config is our own module (if you run
+# python -c "import config", it produces an error saying config doesn't exist).
+# Filed issue #626 to move us out of the global namespace someday, which would
+# prevent stuff like this.
+import config
+import model
+
+import view_tests_base
+
+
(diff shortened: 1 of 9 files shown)


tests/views/test_admin_repo_index.py
```diff
@@ -0,0 +1,301 @@
...
+        self.data_generator.repo()
+        config.set_for_repo('haiti', **AdminRepoIndexViewTests._PRIOR_CONFIG)
+
+    def test_get_as_manager(self):
+        """Tests GET requests as a manager-level admin."""
+        self.login_as_manager()
+        resp = self.client.get('/haiti/admin/', secure=True)
+        doc = self.to_doc(resp)
+        # Check that a couple of the custom messages appear in the doc content.
+        self.assertTrue('en start msg' in doc.content)
```
christopherestif: or assertIs(..., True), whichever you prefer

tests/views/test_admin_repo_index.py
```diff
@@ -0,0 +1,301 @@
...
+        self.assertTrue('fr view msg' in doc.content)
+        # I tried doing this test the other way around -- checking that every
+        # key in the context was among the set of keys we expect manager-level
+        # admins to have access to -- but that didn't work out well, because
+        # Django seems to include basic stuff (e.g., "True", "None", etc.) as
+        # part of the context (in other words, there's stuff in the context dict
+        # that didn't come from us, and I don't want to have to keep a list of
+        # them).
+        for key in AdminRepoIndexViewTests._SUPERADMIN_ONLY_CONTEXT_KEYS:
+            self.assertFalse(
```
christopherestif: or self.assertNone

tests/views/test_admin_repo_index.py
```diff
@@ -0,0 +1,301 @@
...
+            allow_believed_dead_via_ui='true',
+            min_query_word_length='2',
+            show_profile_entry='true')
+        repo_conf = config.Configuration('haiti')
+        self.assertIs(repo_conf.use_family_name, True)
+        self.assertIs(repo_conf.family_name_first, True)
+        self.assertIs(repo_conf.use_alternate_names, True)
+        self.assertIs(repo_conf.use_postal_code, True)
+        self.assertIs(repo_conf.allow_believed_dead_via_ui, True)
+        self.assertIs(repo_conf.min_query_word_length, 2)
```
christopherestif: assertEqual? nbd

tests/views/test_admin_repo_index.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def test_edit_zero_rating_config(self):
+        self.login_as_superadmin()
+        self._post_with_params(zero_rating_mode='true')
+        repo_conf = config.Configuration('haiti')
+        self.assertIs(repo_conf.zero_rating_mode, True)
+
+    def test_edit_spam_config(self):
+        self.login_as_superadmin()
+        self._post_with_params(bad_words='voldemort')
```
christopherestif: just checking, is testing with any string good enough, or would it be worth checking with multiple words?

tests/views/test_admin_repo_index.py
nworden: Sometimes they're empty dicts. Added a comment.

tests/views/test_admin_repo_index.py
nworden: Here it's fine to test with any string, because the value is getting stored as-is; the code under test here isn't expected to parse it.

PR #677, opened 2019-05-11
Set up configurable static file directory.

nworden:
This also moves a number of fixed static files leftover from the earlier PR.


nworden:

app/views/meta/static_files.py
```diff
@@ -0,0 +1,61 @@
...
+
+import resources
+import views.base
+
+
+ConfigurableStaticFile = collections.namedtuple(
+    'ConfigurableStaticFile',
+    [
+        'content_type',
+    ])
```
eliahhecht: is this whitespace correct? it looks weird.

nworden: I agree it looks weird, but I think that's mainly because it's such a short list (it wasn't really necessary to split this across three lines, but eh, we might add fields later). We do have automated style checking for everything under app/views, and that passed, so I think it's fine.

PR #678, opened 2019-05-11
Migrate static pages to Django.

nworden:
Also makes forms.js.template a regular static file. The reason it was made a template ([to include a translated string in the JS](https://github.com/google/personfinder/commit/b8f2577a2da7c163bde300d3588ebe2e9e1fbae0#diff-8066896dd8efd1d39edcc81808541b17)) no longer applies.


nworden:

PR #685, opened 2019-05-15
Corrections in CONTRIBUTING.md

kshithijiyer:


CONTRIBUTING.md
```diff
@@ -1,4 +1,4 @@
-Want to contribute? Great! First, read this page (including the small print at the end).
+# Want to contribute? Great! First, read this page (including the small print at the end).
```
nworden: This would make for a pretty long heading. If you want to put a heading on this page, I think it should just be `# Contributing`, and this existing line should remain regular text.

kshithijiyer: Done.

PR #691, opened 2019-05-18
Move repo feed to Django.

nworden:
Django has some built-in classes for feeds, but it was kinda restrictive (e.g., you can't set a lang attribute on the top-level title), so I went with lxml instead.


nworden:

tests/testutils/data_generator.py
```diff
@@ -78,14 +79,30 @@ class TestDataGenerator(object):
         'is_valid': True,
     }
 
+    DEFAULT_REPO_UPDATED_DATE = 1557676201
```
christopherestif: maybe add a quick comment on what date that number is?

nworden: Seems perfectly obvious but ok :p

Issue #669, opened 2019-05-03
Future of resource bundles


nworden:
TL;DR: I'm tentatively planning to scale back the resource bundles feature to support just static files. LMK if you disagree or otherwise have concerns!

PF has an interface admins can use to change templates, stylesheets, etc. without pushing a code change. It's not really clear how we'd support it with the React UI, nor is it clear that we'd want to if we could: the ICs couldn't remember the last time they needed to make a change that couldn't be handled by the "custom messages" feature. The RB feature adds a substantial amount of complexity which has resulted in bugs when it was used, requires a bunch of datastore queries even for pages without dynamic content, and represents a pretty broad attack surface from a security standpoint. I'm sure all of that was a good trade for most of PF's lifetime, but if the feature's not being used anymore, it doesn't seem like a good trade now.

All that said, those issues mostly apply to pushless editing of *templates*. Static files (e.g., a header image) are a different story:
- Control over static files could be a really useful feature for third-parties running PF, and could help us move away from hard-coded Google branding (even if you turn off "Google branding" in the admin interface, a lot of Google-specific stuff remains). Pushless editing of templates is for the case where someone wanted to make a change without the time/trouble of a push, but with static files it also addresses a different use case: where the admin isn't sufficiently technically inclined to do a push (in these cases, it seems unlikely they'd want to edit a template file or something like that).
- It'd involve fewer datastore queries, plus we could cache them all with a long TTL.
- We can certainly support it with the React UI.
- The security concerns are no different than those related to user-uploaded images.

PR #692, opened 2019-05-18
Use Django's template loader for Django pages.

nworden:
Next step for issue #669.
The webapp2 pages will continue to use `resources.get_rendered` for the top-level template; doesn't seem worth it to change that since we're moving everything off webapp2 anyway.


nworden:

PR #695, opened 2019-05-20
Factor create_person function out of handler.

nworden:
This will let us share it with the React UI (probably the API too, but I haven't gotten there yet).


app/create.py
```diff
@@ -37,6 +37,254 @@ def days_to_date(days):
...
+    try:
+        # If a photo was uploaded, create a Photo entry and get the URL
+        # where we serve it.
+        if photo is not None:
+            photo, photo_url = create_photo(photo, repo, url_builder)
+        if note_photo is not None:
+            note_photo, note_photo_url = create_photo(
+                note_photo, repo, url_builder)
+    except PhotoError, e:
+        raise CreationError(_('There seems to be a problem with your uploaded photo.'))
```
christopherestif: just checking, is it worth specifying which photo/note_photo failed? (in case they can upload both?)

christopherestif: do we know anything from the PhotoError on what could be the problem, and is it worth passing to the user? (eg too big, too small, wrong format, etc.)

nworden: I don't think we ever give users a form with both options, so they should only be providing one or the other. (Related: just filed issue #696 because I'm not sure this needs to handle a `note_photo` field at all.)

nworden: Oh yeah, done.

PR #698, opened 2019-05-22
Real implementation of FE API create handler.

nworden:
Also fleshes out the create page UI, updates to the latest versions of Material libraries, and adds the description field to the create page (which was blocked on a Material bug that's now been fixed).


app/views/frontendapi.py
```diff
@@ -81,3 +82,104 @@ def _result_to_dict(self, person):
...
+                'source_name': utils.strip,
+                'source_url': utils.strip,
+                'status': utils.validate_status,
+                'text': utils.strip,
+                'own_info': utils.validate_yes,
+            })
+        if request.method == 'POST':
+            profile_urls = []
+            profile_field_index = 0
+            while True:
```
eliahhecht: this infinite loop makes me nervous. Could you not just loop over the keys in self.request.POST instead, and take the ones that match `profile-url-[\d]+`?

ui/src/pages/Create.js
```diff
@@ -389,11 +461,12 @@ class Create extends Component {
     );
   }
 
-  renderTextFieldAndInput(formKey, inputName, labelMessage) {
+  renderTextFieldOrAreaAndInput(formKey, inputName, labelMessage, textArea) {
```
eliahhecht: as discussed offline, make this two separate methods for clarity

app/views/frontendapi.py
nworden: I'd prefer not to do that, because a) it would make it inconvenient to maintain their order and b) it'd be marginally less efficient. Instead I changed it to a for loop over range(size of POST dict) (since we certainly don't need to check values higher than that).

ui/src/pages/Create.js
nworden: Done.

app/views/frontendapi.py
nworden: JK, request.POST doesn't have a size function, and you can't get the size without having it iterate through everything afaict, so I just hard-coded a limit.

ui/src/pages/Create.js
```diff
@@ -477,13 +477,32 @@ class Create extends Component {
...
+    return (
+        <TextField
+          label={this.props.intl.formatMessage(labelMessage)}
+          outlined
+        >
+          <Input
+            name={inputName}
+            value={this.state[formKey]}
+            onChange={(e) => this.setState({[formKey]: e.target.value})} />
+        </TextField>
```
eliahhecht: might be worth extracting a renderInput method, to deduplicate? i'm not sure, up to you.

nworden: I mean, it's four lines, so I think I'm just going to leave it.

PR #702, opened 2019-05-24
Allow any hosts for local development server.

gimite:
@nworden I forgot if we have considered this, but how about allowing any hosts for local development server?

In my development environment, the browser and the server run on different machines, so I cannot access the server with "localhost". We may alternatively provide a way to explicitly specify the host name, but I'm not sure if we need such security for local development.


nworden:
Sure, SGTM!

PR #705, opened 2019-05-31
Person view implementation for FE API

nworden:
This also refactors the profile page functions a little so they can be shared with the Django app.


nworden:
Not sure why this hasn't been sent to Travis, but tests passed locally.

PR #707, opened 2019-06-06
Drop disable-notes feature.

nworden:
Outside the test repo, this feature has been used exactly once in the last year.


nworden:
For now I'm preserving the code that prevents adding notes, just in case someone uses the feature for a new record before this gets pushed out. I'll clean it up in a later PR, as part of some refactoring I'm doing for the new UI.

nworden:
You might be wondering why I changed the given names used for a couple of tests. Those tests failed with this change when I ran all server tests (`tools/server_tests`), but each passed individually. We've had issues with tests (specifically in `person_note_tests`) interfering with each other before, and in this case it turns out changing the names resolves it. If that file was going to be around in the long term I'd try to dig into what's going on, but it's getting gradually split up as I do the Django migration, so it doesn't seem worth debugging further.

eliahhecht:
LGTM

Issue #194, opened 2015-10-27
Clearer error message for SMS messages to a number not associated with any repos


gimite:
e.g., "This number is not associated with any active Person Finder repository"

kshithijiyer:
Could you explain more about what is to be changed and where?

nworden:
TL;DR: We'd like to send back a human-readable error message to the user [here](https://github.com/google/personfinder/blob/174e4408a39500d4bd1d315930d55c9195590aa2/app/api.py#L715), instead of just logging an error like we do now.

PF has an SMS feature that lets you make simple queries or record additions by texting "search <name>" or "I am <name>" to a designated number. The PF code for this is in `app/api.py`, specifically in the `HandlerSMS` class. The server that actually receives and sends text messages is separate, since we don't do that directly on App Engine; the SMS server forwards messages to PF itself, along with fields indicating both the sending and the recipient/destination phone numbers, and then sends an SMS back with PF's reply message. (This is also convenient for testing, since you can test Person Finder without having to really send text messages; you can just send HTTP requests like the ones the SMS server sends to PF.)

We support receiving messages sent to multiple different phone numbers, for a couple reasons. One is just so that we can provide local phone numbers for different country. Another reason is so that users don't have to specify what repository to use in their message: phone numbers are mapped to repositories, so we can tell what repository to use based on the destination number for incoming SMSes. This is defined by a global config setting with the key `sms_number_to_repo`, which can be configured on the global admin page (at /global/admin). When `HandleSMS` gets a request, it uses that map to look up which repository to use.

We try (with varying degrees of success, long story there) to maintain numbers for a bunch of countries set up in advance, so that we can launch SMS support more quickly if/when we activate PF for a disaster there. However, that means that, most of the time, those numbers aren't associated with any particular repository, and the existing code doesn't handle that well: it logs an error for us (developers) to see, and doesn't return anything at all to the SMS server. As a result, the SMS server thinks everything's broken and sends a message back saying the service is unavailable. This is pretty confusing, especially for folks on the team that maintain the SMS infrastructure (which is separate from the PF team), because it looks to them like there might be a problem on their side. So, we'd instead like PF to send back a message saying the number's not associated with any repository, so it's clear that the message did make it to Person Finder.

PR #708, opened 2019-06-07
Modifying error messages to be more user friendly

kshithijiyer:
Adding clearer error message for SMS messages
to a number not associated with any repos
fixes:#194
Signed-off-by: kshithijiyer <kshithij.ki@gmail.com>

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -680,10 +680,10 @@ class HandleSMS(BaseApiHandler):
             self.info(
                 403,
                 message=
-                    '"key" URL parameter is either missing, invalid or '
-                    'lacks required permissions. The key\'s repo must be "*", '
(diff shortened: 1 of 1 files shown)


app/api.py
```diff
@@ -713,8 +713,8 @@ def post(self):
             self.info(
                 400,
                 message=
-                    'The given receiver_phone_number is not found in '
-                    'sms_number_to_repo config.',
+                    ('The given phone number: %s is not found in '
```
nworden: I think this message could be made even less technical. How about "You've reached Person Finder, but there's not a repository assigned for %s." % receiver_phone_number
That way people would know the number's a Person Finder number and the message got to us.

kshithijiyer: Done.

Issue #592, opened 2019-03-20
Tell users if they enter an invalid age


nworden:
PFIF requires that age values either be a number or arrange of numbers (formatted as nn-nn). We currently enforce that requirement silently: if you enter something invalid, the form submission will go through, and we'll just drop the age value. We should tell users if the age they submitted is invalid and give them a chance to fix it.

gimite:
It may also be nice if we are a bit more flexible about the accepted format e.g., accepting full with numbers (e.g., "３０−３５") as well, which is commonly used in CJK. Applying Unicode NFC normalization would probably be sufficient but I haven't confirmed it.

yashLadha:
Would like to take up the issue @nworden @gimite

gabru-md:
if this is a good first issue then can I take it up @yashLadha ?

yashLadha:
No problem in that @gabru-md :sweat_smile:

pollyMath:
If this is not worked on and it is still relevant, I would like to take it :)

routrohan:
Is this issue still open?
If so, I would like to work upon it.

nworden:
Yeah, there hasn't been follow-up on the above PR, so feel free to take it. Thanks!

pollyMath:
sorry for great delay, have done another commit to my PR

darshang99:
I'm a beginner to this project can I work on this issue and point me to the resources. Is this still open? I would like to work on it.

PR #709, opened 2019-06-07
592: Tell users if they enter an invalid age

pollyMath:
Worked on https://github.com/google/personfinder/issues/592 ;
It is its most simplest validation form (which doesn't consider the CJK or if the order of the ages is proper (i.e. "34-0" might be considered invalid if the database doesn't handle that)).

### DIFF
--- a/app/resources/static/fixed/forms.js
+++ b/app/resources/static/fixed/forms.js
@@ -361,6 +361,22 @@ function validate_fields() {
     }
   }
 
+  // Validate age format
+  var age_field = $('age');
+  var age_value = age_field.value;
+
+  if (age_value != "") {
+    var curated_age_value = age_value.replace(/\s/g,'');
+    var valid_age_re = /^(\d+-?)+\d+$/;
+
+    var isAgeFormatValid = valid_age_re.test(curated_age_value);
+
+    if (!isAgeFormatValid) {
+      alert("The age entered is not in a valid format.");
+      return false;
+    }
+  }
+
   return true;
 }
 


pollyMath:
I signed it!

app/resources/static/fixed/forms.js
```diff
@@ -361,6 +361,22 @@ function validate_fields() {
     }
   }
 
+  // Validate age format
+  var age_field = $('age');
+  var age_value = age_field.value;
+
+  if (age_value != "") {
+    var curated_age_value = age_value.replace(/\s/g,'');
+    var valid_age_re = /^(\d+-?)+\d+$/;
```
nworden: I think the question mark should go on the outside of the parens, and the plus sign there now can be dropped, like this:
`^(\d+-)?\d+$`

nworden: It might be worth making this RE a constant outside the function, though it's ok here too.

app/resources/static/fixed/forms.js
```diff
@@ -361,6 +361,22 @@ function validate_fields() {
...
+
+    var isAgeFormatValid = valid_age_re.test(curated_age_value);
+
+    if (!isAgeFormatValid) {
+      alert("The age entered is not in a valid format.");
```
nworden: We'd like to be consistent about how we display errors, so could you please handle this like the others. The improper email address one is probably a good example you could follow: it's displayed with code a little bit above in this file, and the HTML is in `app/resources/errors.html.template`.

pollyMath:
@googlebot I fixed it.

pollyMath:
@googlebot I fixed it.

PR #714, opened 2019-06-09
Code for moving dictionary value under app/

ashutoshnarayan:


nworden:
Ashutosh, I really appreciate your persistence in this. I have a few comments though:
* I would continue to strongly recommend that you do this in multiple smaller PRs (e.g., starting with the data files) instead of all at once. There are multiple other people working on Person Finder, and moving a large number of files simultaneously is likely to cause some significant merge conflicts. There are actually already some merge conflicts, because I've moved a number of handlers over the past couple months (e.g., most of the app/admin_* files aren't there anymore).
* It looks like this PR moves stuff without updating references to them. We always want the code on the master branch to build and run, so it's important to update the references for anything that's moved in the same PR.
* I happen to have plans to do work on all the handlers for unrelated reasons (web framework changes), and it's convenient to move them then, so it might be easier if I just do that as I get to each handler. You don't need to move the handlers yourself.

gimite:
Talked with him directly and he said he will try cleaning up his repository to create a clean pull request. Closing this.

Issue #157, opened 2015-09-02
Clean up duplicate code in create.py and add_note.py


Jessie-Newman:
https://github.com/google/personfinder/pull/155 separates create.py into two files, create.py and add_note.py. There's still a lot of duplicate code between the two files that can be cleaned up.

nworden:
I'll do this as part of the refactoring I'm doing for the UI work.

PR #715, opened 2019-06-10
Factor out most add-note code into create.py.

nworden:
This factors most of the code for adding notes out into functions in create.py, so they can be reused by the React UI's API. There's still some duplication around subscriptions (that's worthy of its own PR) and photos (which I'm hoping will be more straight-forward once I can factor everything out).
This intentionally doesn't support note-disabling in the React UI, because we're dropping that feature.
Helps address issue #157.


nworden:
Sent to Christophe.

app/create.py
```diff
@@ -287,6 +274,112 @@ def create_person(
...
+            phone_of_found_person=phone_of_found_person,
+            last_known_location=last_known_location,
+            text=text,
+            photo=photo,
+            photo_url=photo_url,
+            spam_score=spam_score,
+            confirmed=False)
+        note.put_new()
+        raise FlaggedNoteException(note)
+    else:
```
christopherestif: just in case you wanted to save a level of indentation, you can remove this "else" (it might be more of a style preference, so it's your call)

app/create.py
```diff
@@ -287,6 +274,112 @@ def create_person(
...
+            email_of_found_person=email_of_found_person,
+            phone_of_found_person=phone_of_found_person,
+            last_known_location=last_known_location,
+            text=text,
+            photo=photo,
+            photo_url=photo_url)
+        note.put_new()
+        # Specially log notes that make a person dead or switch to an alive
+        # status.
+        if status == 'believed_dead':
```
christopherestif: just a generic comment on hard-coded string statuses vs constants if that's relevant

app/views/frontendapi.py
```diff
@@ -312,3 +312,68 @@ def post(self, request, *args, **kwargs):
...
+            author_name=self.params.author_name,
+            author_email=self.params.author_email,
+            author_made_contact=self.params.author_made_contact,
+            text=self.params.text)
+        photo, photo_url = (None, self.params.photo_url)
+        if self.params.photo is not None:
+            try:
+                photo, photo_url = photo.create_photo(
+                    self.params.photo, self.env.repo, self.build_absolute_uri)
+            except photo.PhotoError, e:
```
christopherestif: you can replace ", e" with "as e" if you want

tests/views/test_frontendapi.py
```diff
@@ -275,3 +275,24 @@ def test_get_with_deactivated_repo(self):
...
+        self.client.post(
+            '/haiti/d/add_note',
+            data={
+                'id': self.person.record_id,
+                'author_name': 'Adam',
+                'text': 'Here is some text.',
+            },
+            secure=True)
+        notes = model.Person.get('haiti', self.person.record_id).get_notes()
+        self.assertEqual(len(notes), 1)
```
christopherestif: you can use assertLen

app/views/frontendapi.py
nworden: Done.

tests/views/test_frontendapi.py
nworden: That's not a function in our testing library.

PR #719, opened 2019-06-17
Map features on new UI's create page

nworden:


nworden:
There's an issue where the "Last known location" label doesn't always move up to the top of the text area like it's supposed to; afaict this is a problem with the React component so I'm not worrying about it for now.

nworden:
Sent to Eliah.

app/views/frontendapi.py
```diff
@@ -98,6 +98,9 @@ def get(self, request, *args, **kwargs):
                 'repoId': self.env.repo,
                 'title': repo_title,
                 'recordCount': self._get_person_count(self.env.repo),
+                'mapDefaultCenter': self.env.config.get('map_default_center'),
+                'mapDefaultZoom': self.env.config.get('map_default_zoom'),
+
```
eliahhecht: nit: stray newline?

ui/src/components/LocationFieldset.js
```diff
@@ -0,0 +1,231 @@
...
+
+class LocationFieldset extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      showMap: false,
+      haveStartedLoadingMapScript: false,
+      haveFinishedLoadingMapScript: false,
+    };
+    this.mapsApiDisabled = ENV.maps_api_key == null || ENV.maps_api_key == '';
```
eliahhecht: Empty string is falsy in JS, so if you wanted you could just do `this.mapsApiDisabled = !!ENV.maps_api_key` (not sure if this is considered good style or not).

Also, I find it easier to understand to have "positive"-valued boolean variables -- so in this case IMO it would be better to have the variable be mapsApi**En**abled.

ui/src/components/LocationFieldset.js
```diff
@@ -0,0 +1,231 @@
...
+            }
+            const resultGeoLocation = results[0].geometry.location;
+            this.props.onLocationLatLngUpdate(
+                [resultGeoLocation.lat(), resultGeoLocation.lng()]);
+          });
+    }
+  }
+
+  render() {
+    var showHideMapButton = null;
```
eliahhecht: maybe use `let` instead of `var`?

ui/src/components/LocationFieldset.js
```diff
@@ -0,0 +1,231 @@
...
+              {this.props.intl.formatMessage(MESSAGES.hideMap)}
+            </Button>
+          ) :
+          (
+            <Button
+              className='pf-button-primary'
+              type='button'
+              onClick={() => this.setState({showMap: true})}>
+              {this.props.intl.formatMessage(MESSAGES.showMap)}
+            </Button>
```
eliahhecht: Can you reduce the duplication here? Something like:

```js
const newMapState = !this.state.showMap;
showHideMapButton = (<Button>...onclick={() => this.setState({showMap: newMapState})...</Button>);
```

ui/src/components/LocationFieldset_test.js
```diff
@@ -0,0 +1,252 @@
...
+            mapDefaultCenter={[40.6782, -73.9442]}
+            mapDefaultZoom={10}
+            />);
+    wrapper.update();
+    wrapper.find('button').at(1).simulate('click');
+    expect(mockLoadExternalScript).toHaveBeenCalledWith(
+        'https://maps.googleapis.com/maps/api/js?key=abc123',
+        expect.anything());
+    wrapper.unmount();
+  });
```
eliahhecht: Add a test that the map script is not loaded twice if the user shows the map twice?

ui/src/testing/mock-google-module.js
```diff
@@ -0,0 +1,81 @@
...
+class LatLng {
+  constructor(lat, lng) {
+    this.lat = lat;
+    this.lng = lng;
+  }
+}
+
+class Bounds {
+  constructor() {
+    // TODO(nworden): see if there's a better way to do this. seems like there
```
eliahhecht: did you mean to leave this unresolved?

nworden: Yeah, for now. It's only in this file so I don't think it's PR-blocking.

ui/src/components/LocationFieldset_test.js
nworden: Done.

ui/src/components/LocationFieldset.js
nworden: Done.

ui/src/components/LocationFieldset.js
```diff
@@ -58,7 +58,7 @@ class LocationFieldset extends Component {
       haveStartedLoadingMapScript: false,
       haveFinishedLoadingMapScript: false,
     };
-    this.mapsApiDisabled = ENV.maps_api_key == null || ENV.maps_api_key == '';
+    this.mapsApiEnabled = ENV.maps_api_key;
```
eliahhecht: I think it would be worth coercing to a boolean here (for clarity/to prevent people from doing weird things with this field in the future)

nworden: Done.

PR #723, opened 2019-06-19
Static pages (user guide, responders page) in React UI

nworden:
I'm deliberately punting on i18n, because a) I expect product folks to make some language changes and b) IDK if we want to include screenshots. The existing page has screenshots, but just in English (and maybe Japanese). If we do end up deciding to include screenshots, I'll see if we can automate screenshots with React so we have language-specific ones.


ui/src/pages/HowToPage.js
```diff
@@ -0,0 +1,172 @@
...
+
+const MESSAGES = defineMessages({
+  helpDocument: {
+    id: 'HowToPage.helpDocument',
+    defaultMessage: 'Help document',
+    description: 'A header for a page with help documentation.',
+  },
+  intro: {
+    id: 'HowToPage.intro',
+    defaultMessage: ('Person Finder can be used by anyone from a PC or mobile '
```
christopherestif: just curious, s/PC/computer?

ui/src/pages/HowToPage.js
```diff
@@ -0,0 +1,172 @@
...
+                  <p>Enter the name of the person you're looking for and then hit the [Enter] key.</p>
+                </li>
+                <li>
+                  <b>Results</b>
+                  <p>If you see a search result that might be the person you're looking for, click it to open a page with more details about that record.</p>
+                </li>
+                <li>
+                  <b>Adding a record</b>
+                  <p>If you can't find a record for the person you're looking for, you can add a record seeking information.</p>
+                  <p>(Note from Nick: It might be helpful to generalize the section below, so that we can tell people to follow those directions here.</p>
```
christopherestif: just checking, is it meant to be visible?

ui/src/pages/HowToPage.js
nworden: I'll ask about that when we get into this with product folks, but the current language says PC.

ui/src/pages/HowToPage.js
nworden: Thanks for catching that. I'd added this for a demo I sent around internally, but it shouldn't be checked in.

PR #725, opened 2019-06-20
Get homepage repo cards centered.

nworden:
This switches to using flexbox instead of grid for the repo cards (seems grids aren't meant for stuff that needs to be flexibly centered like this).


nworden:
Sent to Avi.

PR #726, opened 2019-06-20
A batch of style fixes for the global homepage.

nworden:
* Un-raises the buttons and gives them the right outline color.
* Unraises the cards.
* Fixes the background color of the third-party cards.


nworden:
Sent to Christophe.

PR #727, opened 2019-06-21
Fix photo upload handling in Django.

nworden:
Django handles uploaded files differently than webapp2.
Also refactors create.py to import the photo module. This is what's
recommended by the style guide, and this is a convenient (by convenient,
I mean mandatory) time to do that because the mocking library doesn't
seem to handle the situation well otherwise.


app/utils.py
```diff
@@ -296,6 +296,17 @@ def validate_image(bytestring):
         return False
 
 
+def validate_django_image(uploaded_file):
+    try:
+        image = None
+        if uploaded_file:
+            image = images.Image(uploaded_file.read())
+            image.width
```
eliahhecht: this line looks weird. are you just making this call to validate that the image was constructed properly? if it wasn't, would we even reach this line?

app/utils.py
```diff
@@ -296,6 +296,17 @@ def validate_image(bytestring):
         return False
 
 
+def validate_django_image(uploaded_file):
```
eliahhecht: add documentation for what this method returns?

app/utils.py
nworden: That's my assumption, yeah. This function is really just a Django variant on an existing function (validate_image, above); I added a note about that in the comment.

app/utils.py
nworden: Done.

PR #729, opened 2019-06-26
Style the photo upload field.

nworden:
This also corrects some of the photo URL handling code (although I'd documented it at the time, I'd forgotten how I'd defined the URL builder's signature...)


nworden:
Sent to Christophe.

ui/src/pages/Create.js
```diff
@@ -299,38 +305,54 @@ class Create extends Component {
...
+      return (
+          <div className='create-photouploadline'>
+            {photoInputField}
+            <span className='mdc-typography--body1'>
+              {this.state.formPhotoFile.name}
+            </span>
+            <img
+                className='create-photodeletebutton'
+                src='/static/icons/maticon_clear.svg'
+                onClick={(e) => {this.setState({formPhotoFile: ''})}} />
```
christopherestif: is the image supposed to be clicked? There might be some accessibility limitations of using onClick on a img tag with no discoverability except styling, maybe adding some aria roles (or wrapping the image in a link?) could help?

nworden: Oh yeah, good call. I have another PR in progress to improve accessibility, but didn't keep it in mind here.

PR #730, opened 2019-06-27
Batch of accessibility fixes

nworden:
I'll do a more thorough accessibility review when things are closer to being finalized, but I wanted to at least address some obvious stuff.


nworden:
Sent to Eliah.

PR #731, opened 2019-06-27
Move ui tests to Python 3 Travis run.

nworden:
We have our tests split up into Python 2 and Python 3 Travis runs. It
doesn't really matter which one the UI tests run in, because they don't
involve any Python, but I think there's two reasons it's better to put
them in the Py3 run:
- There's only one Py3 run (we test with two different versions of
  Python 2), and we only need the UI tests to run once.
- The Python 3 tests are faster (because there's fewer of them). By
  moving the UI tests to the Py3 run we'll make things more balanced,
  and we won't have to wait as long to find out if a UI test is failing.


nworden:
Sent to Avi.

PR #736, opened 2019-07-01
Update SearchBar value when props change.

nworden:
There's currently a bug where if you make a query, make a new query, and
then hit the back button, the value in the search bar doesn't get
updated to the original query. This is because it doesn't pay attention
to updates to props. This commit fixes that.


nworden:

PR #738, opened 2019-07-02
Fix page header spacing.

nworden:
This also renames all the CSS classes. I did some refactoring and renamed the component to PageHeader in an earlier PR, but forgot the CSS.


nworden:

PR #751, opened 2020-06-15
Limit to 20 batches of 100 records per count task.

nworden:


app/tasks.py
```diff
@@ -185,6 +186,9 @@ def get(self):
                             break
                     # And put the updates at once.
                     counter.put()
+                    batches_done += 1
+                self.add_task_for_repo(self.repo, self.SCAN_NAME, self.ACTION)
```
eliahhecht, line 190: I don't quite understand this change -- I get the limit on the batching, but why do you now call add_task_for_repo where you didn't before?

Issue #752, opened 2020-07-09
"About Person Finder" link is 500


gimite:
e.g., "About Person Finder" link in the hamberger menu in https://google.org/personfinder/test links to https://google.org/personfinder/global, which returns 500. It should probably link to https://google.org/personfinder/ instead.

nworden:
I don't think I'll have time tomorrow (and this weekend I finally need to do the tax returns I've procrastinated on), but I should be able to get to this early next week.

gimite:
Thanks! It is not urgent, so take your time. Let me know if you are busy and have no time working on this. I will probably have some bandwidth, so I may be able to take this kind of small fixes.

gimite:
This is the mapping from URL patterns to their handlers:
https://github.com/google/personfinder/blob/master/app/app.yaml

When the URL is mapped to `script: wsgi.application`, it is handled by Django. When it is mapped to `script: main.py` (i.e., when it doesn't match other patterns), it is handled by app/main.py, our traditional main handler.

URL "/global" should be alias of /global/home.html (and also "/"), and /global/home.html is handled by Django but /global doesn't look so. This looks like an issue. 

You also need to let Django handle /global properly. This looks configured here:
https://github.com/google/personfinder/blob/master/app/urls.py
I'm not familiar yet about how Django routing is configured though.

ryosuketc:
@gimite Thank you for sharing the findings. Very helpful to catch up.

All the URL patterns (`/`, `/global/home`, `/global`) should be handled by `GlobalIndexView`, but `/global` does not seem to be listed and configured in `url.py`.
At a glance, it would be quick to fix it by adding just a few lines in the `url.py`.

Not very critical but a remaining concern could be at which version it stopped working and what kind of changes were made then. I will try to chase it just in case.

gimite:
Thanks! It was broken a while ago, so we don't really need to know when it got broken :) You can investigate it if you are curious though. All traffic used to be served from main.py, so it sounds likely that it was broken when we migrate it to Django.

ryosuketc:
Noted. Then I will try to fix it anyway.

PR #760, opened 2020-10-28
Fix 500 error on about person finder - fix issue #752

ryosuketc:
The underlying issue was the following:
1.  `global/` URL pattern was passed to `main.py`, which is outdated.
It should be handled by Django.
2.  The pattern was not configured in Django url patterns - 'urls.py`

### DIFF
--- a/app/app.yaml
+++ b/app/app.yaml
@@ -45,6 +45,10 @@ handlers:
   script: wsgi.application
 - url: /setup_datastore/?
   script: wsgi.application
+- url: /global/?
+  script: wsgi.application
+- url: /personfinder/global/?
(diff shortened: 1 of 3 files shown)


ryosuketc:
@gimite The code itself looks okay but it might be affected by another issue, which I reported here.
Archive file for python 2.7.12 is gone in storage.googleapis.com/travis-ci-language-archives/ · Issue #761 · google/personfinder
https://github.com/google/personfinder/issues/761

gimite:
Thanks! Merged.

CI looks already failing, so it would be OK for now. (We should eventually look into it and hopefully fix it though...)

Issue #764, opened 2020-11-11
Packages are missing in GettingStarted and server tests fail


ryosuketc:
`tools/server_tests/ fails. It seems that some packages are missing from [GettingStarted](https://github.com/google/personfinder/wiki/GettingStarted).
The missing ones have to be installed in a global environment, instead of `app/vendors`.

`flake8` might be one of them - https://github.com/google/personfinder/issues/762.

```
======================================================================= ERRORS =======================================================================

___________________________________________ ERROR collecting tests/server_test_cases/person_note_tests.py ____________________________________________
ImportError while importing test module '$HOME/repos/pf/tests/server_test_cases/person_note_tests.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/server_test_cases/person_note_tests.py:22: in <module>
    from parameterized import parameterized
E   ImportError: No module named parameterized
================================================================== warnings summary ==================================================================
tests/server_tests_base.py:69
  /usr/local/google/home/rtachibana/repos/pf/tests/server_tests_base.py:69: PytestDeprecationWarning: the `pytest.config` global is deprecated.  Please use `request.config` or `pytest_configure` (if you're a pytest plugin) instead.
    hostport = pytest.config.hostport

tests/server_tests_base.py:70
  /usr/local/google/home/rtachibana/repos/pf/tests/server_tests_base.py:70: PytestDeprecationWarning: the `pytest.config` global is deprecated.  Please use `request.config` or `pytest_configure` (if you're a pytest plugin) instead.
    mail_server = pytest.config.mail_server

-- Docs: https://docs.pytest.org/en/latest/warnings.html
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
======================================================== 2 warnings, 1 error in 0.31 seconds =========================================================
```

ryosuketc:
-   `parameterized` seems to be the only missing package, which I added to [GettingStarted](https://github.com/google/personfinder/wiki/GettingStarted).
-   Code in https://github.com/google/personfinder/issues/763 passes almost all tests.
    -   Except for the image-related ones, which we can disregard for now as chatted offline with @gimite (with future migration to Python 3 etc in mind).

@gimite 
[...]

Issue #763, opened 2020-11-11
Skip and optimize "Updating translations" in tools/server_tests


ryosuketc:
`tools/server_tests` consists of two parts:
1.  Update translations
2.  Run tests (`pytests`)

1 is not always necessary and we sometimes want to skip 1 temporarily.

As a side note,  the update message looks taking much longer than before, and it looks like it is trying to update unnecessary files. We may want to fix it too. I believe we only need to update our messages, not inside app/vendors.

TODO:
1.  An optional command line argument for `tools/server_tests` might make sense in order to skip update translations.
2. Skip updating translations in `app/vendors`. for faster updates (might file a separate issue for this later).

ryosuketc:
Following up with @gimite internally.

ryosuketc:
Confirmed with him that this implementation and its output for the second TODO item.
The changes will be sent after https://github.com/google/personfinder/issues/764 is addressed and confirmed to pass all server tests.

Before the change
```
% tools/update_messages
 
--- Updating translations
processing locale en
No messages have changed.
processing file django.po in $HOME/repos/pf/app/vendors/django/contrib/admindocs/locale/gd/LC_MESSAGES
... (omitted) ...
processing file django.po in $HOME/repos/pf/app/locale/en/LC_MESSAGES
... (omitted) ...
```

After the change:
```
% tools/update_messages     
processing locale en
No messages have changed.
```

ryosuketc:
TODO 1 was completed and merged.
TODO 2 is reverted and on hold until we migrate to Python 3.
https://github.com/google/personfinder/pull/768

PR #765, opened 2020-11-19
Fix slow server tests #763 #764

ryosuketc:
Optimize server tests. The PR also includes a small fix to support `virtualenv`.
- Support an optional argument --skip_update_translations #763
- Skip app/vendors from compilemessages #763
- Add venv/ to .gitignore to support virtualenv #764

Example usage and outputs:
```
 % tools/server_tests --skip_update_translations
--- Skipped updating translations

--- Running server tests
... (server tests continue) ...
```

```
 % tools/server_tests                                                                                       

--- Updating translations
... (update translations) ...

--- Running server tests
... (server tests continue) ...
```

### DIFF
--- a/tools/server_tests
+++ b/tools/server_tests
@@ -31,11 +31,11 @@ pushd "$(dirname $0)" >/dev/null && source common.sh && popd >/dev/null
 cd "$(dirname $0)/.."
 
 # Parse arguments and set flags
-SKIP_UPDATE_TRANSLATIONS='false'
+SKIP_UPDATE_TRANSLATIONS=false
 for arg in "$@"; do
   shift
   if [[ "${arg}" == "--skip_update_translations" ]]; then
-    SKIP_UPDATE_TRANSLATIONS='true'
+    SKIP_UPDATE_TRANSLATIONS=true
     continue
   fi
   # All arguments other than "--skip_update_translations" remains intact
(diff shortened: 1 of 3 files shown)


tools/server_tests
```diff
@@ -30,9 +30,26 @@ pushd "$(dirname $0)" >/dev/null && source common.sh && popd >/dev/null
 # directory.
 cd "$(dirname $0)/.."
 
+# Parse arguments and set flags
+SKIP_UPDATE_TRANSLATIONS='false'
```
gimite, line 34: Can you use boolean `true` and `false` instead of string `'true'` and `'false'`?
https://stackoverflow.com/a/2953673

gimite (changes requested):
It is OK this time (especially since the PR is still small), but it is generally recommended to keep a PR small so that a single PR does a single thing (e.g., resolving a single issue) rather than combining multiple independent changes to a single PR.

ryosuketc:
> It is OK this time (especially since the PR is still small), but it is generally recommended to keep a PR small so that a single PR does a single thing (e.g., resolving a single issue) rather than combining multiple independent changes to a single PR.

Thank you for the feedback.
Indeed I used different branches to make the changes but I thought the PR was too small... I learn that a PR could never be too small. Appreciated.

ryosuketc:
Mentioned in https://github.com/google/personfinder/issues/764 as well but it might be worth noting here in the PR itself too:
-   This code passes all tests except for:
    -   Python 3 unit tests
    -   (server tests) the image-related ones, which we can disregard for now as chatted offline with @gimite (with future migration to Python 3 etc in mind).

tools/server_tests
ryosuketc: Done.

gimite (approved):

gimite:
LGTM. Merged. Thanks!

ryosuketc:
Send as a new PR.
https://github.com/google/personfinder/pull/767

PR #767, opened 2020-11-26
Revert --ignore=vendors in compilemessages due to version compatibility

ryosuketc:
This reverts commit 0af534c.

--ignore option for compilemessages is supported in greater than django 3.0.
https://docs.djangoproject.com/en/3.1/ref/django-admin/#cmdoption-compilemessages-ignore

makemessages support --ignore even on django 1.x
https://docs.djangoproject.com/en/3.1/ref/django-admin/#cmdoption-makemessages-ignore

Indeed, our django version is 1.11.29, which is the last version for Python 2.x
Django 1.11 is the last version to support Python 2.7. Support for Python 2.7 and Django 1.11 ends in 2020.
https://docs.djangoproject.com/en/3.1/faq/install/

As a result of adding the option to `compilemessages`, `tools/update_messages`
started to faild with exit code of 2.
Based on an internal sync, we decided to keep the change on hold until we
migrate the project to Python 3 (and consequently migrate django to 3.x as
well).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ tests/server_test_cases/.cache/
 .pytest_cache/
 .vscode/
 env/
+venv/
--- a/app/app.yaml
+++ b/app/app.yaml
@@ -45,6 +45,10 @@ handlers:
   script: wsgi.application
 - url: /setup_datastore/?
   script: wsgi.application
+- url: /global/?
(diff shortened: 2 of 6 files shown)


ryosuketc:
I included changes previously merged by mistaken. I will file another PR.

PR #770, opened 2021-01-08
Add an executable wrapper of tools/download_feed.py.

gimite:
Hitting tools/download_feed.py directly doesn't work due to dependencies, so we need an wrapper.

(It used to work somehow, but I don't know how it used to work...)


ryosuketc (approved):
LGTM. Thanks!

gimite:
Thanks for the review!

Issue #771, opened 2021-01-18
appcfg is deprecated: Correct command in GettingStarted Step 6


ryosuketc:
`tools/gae update app -A <project_id> --noauth_local_webserver` fails due to deprecation of `appcfg`.
Step 6 in [GettingStarted](https://github.com/google/personfinder/wiki/GettingStarted)

The command has to be migrated to `gcloud`.

```
 % tools/gae update app -A <project_id> --noauth_local_webserver
2020-11-25 00:19:36,626 WARNING appcfg.py:5217 The appcfg command is deprecated. Please download the Cloud SDK for
App Engine development. https://cloud.google.com/appengine/downloads

...(omitted)...
 
12:21 AM Application deployment failed. Message: Deployments using appcfg are no longer supported. See https://cloud.google.com/appengine/docs/deprecations
2020-11-25 00:21:44,800 ERROR appcfg.py:2564 Application deployment failed. Message: Deployments using appcfg are no longer supported. See https://cloud.google.com/appengine/docs/deprecations 
12:21 AM Rolling back the update.
Could not start serving the given version.
```

ryosuketc:
DEPRECATED: 
```
tools/gae update app -A my-app-id --noauth_local_webserver
```
-> It eventually invokes:
https://github.com/google/personfinder/blob/8c31f8e6044b57e77e3898d8aaf32ed54bfdd3c8/tools/gae#L42

NEW ((my-app-id == Cloud project ID)):
```
gcloud app deploy app/app.yaml --project <my-app-id>
```

However, `application` and `version` needed to be removed to make it compatible with `gcloud` command.
See [app.yaml Reference  |  App Engine standard environment for Python 2](https://cloud.google.com/appengine/docs/standard/python/config/appref) as well as the following command line errors.

```
ERROR: (gcloud.app.deploy) The [application] field is specified in file [app.yaml file path]. This field is not used by gcloud and must be removed. Project name should instead be specified either by `gcloud config set project MY_PROJECT` or by setting the `--project` flag on individual command executions.

ERROR: (gcloud.app.deploy) The [version] field is specified in file [app.yaml file path]. This field is not used by gcloud and must be removed. Versions are generated automatically by default but can also be manually specified by setting the `--version` flag on individual command executions.
```

Note: While executing the NEW command, you might need to grant `Storage Object Viewer` permission to a service account. Follow error messages.

1.   Can we just remove `application` and `version` from `app.yaml`?
[...]

Issue #395, opened 2018-07-09
Migrate from appcfg.py to gcloud command


gimite:
Looks `gcloud` command is the standard way to configure App Engine app nowadays (though `appcfg.py` is still available). It would be better to migrate our tools / instructions to use `gcloud` command.

myoshimu:
Fix might be required for following directory:
tools/gae
tools/console

myoshimu:
https://cloud.google.com/appengine/docs/standard/java/tools/migrating-from-appcfg-to-gcloud

gimite:
Even with this one resolved, should we migrate to Python 3, since its support in App Engine is still beta?:
https://cloud.google.com/appengine/docs/standard/python3/quickstart

> This is a beta release of Python 3.7 in the App Engine standard environment. This feature might be changed in backward-incompatible ways and is not subject to any SLA or deprecation policy.

Stability is important for our product, so I'm a bit concerned to use an infra without SLA. It is up to you though.

nworden:
This is just one of many, many things we have to do before we eventually make the switch to Python 3. We won't be using the v3 runtime anytime soon.

gimite:
I personally feel it may be too early for us to take time to start preparing for Python 3 migration at this moment when it is unclear when App Engine Python 3 support exits beta. But it is also up to you.

nworden:
I'm not sure if you've seen the doc I sent to the team list yesterday (I'm on my phone right now and don't have my work account on it), but it looks like the Python 3 migration is going to be a lot of work for us because a lot of AE APIs (e.g., the Users service) aren't being supported for 3:
https://cloud.google.com/appengine/docs/standard/python3/python-differences#app_engine_apis
It's hard to make good plans because we don't know when the v3 runtime is coming out of beta or how long they're going to support 2.7, but I think it'd be good to start doing some of the work to prepare now, to be safe.

gimite:
Oh sorry I have just read the doc :) I agree that we should eventually migrate to Python 3, so it sounds good to start from something stable, like migrating to Python 3 language itself.

myoshimu:
@gimite @nworden 
Sorry for the delay in seeing this and let me consult how we should handle this issue.

When I tried to use Google Cloud SDK instead of the original App Engine SDK, it worked with the following command as the old SDK(including new React UI) without problem.
$ /opt/google-cloud-sdk/bin/dev_appserver.py app/app.yaml --port 8000 --admin_port=8001

[...]

Issue #772, opened 2021-01-22
Incorrect arguments are passed in docker/gae-run-app.sh


ryosuketc:
The arguments given to https://github.com/google/personfinder/blob/d845d5a801565a512e4501bb34c03f3ca789a125/docker/gae-run-app.sh#L11 might be malformed.
https://cloud.google.com/appengine/docs/standard/python/tools/local-devserver-command

Current: `--host=${IPADDR} --port=${PORT}`
Correct: `--host ${IP_ADDR} --port ${PORT}`

```
 % docker/gae-run-app.sh
Device "eth0" does not exist.
Starting Person Finder server
Setting datastore_path to /tmp/8000.datastore
usage: dev_appserver.py [-h] [-A APP_ID] [--host HOST] [--port PORT]
                        [--specified_service_ports SPECIFIED_SERVICE_PORTS]
                        [--admin_host ADMIN_HOST] [--admin_port ADMIN_PORT]
                        [--auth_domain AUTH_DOMAIN] [--storage_path PATH]
                        [--log_level {debug,info,warning,critical,error}]
                        [--max_module_instances MAX_MODULE_INSTANCES]
                        [--use_mtime_file_watcher [USE_MTIME_FILE_WATCHER]]
                        [--threadsafe_override THREADSAFE_OVERRIDE]
                        [--addn_host ADDN_HOST]
                        [--enable_host_checking [ENABLE_HOST_CHECKING]]
                        [--enable_console [ENABLE_CONSOLE]]
                        [--ssl_certificate_path SSL_CERTIFICATE_PATH]
                        [--ssl_certificate_key_path SSL_CERTIFICATE_KEY_PATH]
                        [--php_executable_path PATH]
                        [--php_remote_debugging [PHP_REMOTE_DEBUGGING]]
                        [--php_gae_extension_path PATH]
                        [--php_xdebug_extension_path PATH]
                        [--appidentity_email_address APPIDENTITY_EMAIL_ADDRESS]
                        [--appidentity_private_key_path APPIDENTITY_PRIVATE_KEY_PATH]
                        [--python_startup_script PYTHON_STARTUP_SCRIPT]
                        [--python_startup_args PYTHON_STARTUP_ARGS]
                        [--python27_executable_path PYTHON27_EXECUTABLE_PATH]
                        [--jvm_flag JVM_FLAG] [--go_work_dir GO_WORK_DIR]
                        [--enable_watching_go_path [ENABLE_WATCHING_GO_PATH]]
[...]

PR #773, opened 2021-01-22
Migrate tools/gae from appcfg to gcloud

ryosuketc:
-   To run a local development server, keep using this wrapper `tools/gae_run` 
-   To deploy app, directly invoke `gcloud` command. Make sure to set it up beforehand.
[Installing Google Cloud SDK  |  Cloud SDK Documentation](https://cloud.google.com/sdk/docs/install)

Previous interface (example):
`tools/gae run app --port=8000`
`tools/gae update app -A <my-app-id> --noauth_local_webserver`

New interface (example):
`tools/gae_run app --port=8000`
`gcloud app deploy app/app.yaml --project <my-app-id> --no-promote`

This PR solely fix https://github.com/google/personfinder/issues/771 and makes progress on https://github.com/google/personfinder/issues/395.


tools/gae_run
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# Copyright 2019 Google Inc.
```
gimite, line 2: Update Copyright year.

tools/gae_run
```diff
@@ -0,0 +1,37 @@
...
+# If no datastore path was given, pick one based on the directory name.
+if [ "$datastore_path" == "" -a "$app_dir" != "" ]; then
+    datastore_path="/tmp/$app_dir.datastore"
+    extra_args="--datastore_path=$datastore_path"
+    echo "Setting datastore_path to $datastore_path"
+fi
+
+# Run the local development appserver.
+# Default ports are as follows but could be overridden by "$@".
+$GCLOUD_DIR/dev_appserver.py --port 8000 --admin_port=8001 $extra_args "$@"
```
gimite, line 37: What is the default behavior of dev_appserver when --port and --datastore_path is omitted respectively? If both are reasonable (with some default), what if we start to use dev_appserver.py directly, instead of keeping this script?

gimite (changes requested):

tools/gae_run
ryosuketc: Default ports are:
--port=8080
--admin_port=8000

--storage_path controls --datastore_path, --blobstore_path, and --logs_path. The default is not specified in --storage_path section but --logs_path says it stores data only in memory.
> --logs_path=...
By default, the logs for the local development server are stored in memory only.  

https://cloud.google.com/appengine/docs/standard/python/tools/local-devserver-command

ryosuketc: I think ports are ok but are we okay to keep data only in memory? The current script enforces to set -datastore_path.

gimite: It's unclear from what you said above if the datastorte is kept in memory or disk. You mentioned *logs* are only stored on memory (which sounds acceptable) but not for datastore. Can you check what is the default behavior for datastore?

I agree that it would be inconvenient if the datastore is stored in memory.

gimite (changes requested):

tools/gae_run
ryosuketc: Sorry for being vague on that.

Datastore itself should be viewable at localhost/datastore ([document](https://cloud.google.com/appengine/docs/standard/python/tools/using-local-server)).
However, I could not pinpoint the file location though I tried grep / find in the shell in my user directory.
Some resources indicates that the file should be in something like a temp directory (probably outside of pf dir) in a user directory though (Note: The resource looks to be several years old). 
https://stackoverflow.com/questions/35719799/where-does-app-engine-dev-appserver-store-the-datastore

Regardless, it might be cleaner to keep datastore in pf dir (= use this wrapper)? WDYT?

gimite: > Regardless, it might be cleaner to keep datastore in pf dir (= use this wrapper)?

The current `tools/gae` stores the data store in a temp directory too, not in the PF project directory. See this line:

`datastore_path="/tmp/$app_dir.datastore"`

So there wouldn't be a difference if the default behavior is to store it in a temp directory.

Can you try running `dev_appserver.py` twice without specifying `--datastore_path` to see if the data is retained? If it is retained (which means the datastore is stored somewhere on the disk, not on memory), I guess the default behavior is sufficient. WDYT?

gimite (changes requested):

tools/gae_run
ryosuketc: You are right - thanks for the correction.
Datastore seems to be retained. Then we can have developers run the command directly by updating the relevant documentation.

Do we want to keep tools/gae or delete it then? If we delete it, we might need to confirm this docker implementation works fine with the new command (https://github.com/google/personfinder/issues/772). But Docker might be lower priority.
https://github.com/google/personfinder/blob/d845d5a801565a512e4501bb34c03f3ca789a125/docker/gae-run-app.sh#L11

gimite: In that case, I prefer to delete `tools/gae` to avoid confusion.

> If we delete it, we might need to confirm this docker implementation works fine with the new command (#772).

Right, it would be nice if we fix Docker config, but that would be in lower priority. You can either:

* Leave it as is for now
* Attempt to fix it (e.g., replace `tools/gae` with `dev_appserver.py`), but maybe without testing it.
* Delete the Docker confing (as it is probably not actively used)

gimite (changes requested):

ryosuketc:
> Attempt to fix it (e.g., replace tools/gae with dev_appserver.py), but maybe without testing it.
If that is okay not to test it perfectly, I would prefer this option since it could cause confusion to leave tools/gae itself and any code that depend on it.

I will go forward and delete it and update documentations.

gimite:
Sounds good, thanks!

tools/common.sh
```diff
@@ -39,6 +39,24 @@ if [ -z "$APPENGINE_DIR" ]; then
     exit 1
 fi
 
+for dir in \
```
ryosuketc, line 42: @gimite I implemented this before we agree on the decision to delete tools/gae.  It might be usable in the future but not used now. Please let me know if you prefer to delete it.

tools/common.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2019 Google Inc.
+# Copyright 2021 Google Inc.
```
gimite, line 2: FYI You don't need to update copyright year of an existing file. (I guess you can, but you don't need to.) This is only necessary for new files.

tools/common.sh
gimite: I prefer to delete this one if we are currently not using it.

But how about $APPENGINE_DIR? I guess it's still necessary to run some scripts, right? Do you expect developers to manually set $APPENGINE_DIR? Or should we update their defaults to point to the default location of AppEngine SDK in Cloud SDK?

gimite (changes requested):

tools/common.sh
ryosuketc: Thanks. I will revert it just in case.

ryosuketc:
As multiple conversations happened so far, I am summarizing the change here:
@gimite Could you review once again? If you agree with the summary, I will update the relevant documents accordingly.


- Delete `tools/gae` and have developers run proper commands offered by Cloud SDK. Here are a few samples.
- `docker/gae-run-app.sh` might not work properly but we decided just to replace the command for now due to low usage rate of docker implementation. 
- This change addresses some of the files in https://github.com/google/personfinder/issues/395#issuecomment-763340181.

Previous interface (example):
```
tools/gae run app --port=8000
tools/gae update app -A <my-app-id> --noauth_local_webserver
```
New interface (example):
```
dev_appserver.py app
gcloud app deploy app/app.yaml --project <my-app-id> --no-promote
```

To run appserver, use `dev_appserver.py` ([ref](https://cloud.google.com/appengine/docs/standard/python/tools/local-devserver-command)).
To deploy appm use `gcloud app deploy`. Use `--no-launch-browser` ([ref](https://cloud.google.com/sdk/gcloud/reference/auth/login?hl=ja#--launch-browser)) where applicable (https://github.com/google/personfinder/issues/771#issuecomment-765172742).

gimite:
Looks good to me!

tools/common.sh
```diff
@@ -22,8 +22,12 @@ export PROJECT_DIR=$(dirname $TOOLS_DIR)
 export APP_DIR=$PROJECT_DIR/app
 export TESTS_DIR=$PROJECT_DIR/tests
 
+# Look for non-standalone SDK inside google-cloud-sdk.
+# Then look for a standalone one (deprecated).
 for dir in \
     "$APPENGINE_DIR" \
+    $HOME/opt/google-cloud-sdk/platform/google_appengine \
+    $HOME/google-cloud-sdk/platform/google_appengine \
     /usr/lib/google_appengine \
```
gimite, line 31: Optional: You can delete the references to the paths to the old AppEngine SDK. They are no longer available, and people should update to the new one.

gimite (approved):

gimite:
Merging this PR now. Create a new PR if you want to apply my optional suggestion.

PR #775, opened 2021-02-10
Add an instruction on Application Default Credentials.

ryosuketc:
An auth error occurs when we access a production host using tools/console  unless ADCs are properly configured.
e.g. $ tools/console googlepersonfinder.appspot.com
https://github.com/google/personfinder/issues/395#issuecomment-776477763

### DIFF
--- a/tools/console
+++ b/tools/console
@@ -14,6 +14,10 @@
 # limitations under the License.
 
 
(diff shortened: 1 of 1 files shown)


gimite:
Thanks! Merged.

PR #776, opened 2021-02-16
Change max retry count of all task queues from infinite to 5.

gimite:
Tasks could continously fail for various reasons. I found it problematic
(e.g., It leaves a lot of pending tasks in the queue over time) to retry
them indefinitely.


ryosuketc (approved):
LGTM. Thanks!

PR #782, opened 2021-04-06
Add a hard dependency to rsa 4.0 specifically.

gimite:
`rsa` is already an indirect dependency of other requirements. Its latest version no longer supports Python 2.7. So it now causes an error on `pip install`. rsa 4.0 is the last version supporting Python 2 according to its release note.


ryosuketc (approved):
LGTM. Thanks!

gimite:
Thanks for the review!

Issue #779, opened 2021-03-16
Require reCAPTCHA ti


ryosuketc:
https://google.org/personfinder/test/create

PR #799, opened 2021-09-15
Require reCAPTCHA to create person records. #779

ryosuketc:
Require reCAPTCHA when creating person records to avoid fraud activities.
image


app/create.py
```diff
@@ -385,11 +386,25 @@ def get(self):
...
         self.render('create.html',
+                    captcha_html=self.get_captcha_html(),
                     profile_websites=profile_websites,
                     profile_websites_json=simplejson.dumps(profile_websites),
                     onload_function='view_page_loaded')
 
     def post(self):
+        captcha_response = self.get_captcha_response()
+        if not captcha_response.is_valid:
+            self.params.create_mode = True
```
gimite, line 397: Can you factor out line 397-405 as a helper method as it is duplicate of get() method?

app/resources/create.html.template
```diff
@@ -723,7 +723,7 @@
         </div>
 
         {% ifequal params.role "provide" %}
-          {% include "add_note_base.html.template" %}
+          {% include "add_note_base.html.template" with enable_recaptcha=True %}
```
gimite, line 726: Can you handle the case where params.role is not "provide"? It looks Captcha is not shown in that case. This is the case when you first click "I'm looking for someone", search for some name and click [Create new record] at the bottom.

gimite (changes requested):

app/create.py
ryosuketc: Done.  _render_create_html.

app/resources/create.html.template
ryosuketc: Done. Added reCAPTCHA in the other branch (sorry my oversight).
image

gimite:
Thanks! LGTM. Merged.

